PR #65, opened 2016-07-12
Set true division as default

simplicbe:
Set true division as the default option: `type(1/2)` will return `<type float>`.


slide:
LGTM

PR #66, opened 2016-07-12
PEP 3114 -- Renaming iterator.next() to iterator.__next__()

simplicbe:
Iterable must now implement `__next__` instead of `next`. Builtin `next` now calls `__next__`.

Works pretty well:

```python
>>> class Test04:
...     def __init__(self):
...         self.l = iter([1, 2, 3, 4])
...     def __iter__(self):
...         return self
...     def __next__(self):
...         return next(self.l)
...
>>> for a in Test04():
...     print (a)
```

Old style does not work anymore:

```python
>>> class Test05:
...     def __init__(self):
...         self.l = iter([1, 2, 3, 4])
...     def __iter__(self):
...         return self
...     def next(self):
...         return next(self.l)
...
>>> for b in Test05():
...     print (b)
...
TypeError: instance has no next() method
```

### DIFF
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -1537,12 +1537,12 @@ namespace IronPython.Modules {
         }
 
         public static object next(CodeContext/*!*/ context, object iter) {
-            return PythonOps.Invoke(context, iter, "next");
+            return PythonOps.Invoke(context, iter, "__next__");
         }
 
         public static object next(CodeContext/*!*/ context, object iter, object defaultVal) {
             try {
-                return PythonOps.Invoke(context, iter, "next");
+                return PythonOps.Invoke(context, iter, "__next__");
             } catch (StopIterationException) {
                 return defaultVal;
             }
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -1503,6 +1503,11 @@ namespace IronPython.Modules {
             return value;
         }
 
+        /// <summary>
+        /// Calls the __next__ method of an iterable
+        /// </summary>
+        /// <param name="iter">Iterable instance</param>
+        /// <returns>Next object or throws an StopIterator exception</returns>
         public static object next(IEnumerator iter) {
             if (iter.MoveNext()) {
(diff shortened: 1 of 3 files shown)


Src/IronPython/Modules/Builtin.cs
```diff
@@ -1503,6 +1503,11 @@ [NotNull]IList enumerator
             return value;
         }
 
+        /// <summary>
+        /// Calls the __next__ method of an iterable
+        /// </summary>
+        /// <param name="iter">Iterable instance</param>
+        /// <returns>Next object or throws an StopIterator exception</returns>
         public static object next(IEnumerator iter) {
             if (iter.MoveNext()) {
```
slide: Doesn't this need to be renamed __next__ for method resolution?

Src/IronPython/Modules/Builtin.cs
```diff
@@ -1511,6 +1516,12 @@ [NotNull]IList enumerator
             }
         }
 
+        /// <summary>
+        /// Calls the __next__ method of an iterable
+        /// </summary>
+        /// <param name="iter">Iterable instance</param>
+        /// <param name="defaultVal">Default operation value</param>
+        /// <returns>Next object or throws an StopIterator exception</returns>
         public static object next(IEnumerator iter, object defaultVal) {
```
slide: Same here

slide:
What happens if someone tries to call __next__ in their own code? It looks to me like it wouldn't resolve correctly unless the methods "next" were renamed to "__next__"

Src/IronPython/Modules/Builtin.cs
simplicbe: No, cause this is the `Builtin next` method, which is can be called directly as a function.

Src/IronPython/Modules/Builtin.cs
simplicbe: No, because this is the built-in  `next` method, not the one that must be implemented:
```python
...     def __next__(self):
...         return next(self.l)
```

simplicbe:
I don't really get what you mean. This is what has changed in IronPython 3, so `next` must be renamed to `__next__` in iterable classes. The built-in function `next` is still existing.

PR #82, opened 2016-08-03
Python 3 Exception chaining

simplicbe:
IronPython exception chaining, which enables the `from` keyword and additional exception properties.

Properties:
1. `__cause__`
2. `__traceback__`
3. `__context__`

Sample

``` python
try:
    int("abc")
except Exception as ex:
    raise NameError from ex # from ex is new in IP 3
```

prints

```
  File "<stdin>", line 2, in <module>
ValueError: Invalid integer literal

The above exception was the direct cause of the following exception:

  File "<stdin>", line 4, in <module>
NameError: Invalid integer literal
```

_Implicit exceptions_

``` python
try:
    int("abc")
except Exception as ex:
    raise NameError
```

prints out:

```
  File "<stdin>", line 2, in <module>
ValueError: Invalid integer literal

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
NameError: None
```


IronPython.sln
```diff
@@ -139,6 +139,8 @@ Global
 		{65E997B7-E99B-4C83-B29E-9951429BB293}.Release|Any CPU.Build.0 = Release|Any CPU
 		{65E997B7-E99B-4C83-B29E-9951429BB293}.Silverlight5Debug|Any CPU.ActiveCfg = Silverlight5Debug|Any CPU
 		{65E997B7-E99B-4C83-B29E-9951429BB293}.Silverlight5Release|Any CPU.ActiveCfg = Silverlight5Release|Any CPU
+		{65E997B7-E99B-4C83-B29E-9951429BB293}.v2Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
slide: What project did you add a v2 config for?

slide:
It would be great if this had some tests along with it :-)

IronPython.sln
simplicbe: Not sure why/when this happend

simplicbe:
Besides a few python scripts - nothing. But I'm trying to get a good and stable test suite running for IP3. There are some general issues (with testing)...

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -2329,27 +2329,31 @@ public static partial class PythonOps {
...
-        private static Exception MakeExceptionWorker(CodeContext/*!*/ context, object type, object value, object traceback, bool forRethrow) {
+        private static Exception MakeExceptionWorker(CodeContext/*!*/ context, object type, object value, object traceback, object cause, bool forRethrow) {
             Exception throwable;
             PythonType pt;
 
             if (type is PythonExceptions.BaseException) {
+                // Set nested python exception
+                var _cause = cause as PythonExceptions.BaseException;
+                ((PythonExceptions.BaseException)type).SetCause(_cause);
+
```
simplicbe: Does anyone has an Idea, who to access an *implicit exception here? For example for:

```
try:
    int("abc")
except Exception as ex:
    raise NameError
```

If `raise NameError` is executed, I need the instance if the exception which is thrown above. When using the `from` keyword I can pass the exception using the new implementation of cause, but not without `from`...

slide:
Looks like there are conflicts on this now. Are they easy to resolve?

simplicbe:
I'll try that later

slide:
Superceded by #132

PR #111, opened 2016-11-14
Fix ellipsis and 1else

slide:
This fixes test_ellipsis and test_float_exponent_tokenization in test_grammar.py

### DIFF
--- a/Src/IronPython/Compiler/Tokenizer.cs
+++ b/Src/IronPython/Compiler/Tokenizer.cs
@@ -420,6 +420,15 @@ namespace IronPython.Compiler {
                         if (ch >= '0' && ch <= '9')
                             return ReadFraction();
 
+                        if(NextChar('.')) {
+                            if (NextChar('.')) {
+                                MarkTokenEnd();
+                                return new ConstantValueToken(PythonOps.Ellipsis);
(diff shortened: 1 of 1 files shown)


Src/IronPython/Compiler/Tokenizer.cs
```diff
@@ -963,8 +972,14 @@ internal Tokenizer(ErrorSink errorSink, PythonCompilerOptions options, bool verb
                         return new ConstantValueToken(LiteralParser.ParseImaginary(GetTokenString()));
 
                     default:
-                        BufferBack();
-                        MarkTokenEnd();
+                        if (!Char.IsDigit((char)ch) && !Char.IsWhiteSpace((char)ch) && ch != EOF) {
+                            BufferBack();
```
slozier: I don't think this one is quite right. Just playing around with py3 there are other cases which are valid but fail `1 if 1e1else else 0`, `1 if 1e0else 0`, `1 if 1e10else 0`, etc.

slide: This was specifically for fixing the grammar test, not a complete fix.

slide:
I'll create a new PR for this in the future.

PR #117, opened 2016-12-07
Make the DLR a submodule of IronPython

slide:


slide:
This depends on https://github.com/IronLanguages/dlr/pull/9

Src/IronPythonTest/Cases/CaseExecuter.cs
```diff
@@ -26,9 +26,30 @@ class CaseExecuter {
                 Executable,
                 "");
 
+            AddSearchPaths(engine);
             return engine;
         }
 
+        private static void AddSearchPaths(ScriptEngine engine) {
+            var paths = new List<string>(engine.GetSearchPaths());
+            if(!paths.Any(x => x.ToLower().Contains("StdLib"))) {
```
slozier: Isn't `x.ToLower().Contains("StdLib")` always `false`?

.gitattributes
```diff
@@ -23,6 +23,10 @@
 *.vdproj text eol=crlf
 
 
+# Declare fiels that will always have LF line endings on checkout
```
slozier: Typo: fiels

Src/IronPythonTest/Cases/CaseExecuter.cs
slide: Yes, you are absolutely correct!

.gitattributes
slide: Thanks

PR #118, opened 2016-12-08
Initial clean-up of builtins

### LINKED ISSUE #119: zip() now returns an iterator
https://docs.python.org/3.0/whatsnew/3.0.html#views-and-iterators-instead-of-lists

### LINKED ISSUE #120: map() and filter() return iterators
https://docs.python.org/3.0/whatsnew/3.0.html#views-and-iterators-instead-of-lists

### LINKED ISSUE #121: The builtin basestring abstract type was removed
https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit

### LINKED ISSUE #122: Cleanup of the sys module: removed sys.exitfunc(), sys.exc_clear(), sys.exc_type, sys.exc_value, sys.exc_traceback
https://docs.python.org/3.0/whatsnew/3.0.html#library-changes

### LINKED ISSUE #123: Moved intern() to sys.intern()
https://docs.python.org/3.0/whatsnew/3.0.html#builtins

### LINKED ISSUE #124: Removed: apply(). Instead of apply(f, args) use f(*args).
https://docs.python.org/3.0/whatsnew/3.0.html#builtins

### LINKED ISSUE #125: Removed coerce(). This function no longer serves a purpose now that classic classes are gone.
https://docs.python.org/3.0/whatsnew/3.0.html#builtins

### LINKED ISSUE #126: Removed execfile(). Instead of execfile(fn) use exec(open(fn).read()).
https://docs.python.org/3.0/whatsnew/3.0.html#builtins

### LINKED ISSUE #127: Removed file. Use open().
https://docs.python.org/3.0/whatsnew/3.0.html#builtins

### LINKED ISSUE #128: Removed reduce(). Use functools.reduce()
https://docs.python.org/3.0/whatsnew/3.0.html#builtins

### LINKED ISSUE #129: Removed reload(). Use imp.reload().
https://docs.python.org/3.0/whatsnew/3.0.html#builtins

slide:
Fix #119, Fix #120, Fix #121, Fix #122, Fix #123, Fix #124, Fix #125, Fix #126, Fix #127, Fix #128, Fix #129

### DIFF
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -19,6 +19,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -46,7 +47,9 @@ using Complex = Microsoft.Scripting.Math.Complex64;
 namespace IronPython.Modules {
     [Documentation("")]  // Documentation suppresses XML Doc on startup.
     public static partial class Builtin {
-        public const string __doc__ = "Provides access to commonly used built-in functions, exception objects, etc...";
+        public const string __doc__ = @"Built-in functions, exceptions, and other objects.
+
+Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
         public const object __package__ = null;
         public const string __name__ = "builtins";
 
@@ -74,12 +77,6 @@ namespace IronPython.Modules {
         public static NotImplementedType NotImplemented {
             get {
                 return NotImplementedType.Value;
-            }
-        }
-
-        public static object exit {
-            get {
-                return "Use Ctrl-Z plus Return to exit";
             }
         }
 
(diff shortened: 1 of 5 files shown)


Src/IronPython/Modules/Builtin.cs
```diff
@@ -942,30 +761,12 @@ public static partial class Builtin {
...
-            return eval(context, raw_input(context, prompt));
-        }
-
         public static PythonType @int {
             get {
                 return DynamicHelpers.GetPythonTypeFromType(typeof(int));
             }
         }
 
-        public static string intern(object o) {
```
slozier: Did you move this to the sys module? I'm not seeing it.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -541,12 +541,25 @@ public static partial class PythonOps {
...
+            return NotImplementedType.Value;
+        }
+
         internal static int GetSequenceMultiplier(object sequence, object count) {
             int icount;
             if (!Converter.TryConvertToIndex(count, out icount)) {
                 PythonTuple pt = null;
                 if (count is OldInstance || !DynamicHelpers.GetPythonType(count).IsSystemType) {
-                    pt = Builtin.TryCoerce(DefaultContext.Default, count, sequence) as PythonTuple;
+                    pt = TryCoerce(DefaultContext.Default, count, sequence) as PythonTuple;
```
slozier: I think `__coerce__` is gone. Maybe we can get rid of this?

Src/IronPython/Modules/Builtin.cs
slide: You're right, I missed the add to sys.cs, I'll push an update.

Src/IronPython/Runtime/Operations/PythonOps.cs
slide: Yes, it is gone, but I was going to remove it as part of the removal of OldClass and friends.

PR #131, opened 2016-12-09
Update tokenizer to no longer accept octals (without the prefix) and the long suffix

slozier:


slozier:
Hmm, maybe hold off on this one, I don't think it's quite right. For example `000` should be valid but it's not.

slozier:
Seems to work now. Any idea if there are tests for this type of stuff?

slide:
test_int.py has some octal tests, but we can't run tests yet because the parser isn't up to snuff. https://github.com/IronLanguages/ironpython3/blob/master/Src/StdLib/Lib/test/test_int.py#L138

slide:
One thing I have been doing is copying some of the tests into a new file and then using -S to skip import of site.py.

slide:
Are you happy with this? If so, I'll go ahead and merge it.

slozier:
@slide Yes. Should be fine for now.

PR #133, opened 2016-12-11
Remove exec and backquote.

slide:
exec is no longer a keyword, just a function. backquote is removed.


slide:
Also, this implements "yield x from expr"

PR #136, opened 2016-12-12
Remove __hex__ and __oct__ and implement hex and oct using __index__.…

slide:
… Rename _winreg and copy_reg


slozier:
I'm not seeing the removal of `__oct__` and `__hex__` from `IntOps.cs` and `LongOps.cs`. Also maybe we could get rid of `FutureBuiltins.cs` at the same time?

PR #137, opened 2016-12-12
Update parser to handle new syntax for the function parameter list

slozier:


slide:
Are there any corner cases in the grammar that you haven't tested yet?

slozier:
Not that I'm aware of. I'll try to come up with a list of variations which we can eventually use for testing.

slide:
We also still need to implement the actual actions for the annotations and what not, correct, this is only the parser part?

slozier:
Yes, this only covers the parser. Things that need to be done outside the parser (I'm probably missing stuff here):
- Get rid of `SublistParameter`
- Do something with annotations
- Properly handle keyword-only arguments (i.e. parameters which follow the `*foo` argument). There's is currently an assertion error if the `*foo` argument isn't the last argument (or followed by `**bar`). Also need to handle the `*` argument.

Useful references:
- [8.6 Function definitions](https://docs.python.org/3/reference/compound_stmts.html#function-definitions)
- [PEP 3102 -- Keyword-Only Arguments](https://www.python.org/dev/peps/pep-3102/)
- [PEP 3107 -- Function Annotations](https://www.python.org/dev/peps/pep-3107/)
- [PEP 3113 -- Removal of Tuple Parameter Unpacking](https://www.python.org/dev/peps/pep-3113/)

PR #141, opened 2016-12-14
Remove OldClass/OldInstance, add nonlocal and async parsing

slide:


Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -367,15 +354,6 @@ public static partial class PythonOps {
                 while (ie.MoveNext()) {
                     PythonType baseType = ie.Current as PythonType;
 
-                    if (baseType == null) {
```
slozier: Should this be?
```C#
if (baseType == null) {
    continue;
}
```

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -925,7 +861,7 @@ public static partial class PythonOps {
 
         public static object ToPythonType(PythonType dt) {
             if (dt != null) {
```
slozier: Just make the function body `return (object)dt;`?

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1381,17 +1303,7 @@ public static partial class PythonOps {
...
-            }
-
-            // Otherwise, if there's a global variable named __metaclass__, it is used.
-            if (context.TryGetGlobalVariable("__metaclass__", out ret) && ret != null) {
-                return ret;
-            }
-
-            //Otherwise, the classic metaclass (types.ClassType) is used.
-            return TypeCache.OldInstance;
+            return DynamicHelpers.GetPythonType(bases[0]);
```
slozier: Does `bases` always have an element?

Src/IronPython/Runtime/PythonDocumentationProvider.cs
```diff
@@ -144,10 +121,6 @@ class PythonDocumentationProvider : DocumentationProvider {
                 kind = MemberKind.Field;
             } else if (value != null && value.GetType().IsEnum()) {
                 kind = MemberKind.EnumMember;
-            } else if (value is PythonType || value is OldClass) {
```
slozier: Did you mean to remove these? Or just the `OldClass` and `OldInstance` part?

Src/IronPython/Runtime/Types/PythonType.cs
```diff
@@ -505,15 +474,6 @@ internal PythonType(PythonContext context, PythonType[] baseTypes, Type underlyi
             foreach (object o in t) {
                 // gather all the type objects...
                 PythonType adt = o as PythonType;
-                if (adt == null) {
```
slozier: Shouldn't this be?
```C#
if (adt == null) {
    throw PythonOps.TypeError("expected tuple of types, got '{0}'", PythonTypeOps.GetName(o));
}
```

Src/IronPython/Runtime/Operations/PythonOps.cs
slide: Yes, new style classes (the only kind in 3.0) always inherit from object. I need to add the check in MakeClass to see if object was already passed as a base.

Src/IronPython/Runtime/Types/PythonType.cs
slide: I think things will always be PythonType now, but I need to verify.

Src/IronPython/Runtime/PythonDocumentationProvider.cs
slide: Good catch, no I didn't mean to remove them.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1347,6 +1345,15 @@ public static partial class PythonOps {
                     break;
                 }
             }
+
+            // we need to make sure that object is always a base
+            if(!bases.Any(x => x == objType)) {
```
slozier: Is this really true? Don't we just need bases to be non-empty?
```Python
class test:
  pass

class test2(test):
  pass

assert test2.__bases__ == (test,)
```

slide: Yes, you are correct.

simplicbe:
It is hard to review, because we have 3 things in one pr. Trying to get through it.

slide:
The vast majority is the old class/old instance removal. The nonlocal and async are localized to the parser and tokenizer for now.

PR #152, opened 2017-01-16
Remove <> operator

slozier:


slide:
Can you also update the generator python script(s) so that if we use them in the future, we don't revert this?

slozier:
Done.

slozier:
Rebased and regenerated files using the generator script.

PR #153, opened 2017-01-16
Remove __*slice__ functions

slozier:


slide:
Please update the generator python scripts as well (_weakref has one).

slozier:
The code in `_weakref.Generated.cs` is outside the generated block.

PR #162, opened 2017-01-20
PEP 3110 - Catching Exceptions in Python 3000

slozier:


simplicbe:
Do we need some tests for the variable scope? Beside: LGTM

slozier:
I except there would be one in the Python3 standard tests, but who knows...

Looks like there's a bit more work to do. The following is failing:
```Python
try:
    raise Exception()
except Exception as e:
    del e
```

PR #186, opened 2017-02-03
Turn off debug mode when running tests

slozier:
Also, as a side effect of turning off debug mode, the tests run in a fraction of the time (e.g. 18 seconds on my machine!)


slide:
Awesome!

slide:
Looks like frames and fullframes are on too, we probably want to make that a default for ipy3

PR #188, opened 2017-02-04
Make all classes inherit from object

slide:
If no base is provided, then object should be an implicit base class.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -1344,6 +1344,11 @@ namespace IronPython.Runtime.Operations {
                 }
             }
 
+            // all classes should inherit from object, so if no base is provided, add object as a base
+            if(bases.Length == 0) {
+                bases = new object[] { DynamicHelpers.GetPythonTypeFromType(typeof(object)) };
+            }
+
             PythonTuple tupleBases = PythonTuple.MakeTuple(bases);
 
             object metaclass;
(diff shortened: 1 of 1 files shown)


slozier:
I actually removed this on purpose since it was causing failures...

slozier:
So it looks like we need to add object to the class bases, but we shouldn't add it before calling the metaclass...

Here's the relevant test case:
```Python
class MyMetaClass(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        assert bases == ()
        return dict()
    def __new__(cls, name, bases, classdict):
        assert bases == ()
        return type.__new__(cls, name, bases, classdict)

class MyClass(metaclass=MyMetaClass):
    pass

assert MyClass.__bases__ == (object,)
```

slide:
I added it because I was seeing some test failures :)

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1344,11 +1344,6 @@ public static partial class PythonOps {
...
             }
 
-            // all classes should inherit from object, so if no base is provided, add object as a base
-            if(bases.Length == 0) {
-                bases = new object[] { DynamicHelpers.GetPythonTypeFromType(typeof(object)) };
-            }
-
             PythonTuple tupleBases = PythonTuple.MakeTuple(bases);
 
             object metaclass;
```
slozier: Did you verify what happens if the branch below is taken?

PR #191, opened 2017-02-05
Temporarily replace yield from with yield in module used by testing

slozier:
This will need to be reverted once yield from is properly implemented.


slide:
Maybe add a TODO to switch back?

slozier:
I'll file an issue when it's merged in.

PR #195, opened 2017-02-05
[WIP] Initial implementation of new super

slide:
Needs cleanup and feedback

### DIFF
--- a/Src/IronPython/Compiler/Ast/CallExpression.cs
+++ b/Src/IronPython/Compiler/Ast/CallExpression.cs
@@ -37,11 +37,11 @@ namespace IronPython.Compiler.Ast {
 
     public class CallExpression : Expression, IInstructionProvider {
         private readonly Expression _target;
-        private readonly Arg[] _args;
+        private readonly List<Arg> _args;
 
         public CallExpression(Expression target, Arg[] args) {
             _target = target;
-            _args = args;
+            _args = new List<Arg>(args);
         }
 
         public Expression Target {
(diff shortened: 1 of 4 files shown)


Src/IronPython/Compiler/Parser.cs
```diff
@@ -1737,12 +1766,18 @@ struct WithItem {
...
 
-                if (metaclass != null) {
-                    l.Add(new AssignmentStatement(new[] { new NameExpression("__metaclass__") }, metaclass));
+                if (clsContext != null) {
+                    if (!clsContext.InMember && clsContext.Metaclass != null) {
+                        l.Add(new AssignmentStatement(new[] { new NameExpression("__metaclass__") }, clsContext.Metaclass));
+                    }
+                    
+                    if (clsContext.InMember) {
+                        l.Add(new AssignmentStatement(new[] { new NameExpression("__class__") }, new NameExpression(clsContext.Name)));
```
slozier: While this would work as an initial implementation but I don't think it should be done at this level. I think I have something that might work for this, if I have time I'll post something this afternoon...

The tests below pass in CPython but fail with this PR:
```Python
class test:
  def test(self):
    assert "__class__" not in locals()

test().test()

class test:
  def test(self, __class__=1):
    assert __class__ == 1

test().test()
```

Src/IronPython/Compiler/Ast/CallExpression.cs
```diff
@@ -37,11 +37,11 @@ namespace IronPython.Compiler.Ast {
 
     public class CallExpression : Expression, IInstructionProvider {
         private readonly Expression _target;
-        private readonly Arg[] _args;
+        private readonly List<Arg> _args;
 
         public CallExpression(Expression target, Arg[] args) {
             _target = target;
-            _args = args;
+            _args = new List<Arg>(args);
```
slide: Because I have to be able to add arguments to the super(). If its an array, I can't add arguments (__class__ and self/cls)

Src/IronPython/Compiler/Parser.cs
slide: Yeah, this is why it's a WIP

slide: Looks like I need to insert an assignment only if super or \_\class\_\_ are found in the method, but only the first time it's encountered in the method.

slide:
@slozier I am not sure I liked how I solved the issue, but it's a WIP and feedback is good.

slozier:
@slide Hmm, so you only add `__class__` if super is defined? The following appears to be perfectly valid in CPython:
```Python
class test:
  def test(self):
    return __class__

assert test().test() == test
```

Here's an example of what I had in mind for adding the `__class__` variable: https://github.com/slozier/ironpython3/commit/ff17804e155e2190287df5251f92683eba34e2fc. It seemed to work with the few tests I did, but the Ast stuff is still something of a mystery for me...

slide:
Yeah, this only fixed one of the places that \_\_class\_\_ was needed, which is if super is used.

slozier:
Anyways, I think I would keep the `__class__` stuff completely out of the parser...

slide:
Yeah, I agree. I wasn't very happy with the way you did it.

slide:
I mean the way I did it...Sorry that came out as "you" I definitely like the way you did it.

PR #203, opened 2017-02-11
Add limit to factorial

slozier:
The CPython `test_math` expects an `OverflowError` for large factorial values.

### DIFF
--- a/Src/IronPython.Modules/math.cs
+++ b/Src/IronPython.Modules/math.cs
@@ -376,7 +376,7 @@ namespace IronPython.Modules {
                 throw PythonOps.ValueError("factorial() not defined for negative values");
             }
             if (value > SysModule.maxsize) {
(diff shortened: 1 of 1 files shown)


Src/IronPython.Modules/math.cs
```diff
@@ -368,25 +368,16 @@ public static partial class PythonMath {
...
-            return (int)val;
+            return factorial((BigInteger)v0);
         }
 
         public static object factorial(BigInteger value) {
             if (value < 0) {
                 throw PythonOps.ValueError("factorial() not defined for negative values");
             }
+            if (value > SysModule.maxsize) {
+                throw PythonOps.OverflowError("factorial() argument should not exceed {0}", int.MaxValue);
```
slide: Should it be SysModule.maxsize instead of int.MaxValue?

slozier: Ah yes, I changed the one but forgot the other.

PR #218, opened 2017-02-17
Fix blocking test in test_syntax

slozier:
Prevents an infinite loop if the test doesn't fail as expected. Might be worth making a PR to CPython for this one.


slide:
Yes, we should submit a bug and patch to CPython.

PR #226, opened 2017-02-18
Get test_unpack working

slozier:
Updates the error message in `test_unpack` to match the CPython 3.5 exception message (which is what we output).


slide:
Just to make sure, when we move the stdlib to 3.5 (or higher) we will be just fine?

slozier:
Yes, this is `test_unpack.py` from CPython 3.5.

PR #239, opened 2017-02-22
Implement the atexit module

slide:
Some tests are still failing, but the overall implementation works.

### DIFF
--- a/Src/IronPython.Modules/atexit.cs
+++ b/Src/IronPython.Modules/atexit.cs
@@ -1,20 +1,53 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 using Microsoft.Scripting;
 
 using IronPython.Runtime;
 using Microsoft.Scripting.Runtime;
 
+using IronPython.Runtime.Operations;
+using IronPython.Runtime.Exceptions;
+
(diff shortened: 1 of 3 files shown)


Src/IronPython.Modules/atexit.cs
```diff
@@ -1,20 +1,53 @@
...
+
+        private class FunctionDescriptor {
+            public FunctionDescriptor(object func, IDictionary<object, object> kwargs, object[] args) {
+                Func = func;
+                KeywordArgs = kwargs;
+                Args = args;
+            }
+
+            public object Func {
+                get; private set;
```
slozier: You don't need the `private set;` since it's a readonly property.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1749,6 +1749,33 @@ public static partial class PythonOps {
         }
 #endif
 
+        public static void PrintException(CodeContext/*!*/ context, Exception/*!*/ exception) {
```
slozier: Any way to refactor this so we don't duplicate so much code?

slide: Yes, I'll look into it

PR #242, opened 2017-02-24
Make zip an IEnumerator instead of an IEnumerable

slozier:


slide:
Does this fix a specific test?

slozier:
There are some tests for zip under test_iter which are now passing with this change.

PR #253, opened 2017-02-28
Fix the issue where the underlying file gets closed when it shouldn't

slide:
The underlying file should not get closed if closefd is False. There are some corner cases that I am missing for sure, so please review.

### DIFF
--- a/Src/IronPython/Modules/_fileio.cs
+++ b/Src/IronPython/Modules/_fileio.cs
@@ -235,9 +235,7 @@ namespace IronPython.Modules {
                     return;
                 }
 
-                if(_closefd)
-                    flush(context);
-
+                flush(context);
                 _closed = true;
 
                 if (_closefd) {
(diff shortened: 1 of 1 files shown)


Src/IronPython/Modules/_fileio.cs
```diff
@@ -244,15 +235,19 @@ public FileIO(CodeContext/*!*/ context, string name, [DefaultParameterValue("r")
                     return;
                 }
 
-                flush(context);
+                if(_closefd)
```
slozier: Hmm, why wouldn't you want the flush to occur if `_closefd` is false?

slide: I believe that is how C Python does it, but I can review to make sure.

slide: Actually, you are correct, the flush DOES need to occur.

PR #255, opened 2017-02-28
Rename socket to _socket.

slide:


slozier:
I wonder if we should do the same for IPy2.7? CPython 2.7 also has the socket.py wrapper in the stdlib.

slide:
I'm good with that.

PR #261, opened 2017-03-02
Fixes for several tests in test_bool

slide:


Src/IronPython/Runtime/Binding/PythonProtocol.cs
```diff
@@ -108,7 +108,21 @@ static partial class PythonProtocol {
...
 
-            return Ast.NotEqual(callAsInt, AstUtils.Constant(0));
+            var res = Expression.Parameter(typeof(int));
+            return Ast.Block(
+                new[] { res },
+                Ast.Assign(res, callAsInt),
+                Ast.IfThen(Ast.LessThan(res, Ast.Constant(0)),
+                    Ast.Throw(
+                        Ast.Call(
+                            typeof(PythonOps).GetMethod("ValueError"),
```
slozier: I wonder if we should really be throwing errors like that in the IL.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -839,7 +845,7 @@ public static partial class PythonOps {
             }
 
             if (res < 0) {
-                throw PythonOps.ValueError("__len__ should return >= 0, got {0}", res);
+                throw PythonOps.ValueError("__len__ should return >= 0");
```
slozier: Missing the parentheses `__len__()`

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -470,6 +470,9 @@ public static partial class PythonOps {
         }
 
         public static object Is(object x, object y) {
+            if(x is bool && y is bool) {
```
slozier: What about reusing https://github.com/IronLanguages/main/pull/1332 instead of having something different? Unless this fixes something else...

slide: Yeah, that's a way better idea. I'll revert what I have for that and merge.

Src/IronPython/Runtime/Binding/PythonProtocol.cs
slide: We have to the way this is implemented. It doesn't use the normal __len__ call, so it has to be thrown in here.

Src/IronPython/Runtime/Operations/PythonOps.cs
slide: Good catch

PR #263, opened 2017-03-02
Add compile flag for Barry as BDFL

slide:


slide:
Yup, it really is :-)

PR #264, opened 2017-03-02
Fix several tests in test_abc.py

slide:
Add \_\_init__ methods to avoid object.\_\_init__ deprecation warnings. Add \_\_isabstractmethod__ implementations.

### DIFF
--- a/Src/IronPython/Runtime/Descriptors.cs
+++ b/Src/IronPython/Runtime/Descriptors.cs
@@ -30,8 +30,63 @@ namespace IronPython.Runtime {
     public class staticmethod : PythonTypeSlot {
         internal object _func;
 
-        public staticmethod(object func) {
-            this._func = func;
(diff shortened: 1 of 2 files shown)


Src/IronPython/Runtime/Descriptors.cs
```diff
@@ -30,10 +30,17 @@ namespace IronPython.Runtime {
...
         internal object _func;
 
-        public staticmethod(object func) {
-            this._func = func;
+        public staticmethod(CodeContext/*!*/ context, object func) {
+            __init__(context, func);
         }
 
+        public void __init__(CodeContext/*!*/ context, object func) {
+            if (!PythonOps.IsCallable(context, func))
```
slozier: It looks like CPython doesn't have this restriction on initialization (similarly for classmethod)

slide: OK, I can remove it.

PR #271, opened 2017-03-08
Fix issue with metaclass assignment corrupting __doc__

slide:
The IPy code looks at _statements[0].Documentation to get the documentation for a class. If the metaclass is present in the class definition, it becomes the the first statement and it's Documentation is null. This fixes this issue by making the metaclass assignment at a minimum the second statement in the class's suite.

### DIFF
--- a/Src/IronPython/Compiler/Parser.cs
+++ b/Src/IronPython/Compiler/Parser.cs
@@ -1696,14 +1696,17 @@ namespace IronPython.Compiler {
                     return ErrorStmt();
                 }
 
-                if (metaclass != null) {
-                    l.Add(new AssignmentStatement(new[] { new NameExpression("__metaclass__") }, metaclass));
-                }
-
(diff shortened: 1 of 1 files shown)


Src/IronPython/Compiler/Parser.cs
```diff
@@ -1696,14 +1696,17 @@ struct WithItem {
...
-                    l.Add(new AssignmentStatement(new[] { new NameExpression("__metaclass__") }, metaclass));
-                }
-
+                bool addMetaclass = true;
                 while (true) {
                     Statement s = ParseStmt();
 
                     l.Add(s);
+
+                    if(addMetaclass && metaclass != null) {
```
slozier: If you wanted to be minimal about it, you could probably get rid of `addMetaclass` and just move it after the while loop with an `l.Insert(1, ...)`.

PR #275, opened 2017-03-09
Add ability to specify base when int(bytes, base)

slide:


Src/IronPython/Runtime/Operations/IntOps.cs
```diff
@@ -174,12 +174,12 @@ public static partial class Int32Ops {
         }
 
         [StaticExtensionMethod]
-        public static object __new__(CodeContext/*!*/ context, PythonType cls, IList<byte> s) {
+        public static object __new__(CodeContext/*!*/ context, PythonType cls, IList<byte> s, int @base=10) {
```
slozier: Should this be `[DefaultParameterValue(10)]` instead? Or are we just using `DefaultParameterValue` all over the place because IronPython predates optional arguments? I'm not really sure...

Src/IronPython/Runtime/Operations/IntOps.cs
```diff
@@ -191,6 +191,8 @@ public static partial class Int32Ops {
             }
         }
 
+        [StaticExtensionMethod]
```
slozier: Don't think we need this since it's an internal method.

Src/IronPython/Runtime/Operations/IntOps.cs
slide: As discussed on gitter, we can move to optional parameters like this since the framework supports it directly.

slide: or rather C# makes it into a DefaultParameterValue attribute automagically

PR #276, opened 2017-03-09
Change DefaultParameterValue to using C# default parameter values

slide:


Src/IronPython.Modules/re.cs
```diff
@@ -306,7 +306,7 @@ internal RE_Pattern(CodeContext/*!*/ context, object pattern)
                 return RE_Match.makeMatch(_re.Match(input, pos), this, input, pos, input.Length);
             }
 
-            public RE_Match match(object text, [DefaultParameterValue(0)]int pos, int endpos) {
+            public RE_Match match(object text, int pos=0, int endpos=int.MaxValue) {
```
slozier: I'd just get rid of the default values here since we already have overloads to deal with those cases.

Src/IronPython/Modules/_bytesio.cs
```diff
@@ -313,7 +313,7 @@ public BytesIO(CodeContext/*!*/ context, [DefaultParameterValue(null)]object ini
                 }
                 
                 return seek(posInt, GetInt(whence));
-            }
+            }
```
slozier: Spaces?

Src/IronPython.Modules/_ctypes/_ctypes.cs
```diff
@@ -470,9 +473,9 @@ public static partial class CTypes {
             return @sizeof(DynamicHelpers.GetPythonType(instance));
         }
 
-        #endregion
```
slozier: Looks like some region tags got misaligned.

PR #284, opened 2017-03-12
Merge in IronPython 2.7 tests changes

slozier:
(excluding .net core stuff)


Tests/modules/system_related/sys_test.py
```diff
@@ -315,29 +321,56 @@ def g(a):
...
 def test_winver():
     import re
     #E.g., "2.5"
     Assert(re.match("^\d\.\d$", sys.winver) != None)
 
+@skip('cli') # BUG, ps1 is not set in non interactive mode
 def test_ps1():
     Assert(not hasattr(sys, "ps1"))
 
+@skip('cli') # BUG, ps2 is not set in non interactive mode
```
slide: Does this need to be filed and fixed on both ipy2 and ipy3?

Tests/test_generator.py
```diff
@@ -351,7 +351,7 @@ def lstate(size):
         func = """
 def fetest(%s):        
     ret = 0
-    for i in range(%i):
+    for i in xrange(%i):
```
slide: This should be range instead of xrange.

Tests/test_property.py
```diff
@@ -118,6 +118,27 @@ def test_prop_doc_only():
     AreEqual(x.fdel, None)
     AreEqual(x.__doc__, 'Holliday')
  
+def test_member_lookup_oldclass():
```
slide: This test shouldn't be there since there are no old classes anymore.

Tests/modules/system_related/sys_test.py
slozier: Probably, looks like the change comes from https://github.com/IronLanguages/main/commit/b6a245d8cafbdad26feff729c68f66ab11fd23c8

Tests/test_generator.py
slozier: I will fix this one since it was correct before and I inadvertently changed it. Note that I noticed quite a number of other similar cases (code inside a string) which are not valid Python 3 (a lot of print statements). I didn't touch them since the goal of this PR was just to bring in all the testing fixes/updates.

Tests/modules/system_related/sys_test.py
slozier: Looks like there was an issue and it got fixed (https://github.com/IronLanguages/main/issues/240). Will remove the skip and make a PR for ipy2.

PR #300, opened 2017-05-01
Removed cmp argument from sorted()

ivanbakel:
Fixes `make` failing, as `sort()` had the same argument removed but this didn't.

### DIFF
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -1465,20 +1465,15 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
         }
 
         public static List sorted(CodeContext/*!*/ context, object iterable) {
-            return sorted(context, iterable, null, null, false);
+            return sorted(context, iterable, null, false);
         }
 
-        public static List sorted(CodeContext/*!*/ context, object iterable, object cmp) {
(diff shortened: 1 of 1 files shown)


Src/IronPython/Modules/Builtin.cs
```diff
@@ -1465,20 +1465,15 @@ public class mapobject : IEnumerable {
...
+            return sorted(context, iterable, null, false);
         }
 
-        public static List sorted(CodeContext/*!*/ context, object iterable, object cmp) {
-            return sorted(context, iterable, cmp, null, false);
-        }
-
-        public static List sorted(CodeContext/*!*/ context, object iterable, object cmp, object key) {
-            return sorted(context, iterable, cmp, key, false);
+        public static List sorted(CodeContext/*!*/ context, object iterable, object key) {
```
slozier: We could probably just get rid of this overload and the one about since we're just calling with the default values.

ivanbakel: Should I make the changes now?

slozier: If you have time sure, otherwise no big deal.

slide: Please make the change, it would be nice to have it cleaned up.

PR #301, opened 2017-05-09
[WIP] new super implementation

slide:
This allows most of the stdlib tests to get by the calls to new super. Reviews would be great.

### DIFF
--- a/Src/IronPython/Compiler/Ast/CallExpression.cs
+++ b/Src/IronPython/Compiler/Ast/CallExpression.cs
@@ -22,6 +22,7 @@ using MSAst = Microsoft.Scripting.Ast;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Runtime.CompilerServices;
 
 using Microsoft.Scripting.Actions;
@@ -38,6 +39,7 @@ namespace IronPython.Compiler.Ast {
     public class CallExpression : Expression, IInstructionProvider {
         private readonly Expression _target;
         private readonly Arg[] _args;
+        private readonly IList<Arg> _implicitArgs = new List<Arg>();
 
         public CallExpression(Expression target, Arg[] args) {
             _target = target;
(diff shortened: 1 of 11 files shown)


Src/IronPython/Compiler/Ast/CallExpression.cs
```diff
@@ -48,10 +50,14 @@ public class CallExpression : Expression, IInstructionProvider {
             get { return _target; }
         }
 
-        public IList<Arg> Args {
+        public Arg[] Args {
             get { return _args; }
         } 
 
+        public IList<Arg> ImplicitArgs {
```
slozier: Maybe make this internal? Seems more like an implementation detail than something we want to expose in the ast.

Src/IronPython/Compiler/Parser.cs
```diff
@@ -1053,13 +1054,19 @@ public class Parser : IDisposable { // TODO: remove IDisposable
             // Save private prefix
             string savedPrefix = SetPrivatePrefix(name);
 
+            var ret = new ClassDefinition(name, bases.ToArray(), keywords.ToArray(), metaclass: metaclass);
```
slozier: Maybe I'm missing something in this file, but other than changing the code to match `FunctionDefinition`, is there anything else going on?

Src/IronPython/Compiler/Ast/PythonNameBinder.cs
```diff
@@ -385,6 +385,16 @@ class PythonNameBinder : PythonWalker {
         // CallExpression
         public override bool Walk(CallExpression node) {
             node.Parent = _currentScope;
+
```
slozier: The method should be moved out of the generated block and `generate_walker.py` should be updated accordingly.

Src/IronPython/Compiler/Ast/CallExpression.cs
```diff
@@ -73,21 +79,25 @@ public class CallExpression : Expression, IInstructionProvider {
                 }
             } else {
                 if (nameExpr.Name == "eval") return true;
-                if (nameExpr.Name == "execfile") return true;
```
slide: exec is now handled like a function and is handled separately.

Src/IronPython/Compiler/Parser.cs
slide: This is just a cleanup so that we don't have to pass the metaclass around.

PR #314, opened 2017-07-01
Port hashlib changes from ipy2

slozier:


Src/IronPython.Modules/hashlib/_sha256.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            [Documentation("copy() -> object (copy of this object)")]
+            public override HashBase<SHA256> copy() {
+                Sha256Object res = new Sha256Object();
+                res._hasher = CloneHasher();
+                return res;
+            }
+
+            protected override void CreateHasher() {
+#if SILVERLIGHT || WP75
```
gpetrou: Can we remove these?

slide: Yes, but let's make it a different PR

PR #325, opened 2017-07-12
Align build setup with ironpython2

slide:


slozier:
I think you may need to update the NuGet.config file as well.

PR #333, opened 2017-08-13
Convert to auto-property

gpetrou:


Src/IronPython/Modules/_ast.cs
```diff
@@ -696,33 +668,29 @@ public abstract class cmpop : AST
...
+                ifs = PythonOps.MakeEmptyList(listIfs.Length);
                 foreach (ComprehensionIf listIf in listIfs)
-                    _ifs.Add(Convert(listIf.Test));
+                    ifs.Add(Convert(listIf.Test));
             }
             
             internal static ComprehensionIterator[] RevertComprehensions(PythonList comprehensions) {
-                Generic.List<ComprehensionIterator> comprehensionIterators =
-                    new Generic.List<ComprehensionIterator>();
+                List<ComprehensionIterator> comprehensionIterators =
```
slide: Can use var here

gpetrou:
Do you need any other changes for this?

PR #334, opened 2017-08-13
Remove redundant qualifiers

gpetrou:


Src/IronPython/Modules/_ast.cs
```diff
@@ -461,11 +461,11 @@ public abstract class AST
             }
 
             internal static PythonList Convert(ComprehensionIterator[] iters) {
-                Generic.List<ComprehensionFor> cfCollector =
-                    new Generic.List<ComprehensionFor>();
-                Generic.List<Generic.List<ComprehensionIf>> cifCollector =
-                    new Generic.List<Generic.List<ComprehensionIf>>();
-                Generic.List<ComprehensionIf> cif = null;
+                List<ComprehensionFor> cfCollector =
```
slide: I'd rather replace these with var since we can tell the type.

gpetrou:
Is there anything left?

slide:
For some reason I am not getting emails on updates. I'll double check and merge soon

PR #335, opened 2017-08-13
Use null propagation

gpetrou:


Src/IronPython/Compiler/Ast/PythonNameBinder.cs
```diff
@@ -549,21 +549,15 @@ class PythonNameBinder : PythonWalker {
             if (node.Left != null) {
```
slide: Can we not use null propagation here too?

gpetrou:
Can we merge this?

PR #339, opened 2017-08-13
Use FindRoot instead of fixed path

gpetrou:


gpetrou:
Sure. Seems like I missed that.

PR #341, opened 2017-08-14
Remove unnecessary features and add FEATURE_TYPE_INFO

gpetrou:


Src/IronPython.Modules/time.cs
```diff
@@ -71,7 +71,6 @@ public static class PythonTime {
 
             // altzone, timezone are offsets from UTC in seconds, so they always fit in the
             // -13*3600 to 13*3600 range and are safe to cast to ints
-#if FEATURE_TIMEZONE
```
slozier: You want to get rid of the TimeZone stuff and keep the TimeZoneInfo stuff. See IronLanguages/ironpython2#203.

gpetrou: Oh, so we actually want to remove the code that is currently used?

slozier: In this case yes.

gpetrou: OK

slide:
@slozier You good with this one now?

PR #345, opened 2017-10-31
Updates from IPy2 changes

slide:


Src/IronPython/Runtime/ClrModule.cs
```diff
@@ -68,8 +68,8 @@ public static class ClrModule {
         private static int _isMono = -1;
         public static bool IsMono {
             get {
-                if (_isMono == -1) {
-                    _isMono = Type.GetType("Mono.Runtime") != null ? 1 : 0;
+                if(_isMono == -1) {
+                    _isMono = Type.GetType ("Mono.Runtime") != null ? 1 : 0;
```
gpetrou: Revert these?

Tests/compat/sbs_func_args.py
```diff
@@ -32,6 +32,8 @@ def f3(arg0, arg1, arg2, *arg3): return "same## %s %s %s %s" % (arg0, arg1, arg2
     g3 = o.M3
     
     #for peverify runs
+    if clr.IsNetStandard:
```
slozier: This would fail because it should be `clr.IsNetCoreApp`. Looks like ipy2 would need to be fixed as well. I'm guessing we don't run the sbs tests?

CurrentVersion.props
```diff
@@ -2,13 +2,11 @@
...
     <MajorVersion>3</MajorVersion>
-    <MinorVersion>0</MinorVersion>
+    <MinorVersion>4</MinorVersion>
     <MicroVersion>0</MicroVersion>
     <ReleaseLevel>alpha</ReleaseLevel>
     <ReleaseSerial>0</ReleaseSerial>
 
     <AssemblyRevision>0</AssemblyRevision>
-    
-    <DlrVersion>1.2.0-alpha1</DlrVersion>
```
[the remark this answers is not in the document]
slide: It is unused because we don't pull the DLR in as a nuget reference, we use a submodule. I can remove from ipy2 as well.

Package/msi/IronPython.Installer.wixproj
```diff
@@ -0,0 +1,144 @@
...
+    </WixConstant>
+    
+    <!-- These are needed by HarvestDirectory, one for each directory. -->
+    <WixConstant Include="DocDir">
+      <Value>Doc</Value>
+    </WixConstant>
+    <WixConstant Include="StdLibDir">
+      <Value>Lib</Value>
+    </WixConstant>
+    <WixConstant Include="PlatformsNet40Dir">
```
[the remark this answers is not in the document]
slide: Yes, the MSI stuff is definitely still a work in progress.

Tests/compat/sbs_func_args.py
slide: I will verify, but I think we are running at least some of the sbs tests.

CurrentVersion.props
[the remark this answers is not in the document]
slide: See ironpython2/249

Tests/compat/sbs_func_args.py
slide: See ironpython2/#250

PR #348, opened 2017-11-19
Use nameof instead of strings

gpetrou:


Src/IronPython/Runtime/Exceptions/PythonExceptions.Generated.cs
```diff
@@ -187,7 +187,7 @@ public partial class _StopIteration : BaseException {
         public static PythonType AttributeError {
             get {
                 if (AttributeErrorStorage == null) {
-                    Interlocked.CompareExchange(ref AttributeErrorStorage, CreateSubType(Exception, "AttributeError", (msg, innerException) => new AttributeErrorException(msg, innerException)), null);
+                    Interlocked.CompareExchange(ref AttributeErrorStorage, CreateSubType(Exception, nameof(PythonOps.AttributeError), (msg, innerException) => new AttributeErrorException(msg, innerException)), null);
```
slozier: Note that this section is part of generated code so the generation script should be modified to generate this code. That being said, it should probably be `nameof(AttributeError)` and not `nameof(PythonOps.AttributeError)`. The same applies to all other changes to this file.

gpetrou: I will revert the Generated files changes.

PR #351, opened 2017-11-19
Remove some silverlight checks from python files

gpetrou:


gpetrou:
You are right, most changes are obsolete now.

PR #347, opened 2017-11-17
Dictionary views

slozier:
Changes dictionary `keys`, `values` and `items` methods to return views instead of lists. Also eliminated the `iterkeys`, `itervalues` and `iteritems` methods.

This is a somewhat experimental PR in that it also makes `PythonDictionary.cs` IPY2/IPY3 compatible using preprocessor directived (meaning that the exact same file should work in both code bases).

Edit: `test_dictviews` should be passing all tests except for pickling related stuff.

PR #352, opened 2017-11-21
Fix test_contains

slozier:
- Adds extra tests to cover more cases than what's in `test_contains`.
- Fixes a bug with https://github.com/IronLanguages/ironpython3/pull/347
- Applies the patch from: https://github.com/IronLanguages/ironpython2/pull/240
- Fixes what appeared to be an unreachable code path in `PythonTypeInfo`.

This PR also applies to IPY2.

### DIFF
--- a/Src/IronPython.Modules/_collections.cs
+++ b/Src/IronPython.Modules/_collections.cs
@@ -267,8 +267,9 @@ namespace IronPython.Modules {
                 lock (_lockObj) {
                     int found = -1;
                     int startVersion = _version;
-                    WalkDeque(delegate(int index) {
-                        if (PythonOps.EqualRetBool(_data[index], value)) {
+                    WalkDeque((int index) => {
+                        var data = _data[index];
+                        if (ReferenceEquals(data, value) || PythonOps.EqualRetBool(data, value)) {
                             found = index;
                             return false;
                         }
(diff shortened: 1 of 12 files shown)


Src/IronPython.Modules/_collections.cs
```diff
@@ -267,8 +267,9 @@ public deque([ParamDictionary]IDictionary<object, object> dict, params object[]
                 lock (_lockObj) {
                     int found = -1;
                     int startVersion = _version;
-                    WalkDeque(delegate(int index) {
-                        if (PythonOps.EqualRetBool(_data[index], value)) {
+                    WalkDeque((int index) => {
+                        var data = _data[index];
+                        if (ReferenceEquals(data, value) || PythonOps.EqualRetBool(data, value)) {
```
slide: Do we want to change PythonOps.EqualRetBool to do the ReferenceEquals check?

slozier: I believe `PythonOps.EqualRetBool` is meant to be equivalent to `a == b`. For the enumerable stuff what we want is essentially `a is b or a == b`. I don't think we should change `EqualRetBool` since it would change the semantics, but I would have no problem with adding a new method (something like `IsOrEqualRetBool` maybe?)

slide: I think that's a good idea

Tests/test_in.py
```diff
@@ -15,7 +15,10 @@
 
 import unittest
 
-from iptest import run_test
+from iptest import run_test, is_cli
```
slide: Nit...please use alphabetic order :-)

PR #360, opened 2017-12-09
Change wpf module implementation, similar to sqlite

gpetrou:
Port of https://github.com/IronLanguages/ironpython2/pull/281


simplicbe:
@gpetrou is there any advantage of using that pattern?

gpetrou:
@simplicbe I just ported this from IronPython2.
I believe in was done for consistency with sqlite, but @slide can give more details on that.

slide:
Correct, it allows the IronPython.Wpf.dll to be in the same directory as the other IronPython assemblies.

PR #362, opened 2017-12-09
Remove some redundant qualifiers

gpetrou:


Src/IronPython/Modules/unicodedata.cs
```diff
@@ -397,17 +397,17 @@ internal static class StreamReaderExtensions {
...
-            this.Decomposition_Type = info[PropertyIndex.Decomposition_Type];
+            Name = info[PropertyIndex.Name].ToUpperInvariant();
+            General_Category = info[PropertyIndex.General_Category];
+            Canonical_Combining_Class = int.Parse(info[PropertyIndex.Canonical_Combining_Class]);
+            Bidi_Class = info[PropertyIndex.Bidi_Class];
+            Decomposition_Type = info[PropertyIndex.Decomposition_Type];
 
             string nvdes = info[PropertyIndex.Numeric_Value_Decimal];
-            this.Numeric_Value_Decimal = nvdes != "" ? (int?)int.Parse(nvdes) : null;
+            Numeric_Value_Decimal = nvdes != "" ? (int?)int.Parse(nvdes) : null;
```
slide: Can this use ?? instead?

gpetrou: We are checking if the string is empty in this case, not null. Or you mean something else?

slide: No, I was reading it at 3am and misread it :)

PR #365, opened 2017-12-09
Join declaration and assignment

gpetrou:


simplicbe:
@gpetrou if you rework those line, maybe switching to `var` is also a good task if possible. Thanks for the contribution!

gpetrou:
Sometimes I do that such as var bytes = new List<byte>(self._bytes); in this PR, but many times the type is not clear, so I leave it as it is.

PR #369, opened 2018-02-08
Update itertools method names

### LINKED ISSUE #267: Implement itertools.filterfalse
https://docs.python.org/3/library/itertools.html#itertools.filterfalse

From looking at the C Python code, it looks like it is an object a la the dict_keys and other things that have been turned into objects in Python 3

rtzoeller:
itertools.ifilterfalse and zip_longest had their "i" prefixes removed, and itertools.ifilter, imap, and izip were replaced by builtins.

Resolves #267.


slozier:
@rtzoeller Thanks!

PR #370, opened 2018-02-08
open now takes an optional opener argument

### LINKED ISSUE #250: open now takes an "opener" in 3.3
See What's New in Python 3.3

rtzoeller:
Resolves #250

### DIFF
--- a/Src/IronPython/Modules/_io.cs
+++ b/Src/IronPython/Modules/_io.cs
@@ -46,9 +46,9 @@ namespace IronPython.Modules {
         private const int O_RDWR = 0x0002;  
         
         private const int O_APPEND = 0x0008;
-        private const int O_CREAT = 0x0200;
-        private const int O_TRUNC = 0x0400;
-        private const int O_EXCL = 0x0800;
+        private const int O_CREAT = 0x0100;
+        private const int O_TRUNC = 0x0200;
+        private const int O_EXCL = 0x0400;
 
         [SpecialName]
         public static void PerformModuleReload(PythonContext/*!*/ context, PythonDictionary/*!*/ dict) {
--- a/Src/IronPython/Modules/_io.cs
+++ b/Src/IronPython/Modules/_io.cs
@@ -2741,7 +2741,8 @@ namespace IronPython.Modules {
             string encoding=null,
             string errors=null,
             string newline=null,
-            bool closefd=true
+            bool closefd=true,
+            object opener=null
         ) {
             int fd = -1;
             string fname = file as string;
@@ -2797,9 +2798,23 @@ namespace IronPython.Modules {
                 mode += '+';
             }
 
-            FileIO fio = fname != null
-                ? new FileIO(context, fname, mode, closefd)
-                : new FileIO(context, fd, mode, closefd);
+            FileIO fio;
+            if (opener != null) {
+                object fdobj = PythonOps.CallWithContext(context, opener, file, mode);
+                if (!(fdobj is int)) {
+                    throw PythonOps.TypeError("expected integer from opener");
+                }
+                
+                fd = (int) fdobj;
+                if (fd < 0) {
+                    throw PythonOps.ValueError("opener returned {0}", fd);
+                }
+                fio = new FileIO(context, fd, mode, closefd);
+            } else {
+                fio = fname != null
+                    ? new FileIO(context, fname, mode, closefd)
+                    : new FileIO(context, fd, mode, closefd);
+            }
 
             bool line_buffering = false;
             if (buffering == 1 || buffering < 0 && fio.isatty(context)) {
--- a/Src/IronPython/Modules/_io.cs
+++ b/Src/IronPython/Modules/_io.cs
@@ -41,6 +41,15 @@ namespace IronPython.Modules {
         private static readonly object _blockingIOErrorKey = new object();
         private static readonly object _unsupportedOperationKey = new object();
 
+        private const int O_RDONLY = 0x0000;
+        private const int O_WRONLY = 0x0001;  
+        private const int O_RDWR = 0x0002;  
+        
+        private const int O_APPEND = 0x0008;
+        private const int O_CREAT = 0x0200;
+        private const int O_TRUNC = 0x0400;
+        private const int O_EXCL = 0x0800;
+
         [SpecialName]
         public static void PerformModuleReload(PythonContext/*!*/ context, PythonDictionary/*!*/ dict) {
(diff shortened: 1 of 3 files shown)


rtzoeller:
I hand tested these changes and they _appeared_ to work, but I'm not sure that everything that needs to be changed is, especially wrt error handling.

Src/IronPython/Modules/_io.cs
```diff
@@ -2797,9 +2798,23 @@ public class TextIOWrapper : _TextIOBase, IEnumerator<object>, IEnumerable<objec
                 mode += '+';
             }
 
-            FileIO fio = fname != null
-                ? new FileIO(context, fname, mode, closefd)
-                : new FileIO(context, fd, mode, closefd);
+            FileIO fio;
+            if (opener != null) {
+                object fdobj = PythonOps.CallWithContext(context, opener, file, mode);
```
slozier: I don't think passing in the `mode` argument is correct here since the opened expects `flags`. (See https://docs.python.org/3/library/os.html#os.open for info in flags). That being said I'm not exactly sure how the flags value is generated.

slozier:
Thanks for the PR.

What sort of tests did you run? Could you add some of them to the Tests folder (I'm wondering if we should add some python3x.py test files to cover version changes -- @slide any thoughts on this?).

The basic test that I expect to see would be something like:
```Python
import os
with open("somefile.txt", "w", opener=os.open) as f:
    f.write("test")

# check file content and cleanup?
```

Src/IronPython/Modules/_io.cs
rtzoeller: Good catch.

It looks like `flags` is [created manually](https://github.com/python/cpython/blob/507434fd504f3ebc1da72aa77544edc0d73f136e/Lib/_pyio.py#L1465-L1495).

rtzoeller: I looked over the test code I had thrown together and realized I had manually set the flags to O_RDONLY in the opener, so I wasn't actually using the flags/mode variable being passed in. Using os.open directly indeed fails.

---

IMHO the cleanest way to go about creating the `flags` int is to create a function which builds the int from `FileMode` and `FileAccess` enums, effectively the opposite of [these functions](https://github.com/IronLanguages/ironpython3/blob/903293eef46c6e8ef764f5b2e35fa5168a0ccdf6/Src/IronPython.Modules/nt.cs#L1753-L1773). This approach would let us model the structure after the CPython implementation, without needing to manually or O_APPEND, O_CREAT, etc. together.

The issue I'm running into is that all of the existing O_* definitions live in `Src/IronPython.Modules/nt.cs`, which is not accessible from `Src/IronPython/Modules/_io.cs`. I'm not sure what the best cross-platform way to handle this is.

What are your thoughts?

slozier: I'm fine with redefining the constants locally. I believe they come from fcntl.h which isn't Python specific anyway.

rtzoeller:
Are their instructions on what needs to be done to run the tests? I'm having issues where all tests I run are failing due to an attribute error.

```
======================================================================
ERROR: test_array_type (__main__.ArrayTest)
----------------------------------------------------------------------
  File "C:\Users\Ryan Zoeller\RiderProjects\ironpython3\bin\Debug\net45\Lib\iptest\ipunittest.py", line 165, in setUp
    self._temporary_dir = os.path.join(self.temp_dir, "IronPython")
  File "C:\Users\Ryan Zoeller\RiderProjects\ironpython3\bin\Debug\net45\Lib\iptest\file_util.py", line 62, in get_temp_dir
    temp = self.get_environ_variable("TMP")
  File "C:\Users\Ryan Zoeller\RiderProjects\ironpython3\bin\Debug\net45\Lib\iptest\ipunittest.py", line 330, in get_environ_variable
    l = [os.environ[x] for x in list(os.environ.keys()) if x.lower() == key.lower()]
AttributeError: '_Environ' object has no attribute 'lower'
```

rtzoeller:
I added some tests to test_file.py, I'm not sure if this is where you envisioned them.

I also did a quick replace of all of the calls to `file()` with calls to `open()`, and the kwarg name change from `name` to `file`. There are still tests which are either broken or no longer applicable due to the changes made to file handling, but the quick renames I did revealed that we are at least passing the sanity tests. I'm happy to re-evaluate the existing tests in a later pull request to decide what's still relevant.

slozier:
For some reason `test_opener` is failing locally for me.
> OSError: (32, 'The process cannot access the file because it is being used by another process')

In theory I'm fine with using test_file (although maybe it should be renamed to test_open?) as long as we don't go and clobber it with the ipy2 version. Final decision is up to @slide though.

rtzoeller:
@slozier I haven't been able to reproduce that locally; can you try running the below code and let me know if either fails? It's a modified version of the existing test code, but also tests the existing method of opening a file from a file descriptor.

```python
import os 

temp_file = "foo.bar"
data = "test message\n"

def test_existing():
    fd = os.open(temp_file, os.O_CREAT | os.O_WRONLY)
    with open(fd, "w") as f:
        f.write(data)

    fd = os.open(temp_file, os.O_RDONLY)
    with open(fd, "r") as f:
        msg = f.read()
        assert(msg == data)

    os.unlink(temp_file)

def test_new():
    with open(temp_file, "w", opener=os.open) as f:
        f.write(data)

    with open(temp_file, "r", opener=os.open) as f:
        msg = f.read()
        assert(msg == data)
    
    os.unlink(temp_file)

try:
    test_existing()
except e:
    print("Existing failed")

try:
    test_new()
except:
    print("New failed")
```

Src/IronPython/Modules/_io.cs
```diff
@@ -41,6 +41,15 @@ public static partial class PythonIOModule {
         private static readonly object _blockingIOErrorKey = new object();
         private static readonly object _unsupportedOperationKey = new object();
 
+        private const int O_RDONLY = 0x0000;
+        private const int O_WRONLY = 0x0001;  
+        private const int O_RDWR = 0x0002;  
+        
+        private const int O_APPEND = 0x0008;
+        private const int O_CREAT = 0x0200;
```
slozier: These (O_CREAT, O_TRUNC, O_EXCL) are off by a factor of 2.

slozier:
@rtzoeller Thanks again for the PR. I fixed up the constants which were causing a failure for me.

rtzoeller:
Thanks for doing that @slozier; busy week at school so I didn't get back to this.

PR #380, opened 2018-03-01
More merges from IronPython2

slide:
IronLanguages/ironpython2/pull/269
IronLanguages/ironpython2/pull/267
IronLanguages/ironpython2/pull/287


slozier:
LGTM. Do we want to port the test from IronLanguages/ironpython2#267 as well?

slide:
Sure, it fails on IronPython3 right now, but it would be good to include the changes.

PR #385, opened 2018-04-14
Remove mandatory future features

slozier:
Makes absolute imports on by default. Removes code related to `with_statement` and `unicode_literals` which are already on by default.

### DIFF
--- a/Src/IronPython/Runtime/ModuleOptions.cs
+++ b/Src/IronPython/Runtime/ModuleOptions.cs
@@ -35,14 +35,6 @@ namespace IronPython.Runtime {
         /// </summary>
         Initialize = 0x0008,
         /// <summary>
-        /// Enable usage of the with statement
-        /// </summary>
-        WithStatement = 0x0010,
-        /// <summary>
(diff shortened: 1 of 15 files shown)


Src/IronPython/Runtime/ModuleOptions.cs
```diff
@@ -66,10 +58,6 @@ public enum ModuleOptions {
         /// </summary>
         Interpret = 0x1000,
         /// <summary>
-        /// String Literals should be parsed as Unicode strings
-        /// </summary>
-        UnicodeLiterals = 0x2000,
-        /// <summary>
```
slide: Are there actually removed from the flags in c Python as well? Or are they there and just unused?

slozier: The CO_FUTURE_* flags for these are still defined but unused. As far as I know ModuleOptions is our own enum so I can't think of a good reason to keep these.

slide:
Can you resolve the conflicts on this?

PR #396, opened 2018-08-17
Changes from ipy2

slozier:


slozier:
Also covers https://github.com/IronLanguages/ironpython2/pull/444

PR #405, opened 2018-08-24
Allow retrieval of StopIteration value from C# code

slide:


Src/IronPython/Runtime/Exceptions/StopIterationException.cs
```diff
@@ -61,10 +56,25 @@ public StopIterationException(string message, Exception innerException)
...
 
+        /// <summary>
+        /// Result of raise StopError(n)
+        /// </summary>
+        /// <returns>
+        /// value passed to StopError, or null if none
+        /// </returns>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1007:UseGenericsWhereAppropriate")]
+        [PythonHidden]
+        public object GetValue() {
```
slozier: What about?
`public object GetValue() => (((IPythonAwareException)this).PythonException as PythonExceptions._StopIteration)?.value;`

slide:
Actually, I'm not sure why I had this change. I don't currently have any need, so I think I will close it for now.

PR #421, opened 2018-08-28
Fix #395

### LINKED ISSUE #395: Change sys.platform to match CPython
Having `sys.platform == "cli"` is one of those things which just causes a bunch of compatibility issues in IronPython 2.7. For IronPython 3 we should consider changing it to a standard value. There are other mechanisms such as `platform.python_implementation() == "IronPython"` which can be used to figure out if we're running on IronPython or not.

Note that there's also `sys.implementation` since Python 3.3 which we could use so we don't have to `import platform` to do checks in the std lib.

slide:
```
IronPython 3.4.0a0 (3.4.0.0) on .NET 4.0.30319.42000 (64-bit)
>>> import sys
>>> assert sys.platform =="win32"
>>> assert sys.platform == "cli"
>>> assert not sys.platform != "win32"
>>> assert not sys.platform != "cli"
>>> assert not sys.platform == "bla"
>>> assert sys.platform != "bla"
```

### DIFF
--- a/Src/IronPython/Modules/sys.cs
+++ b/Src/IronPython/Modules/sys.cs
@@ -9,12 +9,14 @@ using System.IO;
 using System.Numerics;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Security;
 using System.Text;
 
 using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
+using IronPython;
 using IronPython.Runtime;
 using IronPython.Runtime.Exceptions;
 using IronPython.Runtime.Operations;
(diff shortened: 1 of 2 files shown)


slide:
The one thing I am not happy with in this is that `type(sys.platform)` is not `str`

slozier:
Yeah, unfortunately `assert "cli" == sys.platform` fails.

slide:
Yeah, the binder needs to look for the subclass and stuff like in the fix for str and OrderedSet stuff I implemented.

slozier:
Probably needs to be fixed at the `PythonType` level. Seems to me like `Extensible<string>` should have `str` in its bases.

slozier:
Something like adding
```C#
} else if (TryReplaceExtensibleWithBase(_underlyingSystemType, out baseType) && !baseType.IsGenericParameter) {
    // Extensible<x> inherits from x
} else {
```
to `PythonType.AddSystemBases`...

slide:
I'm closing this out because I think we can make a breaking change here and move to `sys.implementation.name` instead with IronPython3.

PR #425, opened 2018-09-10
Add `__length_hint__` support for built in types

rtzoeller:
Add `__length_hint__` to built in types as described in #6.

This also disables a test which was crashing the test runner; I don't think this crash is related to the changes.

### DIFF
--- a/Src/IronPython/Runtime/ByteArray.cs
+++ b/Src/IronPython/Runtime/ByteArray.cs
@@ -95,6 +95,16 @@ namespace IronPython.Runtime {
         }
 
         public void extend(object seq) {
+            // AddRange will ultimately ensure we have enough space without performing lots
+            // of allocations, so we don't actually want to use the length hint.
+            // However, we need to call __len__ and __length_hint__ and attempt to convert the
+            // result to an int in order to match CPython behavior.
+            object len_obj;
+            if (PythonTypeOps.TryInvokeUnaryOperator(DefaultContext.Default, seq, "__len__", out len_obj) ||
+                PythonTypeOps.TryInvokeUnaryOperator(DefaultContext.Default, seq, "__length_hint__", out len_obj)) {
+                int len = Converter.ConvertToInt32(len_obj);
+            }
+
             extend(GetBytes(seq));
         }
 
--- a/Src/IronPython/Runtime/ByteArray.cs
+++ b/Src/IronPython/Runtime/ByteArray.cs
@@ -95,15 +95,11 @@ namespace IronPython.Runtime {
         }
 
         public void extend(object seq) {
-            // AddRange will ultimately ensure we have enough space without performing lots
-            // of allocations, so we don't actually want to use the length hint.
-            // However, we need to call __len__ and __length_hint__ and attempt to convert the
-            // result to an int in order to match CPython behavior.
-            object len_obj;
-            if (PythonTypeOps.TryInvokeUnaryOperator(DefaultContext.Default, seq, "__len__", out len_obj) ||
-                PythonTypeOps.TryInvokeUnaryOperator(DefaultContext.Default, seq, "__length_hint__", out len_obj)) {
-                int len = Converter.ConvertToInt32(len_obj);
-            }
+            // We don't make use of the length hint when extending the byte array.
+            // However, in order to match CPython behavior with invalid length hints we
+            // we need to go through the motions and get the length hint and attempt
+            // to convert it to an int.
+            PythonOps.TryInvokeLengthHint(DefaultContext.Default, seq, out int len);
 
             extend(GetBytes(seq));
         }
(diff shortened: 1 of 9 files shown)


rtzoeller:
I implemented this back in the spring but never submitted a pull-request; I've rebased it onto master and made sure the code behaves the same as it did when I originally wrote it.

The code could be cleaned up, but I wanted to get feedback on it before it rotted any more.

Src/IronPython/Runtime/List.cs
```diff
@@ -91,23 +91,21 @@ public class List : IList, ICodeFormattable, IList<object>, IReversible, IStruct
...
-                    _size = 0;
-                    extend(sequence);
-                } else {
-                    _data = new object[INITIAL_SIZE];
-                    _size = 0;
-                    extend(sequence);
+                object len_obj;
+                if (PythonTypeOps.TryInvokeUnaryOperator(context, sequence, "__len__", out len_obj) ||
+                    PythonTypeOps.TryInvokeUnaryOperator(context, sequence, "__length_hint__", out len_obj)) {
+                    if (!(len_obj is NotImplementedType)) {
```
slozier: You have this `NotImplementedType` check in some places and not in others. Is this something that should be checked all the time? Maybe we should generalize this into a `PythonOps` method? Maybe something like:
```C#
public bool TryInvokeLengthHint(CodeContext context, object sequence, out int hint);
```

Src/StdLib/Lib/test/test_iterlen.py
```diff
@@ -223,7 +232,12 @@ def test_issue1242657(self):
     def test_invalid_hint(self):
         # Make sure an invalid result doesn't muck-up the works
         self.assertEqual(list(NoneLengthHint()), list(range(10)))
-
+        # It's a TypeError if __length_hint__ returns something other than an int or NotImplemented
+        self.assertRaises(TypeError, list, StringLengthHint())
```
slozier: We typically put any custom tests in a file under the root `Tests` folder instead of modifying the StdLib.

Src/StdLib/Lib/test/test_iterlen.py
```diff
@@ -148,6 +148,7 @@ class TestList(TestInvariantWithoutMutations, unittest.TestCase):
     def setUp(self):
         self.it = iter(range(n))
 
+    @unittest.skip("Test crashing test suite")
```
slozier: Probably best to put a comment to indicate the skip is specifically for ironpython. Makes it easier to sync the StdLib files with CPython.

Src/StdLib/Lib/test/test_iterlen.py
rtzoeller: I should have time to pull this out tomorrow.

rtzoeller:
There is one failing test (other than the crashing one being skipped) in `test_iterlen.py`. I believe this failure is caused by https://github.com/IronLanguages/ironpython2/issues/387

slide:
This is looking good to me.

slozier:
If there's only one test failing in test_iterlen then it would be nice to disable it (and add the reference to the issue) to get it passing.

slide:
You can do that using 

```@unittest.skipIf(sys.implementation.name=='ironpython', 'URL FOR ISSUE HERE')
def testSomeFunc(...)
```

slide:
@rtzoeller Can you update the test with the skipIf?

Src/IronPython/Runtime/ByteArray.cs
```diff
@@ -95,6 +95,16 @@ public class ByteArray : IList<byte>, ICodeFormattable, IBufferProtocol {
...
 
         public void extend(object seq) {
+            // AddRange will ultimately ensure we have enough space without performing lots
+            // of allocations, so we don't actually want to use the length hint.
+            // However, we need to call __len__ and __length_hint__ and attempt to convert the
+            // result to an int in order to match CPython behavior.
+            object len_obj;
+            if (PythonTypeOps.TryInvokeUnaryOperator(DefaultContext.Default, seq, "__len__", out len_obj) ||
+                PythonTypeOps.TryInvokeUnaryOperator(DefaultContext.Default, seq, "__length_hint__", out len_obj)) {
+                int len = Converter.ConvertToInt32(len_obj);
```
rtzoeller: This should use the helper function in PythonOps; will update tomorrow.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -833,6 +833,19 @@ public static partial class PythonOps {
             return res;
         }
 
+        public static bool TryInvokeLengthHint(CodeContext context, object sequence, out int hint) {
```
slozier: Maybe make this internal?

PR #428, opened 2018-09-13
Fix #395

### LINKED ISSUE #395: Change sys.platform to match CPython

slide:
Change sys.platform to match the executing platform

### DIFF
--- a/Src/IronPython/Modules/sys.cs
+++ b/Src/IronPython/Modules/sys.cs
@@ -9,12 +9,14 @@ using System.IO;
 using System.Numerics;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Security;
 using System.Text;
 
 using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
+using IronPython;
 using IronPython.Runtime;
 using IronPython.Runtime.Exceptions;
 using IronPython.Runtime.Operations;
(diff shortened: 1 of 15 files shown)


Src/IronPython/Modules/sys.cs
```diff
@@ -751,6 +751,14 @@ public class floatinfo : PythonTuple {
             dict["stdin"] = dict["__stdin__"];
             dict["stdout"] = dict["__stdout__"];
             dict["stderr"] = dict["__stderr__"];
+            
+            if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
+                dict["platform"] = "win32";
+            } else if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) {
+                dict["platform"] = "posix";
+            } else {
+                dict["platform"] = "darwin";
```
slozier: From a future compatibility point of view, not sure if darwin should be the fallback. I know it's just the 3 platforms right now, but in the future who knows what might get added to .NET Core (or if we support .NET Standard there might be even more platforms). Anyway, I'd make it an explicit check for MacOS and maybe fallback to win32, or unknown, or maybe even cli?

slide:
I thought about that, but didn't think of a good fallback, I think cli is probably good

PR #431, opened 2018-09-17
Implement List.copy, cmath.isfinite, array.typecodes

rtzoeller:
These methods/constants were added prior to Python 3.4.


Src/IronPython/Modules/array.cs
```diff
@@ -28,6 +28,8 @@ public static class ArrayModule {
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2104:DoNotDeclareReadOnlyMutableReferenceTypes")]
         public static readonly PythonType/*!*/ ArrayType = DynamicHelpers.GetPythonTypeFromType(typeof(array));
 
+        public static readonly string typecodes = "cbBuHhiIlLfd";
```
slozier: Python 3 doesn't include the `c` typecode. Did you include it because we're currently implementing it?

rtzoeller: Yes; the string was taken directly from the list of characters in `CreateData()`.

On a related note, we'll eventually need to implement `q` and `Q`.

PR #438, opened 2018-09-25
Add array support for `q` and `Q` typecodes, update renamed methods

### LINKED ISSUE #433: Add array support for typecodes q and Q
Add `array` support for typecodes `q` and `Q` (Python 3.3) and also remove `c` (Python 3.0).

rtzoeller:
Resolves #433.

There are quite a few test failures still occurring in `test_array.py`, but it doesn't look like any of them are a direct result of this. `test_tofromstring()` fails with an incorrect number of warnings thrown for calling the newly deprecated methods, but this is an implementation detail and has changed [within CPython](https://github.com/python/cpython/commit/1eb32c20459b87ac42d6aba5d9eade8f9d129893#diff-ef3c1f502fc0bbbb7cef0e1284ed1f3f) following 3.4 to match

### DIFF
--- a/Src/IronPython/Modules/array.cs
+++ b/Src/IronPython/Modules/array.cs
@@ -16,6 +16,7 @@ using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
 using IronPython.Runtime;
+using IronPython.Runtime.Exceptions;
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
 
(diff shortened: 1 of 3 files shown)


slide:
Why close this? I think there were just a couple of recommendations from @slozier. It looked good otherwise!

rtzoeller:
I ran into some implementation difficulties/broken functionality when trying to clean it up, and wanted to make sure the pieces were all there prior to submitting any updates. I don't think I'll have time to do that in the next few days, so I closed it for now. I'll reopen it when I have @slozier's recommendations (which I agree with) in and the module works again.

PR #448, opened 2018-10-07
Added a surrogateescape handler.

ShahneRodgers:
The surrogateescape handler was introduced for Python 3.1 in PEP-383 (https://www.python.org/dev/peps/pep-0383/) and is well described with useful links here: (https://vstinner.github.io/pep-383.html). Most of the UNIX tests were previously falling over due to a "LookupError: unknown error handler name 'surrogateescape'" which this commit will hopefully fix.


ShahneRodgers:
I'm not too sure if I've implemented the surrogate escape handler correctly. If it seems like this might be worthwhile, let me know and I'll add tests and a decoding handler too.

slide:
There's actually another pr with this as well.

ShahneRodgers:
My bad. Sorry about that.

PR #449, opened 2018-10-07
Get test_regressions passing

slozier:


slozier:
Depends on https://github.com/IronLanguages/ironpython3/pull/453

Tests/test_regressions.py
```diff
@@ -705,17 +700,18 @@ def g():
             return (l1, l2)
 
         t1, t2 = f()
-        self.assertEqual(t1.keys(), ['x', 'g'])
+        self.assertEqual(set(t1.keys()), {'x', 'g'})
         self.assertEqual(t2, {})
 
+    @unittest.skip("TODO: import cp20472 is not failing, figure out if the original issue is still relevant")
```
slide: This test is not valid anymore. I believe the issue was with non-ASCII characters in the source code, which is now ok in Python 3.

slozier: I think it's passing because invalid characters in comments are ok? If we were to move the characters to a string instead, then it would probably fail since we're not specifying the encoding (and the default is utf-8).

PR #452, opened 2018-10-16
Drop support of .NET Core 2.0

slozier:
Drop support for .NET Core 2.0 since it reached EOL on Oct 1st.

### DIFF
--- a/Src/IronPython.Modules/IronPython.Modules.csproj
+++ b/Src/IronPython.Modules/IronPython.Modules.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netcoreapp2.0;netcoreapp2.1</TargetFrameworks>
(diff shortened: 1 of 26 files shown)


Src/IronPython.Modules/IronPython.Modules.csproj
```diff
@@ -14,7 +14,7 @@
     <PackageReference Include="Mono.Posix" Version="5.4.0.201" Condition=" '$(Mono)' != 'true' " />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" Condition=" '$(TargetFramework)' == 'netcoreapp2.0' OR '$(TargetFramework)' == 'netcoreapp2.1' " />
```
slide: Can we just change this to something like

```
Condition=" $(TargetFramework.StartsWith('netcoreapp')) "
```

So we don't have to keep modifying this with new versions?

slozier: Sure, we can always change them back to specific versions if we find it no longer works in the future.

PR #455, opened 2018-10-17
Rename List to PythonList

slozier:
Makes it more consistent with other types (e.g. PythonDictionary, PythonTuple).


ShahneRodgers:
This rename seems like a great idea to me, but in case you haven't already noticed, the IronPython.Src.Scripts.test_cgcheck is now failing:

> generate (check-only) /Users/vsts/agent/2.140.2/work/1/s/Src/IronPython/Runtime/Types/TypeCache.Generated.cs ... different!
2018-10-17T03:03:17.6970310Z     generated file saved as: /Users/vsts/agent/2.140.2/work/1/s/Src/IronPython/Runtime/Types/TypeCache.Generated.cs.diff

PR #475, opened 2018-10-29
Fix #474

### LINKED ISSUE #474: binder gets stuck in a loop
It looks like the binder gets stuck in a loop when the following is run:
```Python
class C(object):
    pass

c = C()
C.__getattribute__ = lambda self, x: 0
c.x
del C.__getattribute__
c.x
```

slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/474. Looks like it was caused by https://github.com/IronLanguages/ironpython3/pull/141 which incorrectly refactored the code to remove the old style check. The test case will be covered by `test_attribute_customize`.

### DIFF
--- a/Src/IronPython/Modules/_bytesio.cs
+++ b/Src/IronPython/Modules/_bytesio.cs
@@ -275,13 +275,9 @@ namespace IronPython.Modules {
                 }
             }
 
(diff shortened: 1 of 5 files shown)


slozier:
Currently failing due to https://github.com/IronLanguages/ironpython2/issues/546.

PR #476, opened 2018-10-30
Properly implement hashing of complex numbers

### LINKED ISSUE #436: Properly implement hashing of complex numbers
Hashing of complex number doesn't match CPython (for non-real numbers). See https://docs.python.org/3/library/stdtypes.html#numeric-hash:
>For a complex number z, the hash values of the real and imaginary parts are combined by computing `hash(z.real) + sys.hash_info.imag * hash(z.imag)`, reduced modulo `2**sys.hash_info.width` so that it lies in `range(-2**(sys.hash_info.width - 1), 2**(sys.hash_info.width - 1))`. Again, if the result is `-1`, it’s replaced with `-2`.

Related to: https://github.com/IronLanguages/ironpython3/pull/435

gfmcknight:
Resolves #436.

Implement hashing of complex numbers that adheres to CPython's
standard. Add a unit test verifying that the hash uses both
the real and imaginary components to verify the quality of the
hash.

Some things I was unsure about:
- The code path when the complex's imaginary value is unnecessary (181-183), but I left it thinking that it makes clear that the value is essentially coerced to a double.
- Line 185 is excessively long, but it seems as though the convention for this repository is to keep statements on a single line, regardless the length.

### DIFF
--- a/Src/IronPython/Runtime/Operations/ComplexOps.cs
+++ b/Src/IronPython/Runtime/Operations/ComplexOps.cs
@@ -10,6 +10,7 @@ using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
 using IronPython.Runtime.Types;
+using IronPython.Modules;
 
 namespace IronPython.Runtime.Operations {
     public class ExtensibleComplex : Extensible<Complex> {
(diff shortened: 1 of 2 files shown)


slozier:
Thanks for the PR!

PR #479, opened 2018-11-05
Update NUnit

gpetrou:


NuGet.config
```diff
@@ -5,6 +5,5 @@
   </config>
   <packageSources>
     <add key="nuget.org" value="https://api.nuget.org/api/v3/index.json" />
-    <add key="nunit.myget.org" value="https://www.myget.org/F/nunit/api/v3/index.json" />
```
slide: I'd like to keep this in here, it should be a secondary source, so we should get most packages from the official nuget.

gpetrou: Oh OK. I thought it was something temporary.

PR #480, opened 2018-11-10
Update set repr

slozier:


slide:
Is there already a test for this?

slozier:
Sort of covered by `test_setcomps` but it's disabled since it's failing some closure related stuff.

slozier:
I guess it'll also be covered by `test_set` since apparently I have to fix it...

PR #497, opened 2018-11-26
Fix overflow error with enumerate

### LINKED ISSUE #473: enumerate does not support start > int.MaxValue
```Python
list(enumerate([1], 1<<31))
```
throws an `OverflowError`.

AlexKubiesa:
This should fix #473 
- [x] Figure out why test_builtinfunc.py references #473 and fix the failing test

### DIFF
--- a/Src/IronPython/Runtime/Converter.cs
+++ b/Src/IronPython/Runtime/Converter.cs
@@ -336,23 +336,16 @@ namespace IronPython.Runtime {
             return res.HasValue;
         }
 
-        internal static bool TryConvertToIndex(object value, out object index) {
-            return TryConvertToIndex(value, true, out index);
-        }
-
         /// <summary>
-        /// Attempts to convert value into a index usable for slicing and return the integer
+        /// Attempts to convert value into an index usable for slicing and return the integer or BigInteger
         /// value.  If the conversion fails false is returned.
-        /// 
-        /// If throwOverflowError is true then BigInteger's outside the normal range of integers will
-        /// result in an OverflowError.
         /// </summary>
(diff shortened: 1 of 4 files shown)


slide:
Do you have a test that you could add to `test_regressions.py`?

AlexKubiesa:
Good point, I've added a test.

Src/IronPython/Runtime/Converter.cs
```diff
@@ -336,23 +336,16 @@ public static partial class Converter {
...
-        /// Attempts to convert value into a index usable for slicing and return the integer
+        /// Attempts to convert value into an index usable for slicing and return the integer or BigInteger
         /// value.  If the conversion fails false is returned.
-        /// 
-        /// If throwOverflowError is true then BigInteger's outside the normal range of integers will
-        /// result in an OverflowError.
         /// </summary>
-        internal static bool TryConvertToIndex(object value, bool throwOverflowError, out object index) {
-            index = ConvertToSliceIndexHelper(value, throwOverflowError);
+        internal static bool TryConvertToIndex(object value, out object index) {
```
slozier: Just did a quick search and it looks like this is called in a bunch of places. Does this mean that they will no longer throw?

AlexKubiesa: As far as I can see, this overload is only called in `Enumerate.cs`. The other overload is called in many places but I haven't touched that one.

slozier: Ah I see, the others use the int overload.

slozier:
Thanks!

Issue #502, opened 2018-12-02
Add default keyword argument to min/max


slozier:
Add `default` keyword argument to `min`/`max`. For example:
```Python
max((), default=1)
```

See https://docs.python.org/3/library/functions.html#max and https://docs.python.org/3/library/functions.html#min for details.

There are tests for this in `test_builtin`.

PR #510, opened 2019-01-03
add default keyword arg for min/max

alanmbarr:
#502 added default keyword arg

### DIFF
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -1106,7 +1106,6 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
                 throw PythonOps.TypeError("{1}() takes at most 2 keyword arguments ({0} given)", dict.Count, name);
             }
 
-            var keys = dict.Keys.Where(x => (string)x != "default" && (string)x != "key");
             if (dict.Count > cnt) {
                 var key = dict.Keys.Cast<string>().Where(x => x != "default" && x != "key").First();
                 throw PythonOps.TypeError("{1}() got an unexpected keyword argument ({0})", key, name);
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -901,16 +901,25 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
                 }
                 return ret;
             } else {
-                throw PythonOps.TypeError("max expecting 1 arguments, got 0");
+                throw PythonOps.TypeError("max expected 1 arguments, got 0");
             }
 
         }
 
         public static object max(CodeContext/*!*/ context, object x, [ParamDictionary]IDictionary<object, object> dict) {
             IEnumerator i = PythonOps.GetEnumerator(x);
-            if (!i.MoveNext())
+            
+            var kwargTuple = GetMaxKwArg(dict);
+            object method = kwargTuple.Item1;
+            object def = kwargTuple.Item2;
+
+            if (!i.MoveNext()) {
+                if(def != null) {
+                    return def;
+                }
                 throw PythonOps.ValueError(" max() arg is an empty sequence");
-            object method = GetMaxKwArg(dict);
+            }
+
             object ret = i.Current;
             object retValue = PythonCalls.Call(context, method, i.Current);
             PythonContext pc = context.LanguageContext;
@@ -925,18 +934,23 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
         }
 
         public static object max(CodeContext/*!*/ context, object x, object y, [ParamDictionary] IDictionary<object, object> dict) {
-            object method = GetMaxKwArg(dict);
+            var kwargTuple = GetMaxKwArg(dict);
+            object method = kwargTuple.Item1;
+
             PythonContext pc = context.LanguageContext;
             return pc.GreaterThan(PythonCalls.Call(context, method, x), PythonCalls.Call(context, method, y)) ? x : y;
         }
 
         public static object max(CodeContext/*!*/ context, [ParamDictionary]IDictionary<object, object> dict, params object[] args) {
+            var kwargTuple = GetMaxKwArg(dict);
+            object method = kwargTuple.Item1;
+            object def = kwargTuple.Item2;
             if (args.Length > 0) {
                 int retIndex = 0;
                 if (args.Length == 1) {
                     return max(context, args[retIndex], dict);
                 }
-                object method = GetMaxKwArg(dict);
+                
                 object retValue = PythonCalls.Call(context, method, args[retIndex]);
                 PythonContext pc = context.LanguageContext;
                 for (int i = 1; i < args.Length; i++) {
@@ -948,13 +962,13 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
                 }
                 return args[retIndex];
             } else {
-                throw PythonOps.TypeError("max expecting 1 arguments, got 0");
+                throw PythonOps.TypeError("max expected 1 arguments, got 0");
             }
         }
 
-        private static object GetMaxKwArg(IDictionary<object, object> dict) {
-            if (dict.Count != 1)
-                throw PythonOps.TypeError(" max() should have only 1 keyword argument, but got {0} keyword arguments", dict.Count);
+        private static Tuple<object, object> GetMaxKwArg(IDictionary<object, object> dict) {
+            if (dict.Count != 1 && dict.Count != 2)
+                throw PythonOps.TypeError(" max() should have only 2 keyword arguments, but got {0} keyword arguments", dict.Count);
 
             return VerifyKeys("max", dict);
         }
@@ -989,15 +1003,23 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
                 }
                 return ret;
             } else {
-                throw PythonOps.TypeError("min expecting 1 arguments, got 0");
+                throw PythonOps.TypeError("min expected 1 arguments, got 0");
             }
         }
 
         public static object min(CodeContext/*!*/ context, object x, [ParamDictionary]IDictionary<object, object> dict) {
             IEnumerator i = PythonOps.GetEnumerator(x);
-            if (!i.MoveNext())
+            var kwargTuple = GetMinKwArg(dict);
+            object method = kwargTuple.Item1;
+            object def = kwargTuple.Item2;
+
+            if (!i.MoveNext()) {
+                if (def != null) {
+                    return def;
+                }
                 throw PythonOps.ValueError(" min() arg is an empty sequence");
-            object method = GetMinKwArg(dict);
+            }
+
             object ret = i.Current;
             object retValue = PythonCalls.Call(context, method, i.Current);
             PythonContext pc = context.LanguageContext;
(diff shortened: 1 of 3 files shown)


Src/IronPython/Modules/Builtin.cs
```diff
@@ -901,16 +901,25 @@ public class mapobject : IEnumerable {
...
         public static object max(CodeContext/*!*/ context, object x, [ParamDictionary]IDictionary<object, object> dict) {
             IEnumerator i = PythonOps.GetEnumerator(x);
-            if (!i.MoveNext())
+            
+            var kwargTuple = GetMaxKwArg(dict);
+            object method = kwargTuple.Item1;
+            object def = kwargTuple.Item2;
+
+            if (!i.MoveNext()) {
+                if(def != null) {
```
slozier: Might need to rework this a bit since `None` is a valid default.

Src/IronPython/Modules/Builtin.cs
```diff
@@ -925,18 +934,23 @@ public class mapobject : IEnumerable {
         }
 
         public static object max(CodeContext/*!*/ context, object x, object y, [ParamDictionary] IDictionary<object, object> dict) {
-            object method = GetMaxKwArg(dict);
+            var kwargTuple = GetMaxKwArg(dict);
+            object method = kwargTuple.Item1;
```
slozier: Not something I'd block the PR on, but `GetMaxKwArg` could use an argument to check to make sure default is allowed, and if not throw:
> TypeError: Cannot specify a default for max() with multiple positional arguments

Src/IronPython/Modules/Builtin.cs
```diff
@@ -925,18 +934,23 @@ public class mapobject : IEnumerable {
...
+
             PythonContext pc = context.LanguageContext;
             return pc.GreaterThan(PythonCalls.Call(context, method, x), PythonCalls.Call(context, method, y)) ? x : y;
         }
 
         public static object max(CodeContext/*!*/ context, [ParamDictionary]IDictionary<object, object> dict, params object[] args) {
+            var kwargTuple = GetMaxKwArg(dict);
+            object method = kwargTuple.Item1;
+            object def = kwargTuple.Item2;
```
slozier: default is not used here?

Tests/test_builtinfunc.py
```diff
@@ -317,6 +317,15 @@ def test_max_min(self):
         self.assertTrue(min((1,2,3)) == 1)
         self.assertTrue(min(1,2,3) == 1)
 
+        self.assertTrue(max([], default=1) == 1)
+        self.assertTrue(max((), default=1) == 1)
```
slozier: Maybe add `self.assertTrue(max(), default=None) is None)` to the test cases.

Src/IronPython/Modules/Builtin.cs
alanmbarr: Seems like with python 3 i get the error below. I will try to trigger it another way. If there are some tests cases I would be glad to try to fix this section.
TypeError: Cannot specify a default for max() with multiple positional arguments
>>> max([], True, True, key=lambda x: x, default=1)

alanmbarr: I found some test cases in the main repo. i'll take a look

Src/IronPython/Modules/Builtin.cs
```diff
@@ -925,36 +934,55 @@ public class mapobject : IEnumerable {
...
+                
                 object retValue = PythonCalls.Call(context, method, args[retIndex]);
                 PythonContext pc = context.LanguageContext;
                 for (int i = 1; i < args.Length; i++) {
                     object tmpRetValue = PythonCalls.Call(context, method, args[i]);
-                    if (pc.GreaterThan(tmpRetValue, retValue)) {
+                    var tmptype = tmpRetValue.GetType();
+                    var rettype = retValue.GetType();
+
+                    if ( tmptype == typeof(bool) && rettype == typeof(PythonList) ) {
```
slozier: What are these checks for? Are they to get specific tests passing? If so the checks should be removed and the tests disabled.

I'm not sure if I've filed an issue already, but comparisons between uncomparable types need to be removed. This means that these cases should fail properly without special handling.

Src/IronPython/Modules/Builtin.cs
```diff
@@ -925,36 +934,55 @@ public class mapobject : IEnumerable {
...
+                throw PythonOps.TypeError("max expected 1 arguments, got 0");
             }
         }
 
-        private static object GetMaxKwArg(IDictionary<object, object> dict) {
-            if (dict.Count != 1)
-                throw PythonOps.TypeError(" max() should have only 1 keyword argument, but got {0} keyword arguments", dict.Count);
+        private static Tuple<object, object> GetMaxKwArg(IDictionary<object, object> dict, bool isDefaultAllowed) {
+            if (dict.Count != 1 && dict.Count != 2)
+                throw PythonOps.TypeError(" max() should have only 2 keyword arguments, but got {0} keyword arguments", dict.Count);
```
slozier: Nit: I don't think we need a space at the start of the TypeError strings

Src/IronPython/Modules/Builtin.cs
```diff
@@ -1037,27 +1098,44 @@ public class mapobject : IEnumerable {
...
-                ICollection<object> keys = dict.Keys;
-                IEnumerator<object> en = keys.GetEnumerator();
-                if (en.MoveNext()) {
-                    throw PythonOps.TypeError(" {1}() got an unexpected keyword argument ({0})", en.Current, name);
-                }
+        private static Tuple<object, object> VerifyKeys(string name, IDictionary<object, object> dict) {
+            object method;
+            object def;
+
+            var MyCopiedDictionary = new Dictionary<object, object>(dict);
```
slozier: Would be nice to avoid the extra dictionary allocation here. You can probably do something like:
```
int cnt = 0;
if (dict.TryGetValue(...)) {
    cnt++;
}

if (dict.Count > cnt) {
    throw ...
}
```

Tests/test_builtinfunc.py
```diff
@@ -308,17 +310,118 @@ def test_unichr(self):
...
+            except TypeError:
+                pass
+            else:
+                self.fail(stmt)
+
+        self.assertEqual(max((1,), key=neg), 1)     # one elem iterable
+        self.assertEqual(max((1,2), key=neg), 1)    # two elem iterable
+        self.assertEqual(max(1, 2, key=neg), 1)     # two elems
+        self.assertEqual(max((), default=None), None)    # zero elem iterable
+        #self.assertEqual(min((1,), default=None), 1)     # one elem iterable
```
slozier: Any reason why these cases are commented out?

slozier:
Sorry for all the comments, hopefully you don't mind. :smile:

Tests/test_builtinfunc.py
alanmbarr: I am receiving TypeError: NoneType is not callable from
object retValue = PythonCalls.Call(context, method, args[retIndex]);
Not sure how to deal with these tests cases

slozier:
I had to fix up your PR since you had some merge issues in there. You will probably need to re-pull your master from github.

Src/IronPython/Modules/Builtin.cs
```diff
@@ -989,21 +1006,29 @@ public class mapobject : IEnumerable {
...
+                throw PythonOps.ValueError("min() arg is an empty sequence");
+            }
+
             object ret = i.Current;
             object retValue = PythonCalls.Call(context, method, i.Current);
             PythonContext pc = context.LanguageContext;
             while (i.MoveNext()) {
                 object tmpRetValue = PythonCalls.Call(context, method, i.Current);
-
+
```
slozier: nit: empty space

Src/IronPython/Modules/Builtin.cs
```diff
@@ -1013,51 +1038,74 @@ public class mapobject : IEnumerable {
...
                 }
-                object method = GetMinKwArg(dict);
+                
                 object retValue = PythonCalls.Call(context, method, args[retIndex]);
                 PythonContext pc = context.LanguageContext;
 
                 for (int i = 1; i < args.Length; i++) {
                     object tmpRetValue = PythonCalls.Call(context, method, args[i]);
-
+
```
slozier: nit: empty space

Src/IronPython/Modules/Builtin.cs
```diff
@@ -1013,51 +1038,74 @@ public class mapobject : IEnumerable {
...
+            if (dict.TryGetValue("default", out def)) {
+                cnt++;
+            }
+
+            if (dict.Count > 2) {
+                throw PythonOps.TypeError("{1}() takes at most 2 keyword arguments ({0} given)", dict.Count, name);
+            }
+
+            var keys = dict.Keys.Where(x => (string)x != "default" && (string)x != "key");
```
slozier: Almost there, maybe something like:
```C#

if (dict.Count > cnt) {
    var key = dict.Keys.Cast<string>().Where(x => x != "default" && x != "key").First();
    throw PythonOps.TypeError("{1}() got an unexpected keyword argument ({0})", key, name);
}
```

slide:
@alanmbarr It looks like the merge of master brought back some files that we don't want to merge in, there should only be 3 changed files in your PR. @slozier did some work to rework the history in his force push to your branch, maybe he can outline what he did so you can replicate it to rework the history and remove the Generated files.

alanmbarr:
i'll take a look

Tests/test_builtinfunc.py
slozier: Alright, this one is a bit tricky. The previous code assumed that if you got to the `Call` then it was because `key` was a keyword argument. However in this case we only have `default` and `key` is undefined. It'll require a bit more work to get this working. Off the top of my head you could do something like:
```C#
public static object max(CodeContext/*!*/ context, object x, [ParamDictionary]IDictionary<object, object> dict) {
    // ...

    if (!i.MoveNext()) {
        if (def != UndefinedKeywordArgument) return def;
        throw PythonOps.ValueError("max() arg is an empty sequence");
    }

    if (method == UndefinedKeywordArgument) {
        return max(context, x);
    }

    // ...
}

private static object UndefinedKeywordArgument = new object();

private static Tuple<object, object> VerifyKeys(string name, IDictionary<object, object> dict) {
    object method;
    object def;

    int cnt = 0;
    if (dict.TryGetValue("key", out method)) {
        cnt++;
    } else {
        method = UndefinedKeywordArgument;
    }

    if (dict.TryGetValue("default", out def)) {
        cnt++;
    } else {
        def = UndefinedKeywordArgument;
    }

    // ...
}
```

Or something else so you can know if it's undefined...

Tests/test_builtinfunc.py
```diff
@@ -308,17 +310,118 @@ def test_unichr(self):
...
+        #self.assertEqual(min((1,2), default=None), 1)    # two elem iterable
+        self.assertEqual(max((), default=1, key=neg), 1)
+        self.assertEqual(max((1, 2), default=3, key=neg), 1)
+
+        #self.assertEqual(min((1, 2), key=None), 1)
```
slozier: I think you can get rid of this line since it also fails with CPython.

slozier:
@alanmbarr In case you're interested in how I cleaned up your PR, basically what I did was:
> git rebase -i upstream/master

where I droped all the commits causing issues (specifically the one called "merge"). Then I force pushed the new branch back to github.

> git push origin -f master

In this case `upstream` is `https://github.com/IronLanguages/ironpython3` and `origin` is `https://github.com/alanmbarr/ironpython3`

Src/IronPython/Modules/Builtin.cs
slozier: You can remove this line since keys is no longer used.

alanmbarr: The whole block?

slozier: Just the line with `var keys = ...`. Unless it does something and I'm not seeing it?

alanmbarr: My only concern here is being able to throw the error about the first extra key. if it is unnecessary I can take it out?

slozier: `var key = ...` inside the `if` does that. The `var keys = ...` before the `if` is redundant.

alanmbarr: ah ok I confused myself. thanks!

alanmbarr:
Thanks for your help. Let me know if I need to rebase my commits into one if that makes it easier for you.

Src/IronPython/Modules/Builtin.cs
```diff
@@ -925,36 +938,44 @@ public class mapobject : IEnumerable {
...
+                throw PythonOps.TypeError("max expected 1 arguments, got 0");
             }
         }
 
-        private static object GetMaxKwArg(IDictionary<object, object> dict) {
-            if (dict.Count != 1)
-                throw PythonOps.TypeError(" max() should have only 1 keyword argument, but got {0} keyword arguments", dict.Count);
+        private static Tuple<object, object> GetMaxKwArg(IDictionary<object, object> dict, bool isDefaultAllowed) {
+            if (dict.Count != 1 && dict.Count != 2)
+                throw PythonOps.TypeError("max() should have only 2 keyword arguments, but got {0} keyword arguments", dict.Count);
```
slide: nit: use a $ string? 

This is not required, just a possible option.

PR #514, opened 2019-01-09
Parser cleanup

slozier:
Update parts of parser to more closely match https://docs.python.org/3.4/reference/grammar.html

### DIFF
--- a/Src/IronPython/Compiler/Parser.cs
+++ b/Src/IronPython/Compiler/Parser.cs
@@ -495,7 +495,7 @@ namespace IronPython.Compiler {
 
         // del_stmt: "del" exprlist
         //  for error reporting reasons we allow any expression and then report the bad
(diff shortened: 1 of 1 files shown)


Src/IronPython/Compiler/Parser.cs
```diff
@@ -2401,40 +2308,41 @@ struct WithItem {
...
             } else {
-                return ParseTestListAsExprError();
+                return ParseTestListError();
             }
-        }
 
-        private Expression ParseTestListAsExpr(Expression expr) {
-            List<Expression> l = new List<Expression>();
-            l.Add(expr);
+            Expression ParseTestList(Expression expr) {
```
slide: It's hard to tell with the formatting, but is this a nested method now? If not, why was the `private` removed.

slide: No need to rename, it was just hard to tell from the diff.

PR #521, opened 2019-01-25
Changes from ipy2

slozier:


slide:
Hmmm, looks like RuntimeInformation is not available for some reason.

slozier:
I should really try to build locally before I do a PR.

slide:
It's what we have CI for :-)

Issue #304, opened 2017-05-14
coding20731.py has different line endings


gpetrou:
Hello,

I forked the current repository and cloned it to a Windows folder.
It seems that Src\StdLib\Lib\test\coding20731.py has different line endings than the rest of the files, so git complains that the file is changed and should be committed. Is this on purpose or should be fixed? There is a similar issue with some DLR files. I will open an issue to that repository as well.

PR #531, opened 2019-02-04
Normalize line endings of text files in the repository (#304)

BCSharp:
Line endings for text files in the git repository normalized to LF, following rules in .gitattibutes (Issue #304).

Note that files in the working directory that were already checked out before may not be updated. If it happens, the best way to fix it is:
 * make the working directory clean (no modified file)
 * delete all files in the working directory (except .git)
 * check out again a fresh working directory.


slide:
My only hangup on this is the file from the stdlib. If the file matches the line endings in the C Python repo, then we need to maintain the file as is.

slozier:
@slide I checked locally and coding20731.py is CRLF when installing on Windows so it should be fine.

slozier:
Hmm, nvm my last comment, in the CPython repo it's always CRLF. Looking at their [.gitattributes](https://github.com/python/cpython/blob/master/.gitattributes):
```
Lib/test/cjkencodings/* -text
Lib/test/decimaltestdata/*.decTest -text
Lib/test/test_email/data/*.txt -text
Lib/test/xmltestdata/* -text
Lib/test/coding20731.py -text
Lib/test/test_importlib/data01/* -text
```

BCSharp:
Indeed. But in this case I think we should take over all of those .gitignore specs and make sure all affected files are checked in exactly in the same way as in CPython, do you agree? Otherwise, we may have a situation that some tests (enabled in the future) pass on one platform but fail on another.

On a side note, I see that CPython declares *.ps1 files as always CRLF; IronPython had makefile.ps1 as always LF. I have made .ps1 files auto, which makes more sense in times of portable PowerShell.

slozier:
I agree, we should probably add `Src/StdLib/.gitattributes` with the same specs as CPython. The StdLib comes from CPython 3.4.

BCSharp:
They added those .gitatributes [fairly recently, in 3.7](https://github.com/python/cpython/pull/233) and then backported all the way to 3.5 and 2.7.

BCSharp:
I took over those for .sln, .proj as well (although there are no such files in this location), just to prevent any surprises in the future.

slozier:
Thanks for the PR!

PR #535, opened 2019-02-08
 Enable test_codecs and test_unicode

slozier:
@slide Sorry for yet another massive PR.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -175,6 +175,15 @@ namespace IronPython.Runtime.Operations {
             throw PythonOps.LookupError("unknown encoding: {0}", encoding);
         }
 
+        internal static PythonTuple LookupTextEncoding(CodeContext/*!*/ context, string encoding, string alternateCommand) {
+            var tuple = LookupEncoding(context, encoding);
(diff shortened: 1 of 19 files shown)


Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -175,6 +175,15 @@ public static partial class PythonOps {
             throw PythonOps.LookupError("unknown encoding: {0}", encoding);
         }
 
+        internal static PythonTuple LookupTextEncoding(CodeContext/*!*/ context, string encoding, string alternateCommand) {
+            var tuple = LookupEncoding(context, encoding);
+            if (TryGetBoundAttr(tuple, "_is_text_encoding", out object isTextEncodingObj)
+                    && isTextEncodingObj is bool isTextEncoding && !isTextEncoding) {
+                throw LookupError("'{0}' is not a text encoding; use {1} to handle arbitrary codecs", encoding, alternateCommand);
```
slide: nit: $ string?

slozier: The only thing that annoys me with using string interpolation in this case is that LookupError also calls `string.Format`. Maybe I should just add the argument-less overloads (which doesn't call `string.Format`) and be done with it...

PR #536, opened 2019-02-10
Only use __lt__ in list.sort

slozier:
Prep work to eliminate `__cmp__`.


slide:
Can you highlight what tests we have that cover this scenario already?

slozier:
You mean tests to make sure sorting still works? We're running [`CPythonTests.test_sort`](https://github.com/IronLanguages/ironpython3/blob/master/Src/StdLib/Lib/test/test_sort.py).

Issue #537, opened 2019-02-11
str() on bytes object does not produce an ASCII string


BCSharp:
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description

The `str()` conversion of a `bytes` object should produce the same string as `repr()` on the same object. This is CPython 3.4 behavior, but not IronPython 3 behavior.

The same applies to `bytearray`.

### Steps to Reproduce

Run code:

```python
print(b"\xd0\x9f\xd0\xb8\xd1\x82\xd0\xbe\xd0\xbd")
```

**Expected behavior:**

`b'\xd0\x9f\xd0\xb8\xd1\x82\xd0\xbe\xd0\xbd'` printed

**Actual behavior:**

`Ð�Ð¸Ñ�Ð¾Ð½` printed (can be a different output depending on current console encoding)

### Versions

You can get this information from executing `ipy -V`.
IronPython 3.4.0 3.4.0.0 on .NET 4.0.30319.42000
Python 3.4.4

slozier:
I already have a PR that includes a fix for this lined up. Just waiting for https://github.com/IronLanguages/ironpython3/pull/535.

BCSharp:
@slozier too bad, I have fixed it too... :)

BCSharp:
Nevertheless, it was a good exercise in getting to know the internals. Would this have been correct?:

`StringOps.cs:`
```python
        [StaticExtensionMethod]
        public static object __new__(CodeContext/*!*/ context, PythonType cls,
            [BytesConversion]IList<byte> @object,
            string encoding=null,
            string errors=null) {

            if (cls == TypeCache.String) {
                if (encoding == null && errors == null) {
                    if (@object is ByteArray ba) return ba.__repr__(context);
                    else return ((Bytes)@object).__repr__(context);
                } else {
                    return (@object as Bytes ?? (new Bytes(@object))).decode(context, encoding ?? "utf-8", errors ?? "strict");
                }
            } else {
                return cls.CreateInstance(context, __new__(context, TypeCache.String, @object, encoding, errors));
            }
        }
```

slozier:
Your fix probably would have worked for most cases, I did it slightly differently (by adding an overload instead):
```C#
        [StaticExtensionMethod]
        public static object __new__(CodeContext/*!*/ context, PythonType cls, [BytesConversion]IList<byte> @object) {
[...]

Issue #470, opened 2018-10-27
assertRaises does not properly catch .NET Exceptions


slozier:
First test fails, second one passes:
```Python
def test_assertraises(self):
    import System
    with self.assertRaises(System.ArgumentException):
        System.Char.MaxValue.CompareTo("a")

def test_tryexcept(self):
    import System
    try:
        System.Char.MaxValue.CompareTo("a")
    except System.ArgumentException:
        pass
```

This might only be for exceptions that map to other types, in this case `ArgumentException` maps to `ValueError`.

PR #538, opened 2019-02-11
Enable test_bytes

slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/537 among other things.

Also resolves https://github.com/IronLanguages/ironpython3/issues/470


slozier:
Hmm, lots of failures, thought I ran the test suite for this. I guess there's more work to be done.

PR #558, opened 2019-03-01
 Fix PythonSurrogateEscapeEncoding to support block-wise conversion 

BCSharp:
My understanding of the meaning of the `flush` parameter in the functions `GetByteCount`, `GetBytes`  was incorrect (class `System.Text.Encoder`). This looked to me as an additional `Reset()` at the end of processing, but careful reading of the documentation reveals that it is more like `lastBlock` flag.

The [documentation of `GetByteCount`](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoder.getbytecount?view=netframework-4.5) reads:

> **`flush`**    Boolean 
> `true` to simulate clearing the internal state of the encoder after the calculation; otherwise, `false`.

The [documentation of `GetBytes`](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoder.getbytes?view=netframework-4.5) reads:

> **`flush`**   Boolean 
> `true` to clear the internal state of the encoder after the conversion; otherwise, `false`.

Additionally, remarks section contains:

>The application should call GetByteCount on a block of data immediately before calling GetBytes on the same block, so that any trailing characters from the previous block are included in the calculation.

From the source code of `GetBytes`:

>// The flush parameter indicates whether
            // the encoder should flush any shift-states and partial characters at the
            // end of the conversion. To ensure correct termination of a sequence of
            // blocks of encoded bytes, the last call to GetBytes should specify
            // a value of true for the flush parameter.

So the proper protocol of block-wise encoding is:

```C#
... = encoder.GetByteCount(blockA, ..., flush: false)
... = encoder.GetBytes(blockA, ..., flush: false)
... = encoder.GetByteCount(blockB, ..., flush: false)
... = encoder.GetBytes(blockB, ..., flush: false)
...
...
... = encoder.GetByteCount(blockZ, ..., flush: true)
... = encoder.GetBytes(blockZ, ..., flush: true)
```
It is the same story for `Decoder`.

In this PR I have corrected `PythonSurrogateEscapeEncoding` implementation and added some test cases to cover block-wise encoding and decoding.

Further this PR contains corrected checks (and test cases) for big-endian encodings and some minor performance improvements.


BCSharp:
@slozier Please hold off merging of this PR for now. Commit 16cabcb broke handling of UTF-8.

slozier:
@BCSharp Alright, interesting that no tests picked up the failure.

BCSharp:
Well, the tests that would have picked up this regression are in `CPython.test_codecs`, which is being skipped right now. And it is still a way before it can be enabled, so I think I will add equivalent tests to my NUnit test suite to prevent further regressions until we can run `test_codecs`.

slozier:
In case you're looking for some test cases: `https://github.com/PythonCharmers/python-future/blob/master/tests/test_future/test_surrogateescape.py`

Src/IronPython/Runtime/PythonSurrogateEscapeEncoding.cs
```diff
@@ -16,33 +16,58 @@ internal class PythonSurrogateEscapeEncoding : Encoding {
         private const char Pass1SurrogateMarker = '?';
         private const char Pass2SurrogateMarker = '-';
 
+        public int CharacterWidth { get; }
+        public bool IsBigEndian { get; } // meaningul only for wide-char encodings
```
slozier: typo in comment

slozier:
@BCSharp Let me know if you're happy with the state of your PR and I'll merge.

BCSharp:
`netcoreapp2.1.IronPython.test_regressions` is failing on Windows CI, but is passing on my machine.

Src/IronPython/Runtime/PythonSurrogateEscapeEncoding.cs
BCSharp: done

slozier:
@BCSharp Failure is unrelated to your PR.

BCSharp:
Then the PR is ready for merging from my side.

slozier:
Thanks again for the PR! If you're interested in another encoding related issue, then `surrogatepass` would be a useful one.

PR #567, opened 2019-03-07
Enable some tests

slozier:


slozier:
Note that this breaks the following interop:
```
b = bytes(range(5))
System.Array[System.Byte](b) == b
```

PR #574, opened 2019-03-14
Make PythonAsciiEncoding compliant with System.Text.Encoding protocol

BCSharp:
This is primarily to enable `PythonAsciiEncoding` to work with the `PythonEncoding` family (`PythonSurrogateEscapeEncoding`, `PythonSurrogatePassEncoding`, and future user-registered encoding error handlers).

The old `PythonAsciiEncoding` was assuming that all fallbacks were stateless. With this change, the workaround for `surrogateescape` to downgrade `PythonAsciiEncoding` to .NET `Encoding.ASCII` is not necessary anymore.

Also, some tests are added to test workings of `PythonAsciiEncoding` with `PythonSurrogateEscapeEncoding`.

### DIFF
--- a/Src/IronPython/Runtime/PythonAsciiEncoding.cs
+++ b/Src/IronPython/Runtime/PythonAsciiEncoding.cs
@@ -27,10 +27,10 @@ namespace IronPython.Runtime {
         }
 
         internal PythonAsciiEncoding(EncoderFallback encoderFallback, DecoderFallback decoderFallback)
-#if NETCOREAPP2_1
+#if !NET45
+            // base(0, encoderFallback, decoderFallback) publicly accessible only in .NET Framework 4.6 and up
             : base(0, encoderFallback, decoderFallback) {
 #else
-            // base(0, encoderFallback, decoderFallback) publicly accessible only in .NET Framework 4.6 and up
             : base() {
             // workaround for lack of proper base constructor access - implementation dependent
             typeof(Encoding).GetField("encoderFallback", BindingFlags.Instance | BindingFlags.NonPublic).SetValue(this, encoderFallback);
@@ -38,7 +38,7 @@ namespace IronPython.Runtime {
 #endif
         }
 
-    internal static Encoding MakeNonThrowing() {
+        internal static Encoding MakeNonThrowing() {
             Encoding enc;
 #if FEATURE_ENCODING
             enc = new PythonAsciiEncoding(new NonStrictEncoderFallback(), new NonStrictDecoderFallback());
--- a/Src/IronPython/Runtime/PythonAsciiEncoding.cs
+++ b/Src/IronPython/Runtime/PythonAsciiEncoding.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Runtime.Serialization;
 using System.Text;
+using System.Reflection;
 
 using IronPython.Runtime.Operations;
 
@@ -17,6 +18,7 @@ namespace IronPython.Runtime {
     /// </summary>
     [Serializable]
     sealed class PythonAsciiEncoding : Encoding {
+        // Singleton (global) instances are readonly, so their fallbacks cannot be accidentally modified unless cloned
         internal static readonly Encoding Instance = MakeNonThrowing();
         internal static readonly Encoding SourceEncoding = MakeSourceEncoding();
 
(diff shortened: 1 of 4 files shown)


slozier:
@BCSharp Any idea what advantage `PythonAsciiEncoding` provides over plain `Encoding.ASCII`?

BCSharp:
`PythonAsciiEncoding` in its default fallback configuration (the `Instance`) will never throw any exceptions, even for bytes >0xff. Using the same fallbacks with `ASCII` will still result in exceptions due to the way how the .NET fallback machinery works. However, I believe that using  `cp1252` would have the same effect as `PythonAsciiEncoding`, if we could get it available on all platforms. This would also require significant regression testing.  

The second instance is used only for source file decoding, and when I look at how it is being used I think it could be abolished completely without much effort. I was planning to look into it after ‘surrogatepass’.  

So, in the end, I think we could get rid of `PythonAsciiEncoding` eventually.

Src/IronPython/Runtime/PythonAsciiEncoding.cs
```diff
@@ -17,44 +18,62 @@ namespace IronPython.Runtime {
...
         internal static readonly Encoding SourceEncoding = MakeSourceEncoding();
 
         internal PythonAsciiEncoding()
             : base() {
         }
 
-        internal static Encoding MakeNonThrowing() {
-            Encoding enc = new PythonAsciiEncoding();
+        internal PythonAsciiEncoding(EncoderFallback encoderFallback, DecoderFallback decoderFallback)
+#if NETCOREAPP2_1
```
slozier: Should be either `#if NETCOREAPP2_1 || NETSTANDARD2_0` or `#if !NET45`. Second one might be better so we don't have to mess with it when we add new targets.

slozier:
Just looking at the usage of `PythonAsciiEncoding` in `StringOps` it looks like we're replacing the fallback in the most cases anyway so the non-throwing fallback isn't of much use.

Src/IronPython/Runtime/PythonAsciiEncoding.cs
BCSharp: Will do.

BCSharp: Is `NET45` also defined for Mono? And if not, what is defined for Mono?

slozier: Yes, `NET45` is also for Mono. Looking at this again, why do we need reflection here? Can't we just do:
```C#
EncoderFallback = encoderFallback;
DecoderFallback = decoderFallback;
```

Src/IronPython/Runtime/PythonAsciiEncoding.cs
```diff
@@ -17,44 +18,62 @@ namespace IronPython.Runtime {
...
+#else
+            // base(0, encoderFallback, decoderFallback) publicly accessible only in .NET Framework 4.6 and up
+            : base() {
+            // workaround for lack of proper base constructor access - implementation dependent
+            typeof(Encoding).GetField("encoderFallback", BindingFlags.Instance | BindingFlags.NonPublic).SetValue(this, encoderFallback);
+            typeof(Encoding).GetField("decoderFallback", BindingFlags.Instance | BindingFlags.NonPublic).SetValue(this, decoderFallback);
+#endif
+        }
+
+    internal static Encoding MakeNonThrowing() {
```
slozier: nit: indent is off

Src/IronPython/Runtime/PythonAsciiEncoding.cs
BCSharp: This won’t work because every newly instantiated `Encoding` is read-only and it should stay so if used as a global object.

BCSharp:
> looks like we're replacing the fallback in the most cases

My working hypothesis is that `PythonAsciiEncoding` is a remnant from Python 2. It behaves like cp1253 (aka Latin-1 or Windows-1252), which is afaik Python's 2 default, but unlike the cp1252 encoding from .NET, it still allows for stateles fallbacks to be plugged in (and with this PR, also stateful fallbacks). While I can see a need for such encoding in Python 2, I don't see any compelling reason for it in Python 3. However, I would like to have `test_codecs` passing before attempting to remove `PythonAsciiEncoding`.

Src/IronPython/Runtime/PythonAsciiEncoding.cs
slozier: I see, that is quite unfortunate. Does Mono use the same private member names? Probably not much of an issue if we're going to (hopefully) drop `PythonAsciiEncoding`... Maybe we should just drop NET45 and make the minimum supported version NET461?

BCSharp: Yes, Mono has the same private names and unfortunately the desired constructor internal. I have changed the condition to `!NET45`, this will do for now. This seems to me a too small reason to drop .NET 4.5 support.

slozier:
@BCSharp Thanks for the PR!

PR #576, opened 2019-03-16
Correct NuGet API URL

BCSharp:
I have finally got the build working on my macOS, but it required a change in the NuGet API URL. The old URL points to a non-existent resource and I don't know how the build could have been working on Linux, CI, or on macOS for other people. Perhaps some clever rewriting rules that I don't have. Anyway, the correct URL for the packages index can be easily verified in a web browser.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,7 +4,7 @@
     <add key="globalPackagesFolder" value="./packages" />
   </config>
   <packageSources>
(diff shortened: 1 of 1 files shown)


slozier:
LGTM though I'm not sure we even need the `packageSources`. I'm pretty sure msbuild will pull from the proper source anyway.

slozier:
@BCSharp I'm glad you have a mac to test this stuff, I don't have one so it's hard to debug things.

PR #584, opened 2019-03-25
Add 'surrogatepass' codec error handler

BCSharp:
To perform _surrogatepass_ during decoding, the correct value of `index` of the fallback buffer is essential. Therefore I have added a workaround for the dotnet/corefx#29898 bug. The bug should be resolved in .NET Core 3.0 (and presumably Mono 5.18), but there are still some other frameworks around with it that we want to support.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -3632,8 +3632,6 @@ namespace IronPython.Runtime.Operations {
         }
 
         public static Exception UnicodeEncodeError(string encoding, char charUnknown, int index, string format, params object[] args) {
-            var ctor = typeof (EncoderFallbackException).GetConstructor(
-                BindingFlags.NonPublic | BindingFlags.Instance, null, new [] { typeof(string), typeof(char), typeof(int) } , null);
             var ex = (EncoderFallbackException)UnicodeEncodeError(string.Format(format, args), charUnknown, index);
             ex.Data["encoding"] = encoding;
             return ex;
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -3631,12 +3631,18 @@ namespace IronPython.Runtime.Operations {
             return PythonExceptions.CreateThrowable(PythonExceptions.UnicodeEncodeError, encoding, @object, start, end, reason);
         }
 
-        public static Exception UnicodeEncodeError(string encoding, char charUnkown, int index, string format, params object[] args) {
+        public static Exception UnicodeEncodeError(string encoding, char charUnknown, int index, string format, params object[] args) {
             var ctor = typeof (EncoderFallbackException).GetConstructor(
                 BindingFlags.NonPublic | BindingFlags.Instance, null, new [] { typeof(string), typeof(char), typeof(int) } , null);
-            var ex = (EncoderFallbackException)ctor.Invoke(new object[] { string.Format(format, args), charUnkown, index });
+            var ex = (EncoderFallbackException)UnicodeEncodeError(string.Format(format, args), charUnknown, index);
             ex.Data["encoding"] = encoding;
             return ex;
+        }
+
+        public static Exception UnicodeEncodeError(string message, char charUnknown, int index) {
+            var ctor = typeof(EncoderFallbackException).GetConstructor(
+                BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance, null, new[] { typeof(string), typeof(char), typeof(int) }, null);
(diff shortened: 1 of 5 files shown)


Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -137,6 +137,22 @@ public override Decoder GetDecoder()
         public override byte[] GetPreamble() => Pass1Encoding.GetPreamble();
         public override bool IsAlwaysNormalized(NormalizationForm form) => false;
 
+        public bool HasBugCorefx29898 {
```
slozier: Could this be a static?

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -367,23 +437,29 @@ public abstract class PythonDecoderFallback : DecoderFallback, ICloneable {
...
 
-            public PythonDecoderFallbackBuffer(bool isPass1, int encodingCharWidth)
-            {
+            public PythonDecoderFallbackBuffer(bool isPass1, PythonEncoding encoding) {
                 _marker = isPass1 ? Pass1Marker : Pass2Marker;
                 _fallbackChars = isPass1 ? null : new Queue<char>();
-                this.EncodingCharWidth = encodingCharWidth;
+                this.EncodingCharWidth = encoding.CharacterWidth;
+                this.CodePage = encoding.CodePage;
+                _hasIndexBug = encoding.HasBugCorefx29898;
```
slozier: If `HasBugCorefx29898` can be static then we don't need `_hasIndexBug`.

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -504,4 +591,288 @@ public SurrogateEscapeDecoderFallbackBuffer(bool isPass1, int encodingCharWidth)
...
+                }
+
+                byte[] fallbackBytes;
+                if (this.EncodingCharWidth > 1) {
+                    fallbackBytes = BitConverter.GetBytes(charUnknown);
+                    if (fallbackBytes.Length == this.EncodingCharWidth) {
+                        // UTF-16LE or UTF-16BE
+                        if (BitConverter.IsLittleEndian == _isBigEndianEncoding) {
+                            // swap bytes for non-native endianness encoding
+                            //(fallbackBytes[0], fallbackBytes[1]) = (fallbackBytes[1], fallbackBytes[0]);
```
slozier: I think the usage here doesn't end up adding a reference to the `System.ValueTuple` assembly so in theory we could reference the package when building but not include it in the release. Although I'm not seeing a way to prevent using it in a way that adds a reference so if we do that then we might end up using it by accident?

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -3631,14 +3631,20 @@ public static partial class PythonOps {
             return PythonExceptions.CreateThrowable(PythonExceptions.UnicodeEncodeError, encoding, @object, start, end, reason);
         }
 
-        public static Exception UnicodeEncodeError(string encoding, char charUnkown, int index, string format, params object[] args) {
+        public static Exception UnicodeEncodeError(string encoding, char charUnknown, int index, string format, params object[] args) {
             var ctor = typeof (EncoderFallbackException).GetConstructor(
```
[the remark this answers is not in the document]
BCSharp: No, we don't. This an artifact of my imperfect interactive git-rebase. I'm going to remove it.

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -235,18 +259,31 @@ private struct ByteCounter {
...
+                this.CodePage = encoding.CodePage;
             }
 
             protected int EncodingCharWidth { get; }
+            protected int CodePage { get; }
 
             public abstract byte[] GetFallbackBytes(char charUnknown, int index);
 
             public override bool Fallback(char charUnknown, int index) {
+                // The design limitation fow wide-char encodings is that
```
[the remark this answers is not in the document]
BCSharp: OK.

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: Yes, indeed.

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: Right.

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: Sorry, I meant to delete this comment. I don't think that dragging a whole package into IronPython just to be able to swap two bytes is justified. A `temp` variable does just fine.

Unless you have bigger plans for `ValueTuple` in other places.

slozier:
Thanks again for all this encoding work!

PR #590, opened 2019-03-31
Document differences with CPython around handling encodings

BCSharp:
Notes from my recent work with .NET encodings. 

I've only documented differences that are not results of bugs in IronPython or CPython, but rather result of how .NET operates, so it is unlikely they will ever be resolved. Perhaps it is even not desirable to resolve them — after all IronPython is a .NET language.


BCSharp:
Yes, this may be a good idea. I will add a note.

PR #591, opened 2019-04-05
Get test_csv passing

gfmcknight:
This passes all of the currently enabled tests, but it's still sort of a work in progress as I'm looking for some feedback on my approach in `StringFormatter`. Some discrepancies between our formatting of float values and CPython's formatting were causing some test failures (and are undoubtedly the cause of some other failing tests from CPython).

The main concern here is with the 'hack' for formatting doubles in `__repr__`. We call into .NET's `String.Format()` which seems to be giving slightly off-values (such as 0.866 turns into 0.8659999...). If the double values are converted to decimals, the formatting gets much closer to that of CPython. This is also an issue that CPython had, and allegedly added a new implementation, but seemingly now uses printf's formatting.

Where should I go from here?

---
Here are some relevant sources on this:
- Using the same precision on str and repr for floats: `https://bugs.python.org/issue9337`
- Initial discussion of the rounding: `https://bugs.python.org/issue1580`
- It appears that now, python relies on C's floating point formatting, as seen in the use of `_PyOS_ascii_formatd()` (which eventually calls vsprintf) on line 920 of this file: `https://github.com/python/cpython/blob/master/Python/pystrtod.c#L920`. This function I'm pointing out, `PyOS_double_to_string()`, is called by `__repr__` for floats.

As far as the modified IronPython test (strformat.py), here are the results I got in Python 3.6
```
>>> format(1000000000000.0, '')
'1000000000000.0'
>>> format(1000000000000.0, 'g')
'1e+12'
>>> format(-1000000000000.0, '')
'-1000000000000.0'
>>> format(-1000000000000.0, 'g')
'-1e+12'
```


slozier:
Thanks for the PR, overall it looks good. The double to decimal hack is however problematic. While it does generally give you a representation closer to that of CPython it most cases it fails the "round-tripping" test which is arguably more important (you may need to run this a few times but you should quickly see round-trip failures):
```Python
import random
import decimal

def test(n):
    for _ in range(n):
        x = random.random()
        if x != float(repr(x)):
            print(decimal.Decimal(x), "!=", x)

test(1000000)
```

`test_csv` seems to be passing even without the hack so maybe you could revert it for this PR?

gfmcknight:
When I do this, these tests fail because they expect short `__str__` results:
- net45.AllCPython.test_difflib
- net45.AllCPython.test_optparse

I think this needs to this in some way before bringing it in. What I didn't realize is that CPython still uses David Gay's code; it's just that the printf functions call into his dtoa instead of the standard one. Here's what I think the options are:
1. Try to modify the above tests and call this difference an implementation of IronPython.
2. Modify test_csv (using repr instead of str for the failing test) and don't unify str and repr until later.
3. Come up with an algorithm that satisfies the round tripping test, or try to use David Gay's dtoa.

I think that the first option is best because of this discussion:
```
On Tue, Apr 7, 2009 at 3:10 AM, Mark Dickinson <report@bugs.python.org> wrote:
> A proposal: I propose that the short float representation should be
> considered an implementation detail for CPython, not a requirement for
> Python the language.  This leaves Jython and IronPython and friends free
> to do as they wish.

In principle that's fine with me.
```
I'll update the PR with option 1 once I've tested it.

slozier:
Ah I see, I hadn't run the full test suite. In this case I agree that the first option is best.

I've looked at implementing the dtoa algorithm in C# however it's quite complicated and ended up scrapping it. A more reasonable approach would be to cannibalize the [FormatDouble](https://github.com/dotnet/coreclr/blob/10998b4750af123d81b18c5b68b687027fec2bf9/src/System.Private.CoreLib/shared/System/Number.Formatting.cs#L523) from .NET Core which has a proper shortest roundtrip representation.

slozier:
Thanks for the PR! I tweaked your `test_optparse` bypass so that it will fail (and we remember to revert it) when the float formatting gets updated.

PR #596, opened 2019-04-07
Interop with System.Text.Encoding, DecoderFallbackException and EncoderFallbackException

### LINKED ISSUE #551: encode/decode broken with System.Text.Encoding objects
Looks like I probably broke it in https://github.com/IronLanguages/ironpython3/pull/535

Basically we should support:
```Python
"aaa".encode(System.Text.Encoding.ASCII).decode(System.Text.Encoding.ASCII)
```

There might be some disabled tests that cover this, if not then we should definitely add some.

BCSharp:
To get encoding names on .NET Core, the code detects the most common encodings and gives (most of) them [canonical .NET names](https://docs.microsoft.com/en-us/dotnet/api/system.text.encodinginfo.name?view=netframework-4.7.2). These are not always the same as [CPython's canonical names](https://docs.python.org/3/library/codecs.html?highlight=codecs#standard-encodings), however, all of them are recognized by CPython as at least aliases. I think it is more appropriate to use .NET names here as the code tries to give a name to a .NET object, possibly user-created. Also, it will align the names when run on .NET Core with names from .NET Frm/Mono.

Here are the differences:

| IronPython (and CPython alias) | CPython | .NET Framework |
| --- | --- | --- |
| us-ascii | ascii | us-ascii |
| iso-8859-1 | latin-1 | iso-8859-1 |
| utf-16LE | utf-16-le | utf-16 (note 1) |
| utf-16BE | utf-16-be | unicodeFFFE (note 2) |
| utf-32LE | utf-32-le | utf-32 (note 1) |
| utf-32BE | utf-32-be | utf-32BE |

There are no differences for 'utf-7' and 'utf-8'.

_Note 1_: CPython also has encoding 'utf-16' but it designates a native encoding that generates (or swallows on decoding) the preamble with BOM. .NET Unicode encoding can do the same, so at first I had code detect the preamble in the given encoding and report 'utf-16' or 'utf-16LE' accordingly. However, it turns out that CPython will **always** report the specific codec name in use when throwing an instance of a subclass `UnicodeError` . So 'utf-16' will never be seen in `UnicodeError`, always the specific codec 'utf-16-le' or 'utf-16-be'. It is the same situation for 'utf-32': it is always converted to 'utf-32-le' or 'utf-32-be' and for 'utf-8-sig' converted to 'utf-8'.

_Note 2_: The .NET name is a clear misnomer and better not be used (in fact, the first two bytes generated by this encoding will be FE FF and not what the name suggests). Also `WebName` on this encoding (where defined) will return 'utf-16BE'.

Another noteworthy change is handling of `UnicodeEncodeError` and `UnicodeDecodeError`. These are not so interoperable with `EncoderFallbackException` (EFE) and `DecoderFallbackException` (DFE) as the code originally assumed. .NET exceptions carry the position of the error and the erroneous data itself. Python specification is that the exception carries the full original input, and the `start:end` slice indicates the erroneous data. That is, the following invariant must be satisfied: `ex.object[ex.start:ex.end] == bad_data` One way of getting minimum compliance here is to always set `start` to 0 and `end` to the length of bad data carried in the .NET exception, however, information about the location (index) of the error would be lost. This is actually allowed in Python in the incremental encoding/decoding case, but very unfortunate, as the error position is invaluable in investigating the error. The code in this PR then extends EFE and DFE with extra data that will allow for full Python compliance. To prevent accidental creation of EFE and DFE without data, some overloads if `PythonOps.UnicodeEncodeError` and `PythonOps.UnicodeDecodeError` are removed. The data-less EFE and DFE instances can still be created in the user code so the interop code will handle them correctly, although they may not be as informative.

Fixes #551.

### DIFF
--- a/Src/IronPython/Runtime/ByteArray.cs
+++ b/Src/IronPython/Runtime/ByteArray.cs
@@ -14,6 +14,7 @@ using Microsoft.Scripting.Utils;
 
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
+using System.Text;
 
 namespace IronPython.Runtime {
     /// <summary>
@@ -263,6 +264,12 @@ namespace IronPython.Runtime {
         public string decode(CodeContext/*!*/ context, [NotNull]string encoding = "utf-8", [NotNull]string errors = "strict") {
             lock (this) {
                 return StringOps.RawDecode(context, _bytes, encoding, errors);
+            }
+        }
+
+        public string decode(CodeContext/*!*/ context, [NotNull]Encoding encoding, [NotNull]string errors = "strict") {
+            lock (this) {
+                return StringOps.DoDecode(context, _bytes, errors, StringOps.GetEncodingName(encoding, normalize: false), encoding);
             }
         }
 
--- a/Src/IronPython/Runtime/ByteArray.cs
+++ b/Src/IronPython/Runtime/ByteArray.cs
@@ -8,13 +8,13 @@ using System.Collections.Generic;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Text;
 
 using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
-using System.Text;
 
 namespace IronPython.Runtime {
     /// <summary>
--- a/Src/IronPython/Runtime/Bytes.cs
+++ b/Src/IronPython/Runtime/Bytes.cs
@@ -10,16 +10,16 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Numerics;
 using System.Runtime.InteropServices;
+using System.Text;
 
 using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
-using System.Text;
 
 namespace IronPython.Runtime {
-    [PythonType("bytes")]
+    [PythonType("bytes"), Serializable]
     public class Bytes : IList<byte>, IEquatable<Bytes>, ICodeFormattable, IExpressionSerializable, IBufferProtocol {
         internal byte[]/*!*/ _bytes;
         internal static Bytes/*!*/ Empty = new Bytes();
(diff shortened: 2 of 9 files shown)


Src/IronPython/Runtime/ByteArray.cs
```diff
@@ -14,6 +14,7 @@
 
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
+using System.Text;
```
slozier: nit: Move to up to the "using System" group?

Src/IronPython/Runtime/Bytes.cs
```diff
@@ -16,6 +16,7 @@
 
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
+using System.Text;
```
slozier: nit: Move to up to the "using System" group?

Src/IronPython/Runtime/Bytes.cs
```diff
@@ -16,6 +16,7 @@
...
 
 namespace IronPython.Runtime {
     [PythonType("bytes")]
```
slozier: Looks like you'll need to add `Serializable` to `Bytes` for .NET Framework since `Exception.Data` elements need to be serializable.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -133,8 +133,8 @@ public class ExtensibleString : Extensible<string>, ICodeFormattable, IStructura
         }
 
         public virtual bool __contains__(object value) {
-            if (value is string) return Value.Contains((string)value);
-            else if (value is ExtensibleString) return Value.Contains(((ExtensibleString)value).Value);
+            if (value is ExtensibleString es) return Value.Contains(es.Value);
+            if (value is string s) return Value.Contains((string)value);
```
slozier: nit: `Value.Contains(s)`

Src/IronPython/Runtime/ByteArray.cs
BCSharp: OK

Src/IronPython/Runtime/Bytes.cs
BCSharp: OK

Src/IronPython/Runtime/Bytes.cs
BCSharp: it just shows the necessity of always testing on Windows

Src/IronPython/Runtime/Operations/StringOps.cs
BCSharp: good catch - that was the whole point of the change to avoid double casting

slozier:
Thanks for the PR!

PR #597, opened 2019-04-09
Allow optional characters in UTF-7

BCSharp:
This makes 'utf-7' codec behaviour much closer to CPython's, although still not quite 100%. The documentation and tests are updated accordingly.

As a part of this change I had to extend `EncodingInfoWrapper` to support yet another usage variant, on top of the existing two. As the code was becoming more obscure, I took the liberty to simplify `EncodingInfoWrapper` and eliminate `EncodingWrapper` altogether. The remaining functionality in `EncodingInfoWrapper` had little to do with wrapping `EncodingInfo` any more and a new name `EncodingSingletonFactory` seemed more fitting.

### DIFF
--- a/Documentation/differences-from-c-python.md
+++ b/Documentation/differences-from-c-python.md
@@ -214,9 +214,9 @@ _IronPython_
 b'\x81'
 ```
 
-* When using the UTF-7 encoding, IronPython (and .NET) always terminates the modified Base64 encoded blocks with a '-' while CPython ommits the '-' if allowed.
+* When using the UTF-7 encoding, IronPython (and .NET) always terminates the modified Base64 encoded blocks with a '-' while CPython omits the '-' if allowed.
 
-The UTF-7 standars allows encoders for some freedom of implementation. One optionality allowed in UTF-7 is how to end a sequence encoded in the modified Base64 code. In principle, `+` marks the start of the sequence, and `-` is the terminator. However, it is allowed to ommit the terminating `-` if the next character unabiguously does not belong to the encoded Base64 block. CPython chooses to drop the terminating `-` in such cases, while IronPython will always terminate Base64-encoded blocks with a `-`:
+The UTF-7 standard allows encoders for some freedom of implementation. One optionality allowed in UTF-7 is how to end a sequence encoded in the modified Base64 code. In principle, `+` marks the start of the sequence, and `-` is the terminator. However, it is allowed to omit the terminating `-` if the next character unambiguously does not belong to the encoded Base64 block. CPython chooses to drop the terminating `-` in such cases, while IronPython will always terminate Base64-encoded blocks with a `-`:
 
 _CPython_
 ```
--- a/Documentation/differences-from-c-python.md
+++ b/Documentation/differences-from-c-python.md
@@ -214,25 +214,9 @@ _IronPython_
 b'\x81'
 ```
 
-* When using the UTF-7 encoding, IronPython (and .NET) uses the long encoding form and encodes optional characters, while CPython uses the short form and passes the optional characters unmodified.
+* When using the UTF-7 encoding, IronPython (and .NET) always terminates the modified Base64 encoded blocks with a '-' while CPython ommits the '-' if allowed.
 
-The UTF-7 standars allows encoders for some freedom of implementation. There are some characters, called _direct characters_, that are allways passed through literally on encoding (letters, digits, and 9 symbols: ` ' ( ) , - . / : ?. `). The other main group, known as "optional direct characters", contains all other printable characters in the range U+0021÷U+007E except ` ~ \ +`. Using the optional direct characters reduces size and enhances human readability but also increases the chance of breakage by things like badly designed mail gateways. 
-
-Here is where CPython and IronPython make different choices. CPython chooses on encoding to pass through the optional characters, while IronPython will encode them in the `+ -` delimited modified Base64 encoded sequence:
-
-_CPython_
-```
->>> '!"#$%&*;<=>@[]^_`{|}'.encode('utf-7')
-b'!"#$%&*;<=>@[]^_`{|}'
-```
-
-_IronPython_
-```
->>> '!"#$%&*;<=>@[]^_`{|}'.encode('utf-7')
-b'+ACEAIgAjACQAJQAmACoAOwA8AD0APgBAAFsAXQBeAF8AYAB7AHwAfQ-'
-```
-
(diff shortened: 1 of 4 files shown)


Documentation/differences-from-c-python.md
```diff
@@ -214,25 +214,9 @@ _IronPython_
...
-```
-
-_IronPython_
-```
->>> '!"#$%&*;<=>@[]^_`{|}'.encode('utf-7')
-b'+ACEAIgAjACQAJQAmACoAOwA8AD0APgBAAFsAXQBeAF8AYAB7AHwAfQ-'
-```
-
-Another optionality allowed in UTF-7 is how to end a sequence encoded in the modified Base64 code. In principle, `+` marks the start of the sequence, and `-` is the terminator. However, it is allowed to ommit the terminating `-` if the next character unabiguously does not belong to the encoded Base64 block. CPython chooses to drop the terminating `-` in such cases, while IronPython will always terminate Base64-encoded blocks with a `-`:
+The UTF-7 standars allows encoders for some freedom of implementation. One optionality allowed in UTF-7 is how to end a sequence encoded in the modified Base64 code. In principle, `+` marks the start of the sequence, and `-` is the terminator. However, it is allowed to ommit the terminating `-` if the next character unabiguously does not belong to the encoded Base64 block. CPython chooses to drop the terminating `-` in such cases, while IronPython will always terminate Base64-encoded blocks with a `-`:
```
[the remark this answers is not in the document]
BCSharp: OK

Documentation/differences-from-c-python.md
```diff
@@ -214,25 +214,9 @@ _IronPython_
 b'\x81'
 ```
 
-* When using the UTF-7 encoding, IronPython (and .NET) uses the long encoding form and encodes optional characters, while CPython uses the short form and passes the optional characters unmodified.
+* When using the UTF-7 encoding, IronPython (and .NET) always terminates the modified Base64 encoded blocks with a '-' while CPython ommits the '-' if allowed.
```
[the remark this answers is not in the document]
BCSharp: OK

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1448,18 +1448,24 @@ public static class StringOps {
 #if FEATURE_ENCODING
             name = NormalizeEncodingName(name);
 
-            EncodingInfoWrapper encInfo;
-            if (CodecsInfo.Codecs.TryGetValue(name, out encInfo)) {
-                encoding = (Encoding)encInfo.GetEncoding().Clone();
+            EncodingSingletonFactory efac;
```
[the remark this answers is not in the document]
BCSharp: I thought about it too but I was not sure. `EncodingSingletonFactory` is now equivalent to `Lazy<Encoding>` with `isThreadSafe: false`. I wasn't sure if it wouldn't allocate a more heavy-weight instance in memory, and the footprint of the change would be bigger: all `GetEncoding` calls would change to `Value`. 

But I'm glad you pointed it out, I prefer to use standard classes where possible.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1701,7 +1706,7 @@ public static class StringOps {
 #if FEATURE_ENCODING
             // CLR's encoder exceptions have a 1-1 mapping w/ Python's encoder exceptions
             // so we just clone the encoding & set the fallback to throw in strict mode.
-            e = (Encoding)e.Clone();
+            if (e.IsReadOnly) e = (Encoding)e.Clone();
```
slozier: I imagine you put this check for perf, however couldn't this cause all sorts of issues? For example, potentially mutating the .NET encoding that was given in `decode`? Race conditions with threading? Same comment for `DoEncode`.

Src/IronPython/Compiler/Tokenizer.cs
```diff
@@ -1575,6 +1575,7 @@ private class TokenEqualityComparer : IEqualityComparer<object> {
             // and we have the magic ending as well...
             if (StringOps.TryGetEncoding(encName, out enc)) {
 #if FEATURE_ENCODING
+                if (enc.IsReadOnly) enc = (Encoding)enc.Clone();
```
slozier: See StringOps comment.

Src/IronPython/Runtime/Operations/StringOps.cs
BCSharp: Yes, it was meant to reduce the amount of cloning happening, but it was a bad idea - as you pointed out, the encoding may have been already cloned by the user and would have been mutated here. As for race conditions, is `Clone` itself thread-safe?

slozier: I have no idea is `Clone` itself is thread safe, it just does a memberwise copy. When I mentioned race conditions what I had in mind was more multiple `DoDecode` calls at the same time with an encoding that was already mutated:
1. (thread 1) e.DecoderFallback = bla
2. (thread 2) e.DecoderFallback = bla2
3. (thread 1) e.GetString() // this one ends up with the bla2 fallback
4. (thread 2) e.GetString()

BCSharp: Yes, cloning will prevent such scenario. I think that the best rule here is to always (unconditionally) clone before setting a fallback and do it immediately before (to avoid unnecessary cloning if done earlier). I will change the relevant code in `Tokenizer` too to follow this rule, so that there is no assumption there on whence the encoding comes when returned by `TryGetEncoding`. 

But if `DoDecode` is meant to be thread-safe (and why not), I was thinking about such scenario:

1. (thread 1) e.Clone() // thread preempted half-way
2. (thread 2) e.SomeProperty = newValue
3. (thread 1) complete e.Clone() // object in an inconsistent state
4. (thread 1) e.GetString() // fails because of invalid state of e

If `MemberwiseCopy` copy is atomic then there is no problem, unless a derived encoding overrides `Clone`. And even if `MemberwiseCopy` is not atomic, it is no problem for the stock .NET encodings, as the only writable properties are the fallbacks, and they will be reset afterwards anyway. A derived encoding with a more complex cloning strategy will have to implement its own locks. Currently, there no such cases in IronPython but the codecs module is not finished yet. I am making a mental note to watch out for it.

Src/IronPython/Compiler/Tokenizer.cs
BCSharp: See StringOps comment.

Src/IronPython/Runtime/Operations/StringOps.cs
slozier: I agree unconditional cloning is the way to go.

I would argue that even if `Clone` is not atomic, then `DoDecode` would still be thread safe (by any sane definition of thread safety). If a user calls a method with an object and then spins up a thread to mutate the object, then it seems like they're asking for trouble. Anyway, other than being an interesting discussion I guess it's doesn't affect this PR. :smile:

BCSharp: No, it was just to figure out what needs to be done in the future, if anything.

For this PR, I have the changes ready, just waiting for the tests to finish.

Issue #580, opened 2019-03-22
open with a bad file descriptor raise SystemError


slozier:
```Python
open(12345)
```
should raise an `OSError`:
> OSError: [Errno 9] Bad file descriptor

PR #600, opened 2019-04-15
#580 fix open(bogus file descriptor) to raise OSError

simonwyatt:
Resolves https://github.com/IronLanguages/ironpython3/issues/580

### DIFF
--- a/Src/IronPython/Modules/_fileio.cs
+++ b/Src/IronPython/Modules/_fileio.cs
@@ -291,7 +291,7 @@ namespace IronPython.Modules {
             public override int fileno(CodeContext/*!*/ context) {
                 _checkClosed();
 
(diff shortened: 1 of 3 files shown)


slozier:
I'm fine with this PR as is, however if we want to strengthen type safety we should also restrict `GetOrAssignIdForObject` (or just remove it?) . Also, I don't remember where I encountered issue https://github.com/IronLanguages/ironpython3/issues/580, so we could probably use a simple test (in `test_regressions.py`). Something like:
```Python
    def test_ipy3_gh580(self):
        """https://github.com/IronLanguages/ironpython3/issues/580"""
        with self.assertRaises(OSError):
            open(12345)
```

simonwyatt:
Thanks for the feedback. Your suggestions incorporated.

slozier:
Thanks for the PR!

Issue #22, opened 2014-03-21
PEP 3120 -- Using UTF-8 as the default source encoding


jdhardy:
http://www.python.org/dev/peps/pep-3120/

slozier:
CPython test for this is: Src\StdLib\Lib\test\test_pep3120.py

PR #603, opened 2019-04-17
Implement PEP-263 and PEP-3120 (WIP)

BCSharp:
Sorry for a large PR; things turned out to be quite interdependent. 
I decided to open this PR as WIP, to make the changes more manageable. The rest will come after this PR is reviewed, accepted, and merged.

The most prominent feature not working is that `PythonDecodeError`/`DecoderFallbackException` is not always converted to `SyntaxError` when source files are read.

Tests for this PR are in `test_source_encoding` in StdLib. A few of them are still failing so I did not enable them yet.

The few IronPython tests for this functionality were not quite correct so needed to be changed.


Src/IronPython.Modules/_codecs.cs
```diff
@@ -122,7 +122,7 @@ public static PythonTuple ascii_encode(CodeContext context, string input, string
 
         public static object decode(CodeContext/*!*/ context, object obj, string encoding = null, string errors = "strict") {
             if (encoding == null) {
-                encoding = context.LanguageContext.DefaultEncoding.EncodingName;
+                encoding = context.LanguageContext.GetDefaultEncodingName();
```
slozier: I haven't looked into this, but any idea why we wouldn't just use the default encoding directly with `DoDecode` instead of going via `lookup`?

Src/IronPython/Compiler/Tokenizer.cs
```diff
@@ -1529,15 +1529,16 @@ private class TokenEqualityComparer : IEqualityComparer<object> {
             }
         }
 
-        internal static bool TryGetEncoding(Encoding defaultEncoding, string line, ref Encoding enc, out string encName) {
-            // PEP 0263 defines the following regex for the encoding line
-            // coding[:=]\s*([-\w.]+)
+        internal static bool TryGetEncoding(string line, ref Encoding enc, out string encName) {
```
slozier: This method signature is slightly confusing. What about changing it to:
```C#
internal static bool TryGetEncodingName(string line, out string encName);
```
and moving the actual encoding lookup into `GetSourceReader`?
```C#
            // lookup the encoding
            Encoding encoding = defaultEncoding;
            if (gotEncodingName && StringOps.TryGetEncoding(encodingName, out encoding)) {
#if FEATURE_ENCODING
                encoding = (Encoding)encoding.Clone();
                encoding.DecoderFallback = DecoderFallback.ExceptionFallback;
#endif
            }
```

Src/IronPython/Modules/Builtin.cs
```diff
@@ -217,9 +217,15 @@ public static partial class Builtin {
         }
 
         [Documentation("")] // provided by first overload
-        public static object compile(CodeContext/*!*/ context, [BytesConversion]IList<byte> source, string filename, string mode, object flags = null, object dont_inherit = null, int optimize = -1)
-            // TODO: Detect encoding as prescribed in PEP 263
-            => compile(context, StringOps.RawDecode(context, source, "utf-8", "strict"), filename, mode, flags, dont_inherit, optimize);
+        public static object compile(CodeContext/*!*/ context, [BytesConversion]IList<byte> source, string filename, string mode, object flags = null, object dont_inherit = null, int optimize = -1) {
+            byte[] bytes = source as byte[] ?? ((source is Bytes b) ? b.GetUnsafeByteArray() : source.ToArray());
```
slozier: I have some similar changes in https://github.com/IronLanguages/ironpython3/pull/599. Could you take a look and see if the two could be unified? I'm not sure if it's better one way or another...

Src/IronPython.Modules/_codecs.cs
BCSharp: I suppose you are right. If I had to guess, I would say it was because with `setdefaultencoding` it was possible to set any encoding as default, including a user-defined one. 

I haven't looked into this either, except noticing that there are a number of things in `_codecs` that don't seem quite right. I was thinking that once I start pulling on this thread, more things start to unravel, which are better left for a separate PR. So my intention for this WIP was to have minimal changes to `_codecs` that will just let this PR to pass the tests.

I will implement your suggestion now if the tests still pass.

Src/IronPython/Compiler/Tokenizer.cs
BCSharp: I like your suggestion. I too find this method semantics confusing, mostly because of the double meaning of `false` returned from it: it means either that there is no "magic" comment, or there is one but the encoding name is invalid. The way to differentiate them is different handling of the `ref` and `out` arguments.

Src/IronPython/Modules/Builtin.cs
BCSharp: I see. Will look into it. I saw the title of your PR passing by but did not expect it to cross with my focus here, which is #22 ("good first issue" :smirk:). Since PEP-3120 builds on top of PEP-263, the latter one got dragged in.

slozier: I was also surprised that it lead me there. Apparently importlib uses compile to load its Python modules.

Tests/modules/io_related/test_codecs.py
```diff
@@ -390,12 +404,27 @@ def test_cp11334(self):
...
+            self.assertEqual(t_out_lines[0].rstrip(), b"\xe6ble")
         else:
-            self.assertTrue(t_err_lines[1].startswith(b"SyntaxError: Non-UTF-8 code starting with '\\xb5' in file"))
+            self.assertEqual(t_out_lines[0].rstrip(), b"\xb5ble")
+        self.assertEqual(len(t_out_lines), 1)
 
-        #--Test that using "# coding ..." is OK
+        #--Test that using other whitespace characters (rather than [ \t\f]) is OK
+        #but non-ASCII letters in the encoding name are not accepted
         p = subprocess.Popen([sys.executable, os.path.join(self.test_dir, "encoded_files", "cp11334_ok.py")], shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
slozier: Shouldn't this test be using `cp11334_ok2.py`?

Src/IronPython/Modules/Builtin.cs
BCSharp: It is not easy to choose simply one way or another. I see some some good elements in #599 and I prefer some other solutions from #603. To summarize:

 1. We both agree that a `MemoryStream` backed content provider is needed in the base `IronPython` project. The implementations we chose are virtually the same. Since there was already a class like this in zipimport, I think it is better to reuse it my moving to the main project rather than duplicating the code. Also the name `MemoryStreamContentprovider` seems more accurate than `PythonStreamContentProvider`. This is what #603 does.

 2. I like the idea of factoring out parts of `compile` to `ValidateMode` (I would prefer calling it `ValidateCompileMode`) and `CompileHelper`.

 3. The conversion to `byte[]` in the `IList<Byte>` overload of `compile` in #599 is somewhat restrictive. The conversion in #603 is wider and follows the pattern from `StringOps.DoDecode`.

 4. #603 addresses all three: `compile`, `eval`, and `exec`, while #599 only `compile`. In #603, only `exec` goes through `SourceUnit`, using directly `MemoryStream` and `SourceCodeReader` in the other two cases (I did it so for performance reasons). #599 uses `SourceUnit` for `compile`, which makes the code more regular. I propose to adopt the #599 approach to `compile`, leaving `eval` as it is (unless some neat refactoring opportunity emerges).

 5. The the last change in `PythonContext` in #599 (reporting unknown encoding name, around line 994) is superseded by more in depth changes in #603. I think it is best to revert this particular change in #599 to prevent merge conflicts.

 6. #599 has some whitespace-only corrections in `PythonContext` that will generate merge conflicts with #603. Best to revert (some of) them.

Let me know what you think of it.

Tests/modules/io_related/test_codecs.py
BCSharp: right

Src/IronPython/Modules/Builtin.cs
slozier: Agreed on everything. Good analysis.

1. The reason I renamed it was because I wanted to use it with a `Stream` and so having `MemoryStream` in the name no longer seemed accurate. However, I think leaving it as `MemoryStreamContentProvider` in here is fine since I have not finalized #599 and may still change things.

BCSharp:
Thanks for merging.

BCSharp:
Unfortunately, my last change to `compile` broke some things that were already working. The test run did not detect it because the tests that cover it are in `test_source-encoding` from StdLib. I will make a separate PR for that. I wish it were possible to enable/ignore tests from StdLib on the individual test level, rather than the whole module.

slozier:
If there are only a few failures I'm usually fine with disabling the failing tests (as long as there's an issue about the failures).

BCSharp:
OK, I will keep it in mind, although I've realized that it would not have helped in this case: some tests are failing in both cases, as the only indication of a regression is the way how they are failing.

I am thinking of simply making a modified copy of the test from StdLib and putting it with the IronPython tests. Modified in a way that it tests what is already working but has proper workaround for what is not yet done. Once there are no more workarounds required, the test can be deleted and the test from StdLib takes over.

PR #604, opened 2019-04-18
Allow IFormattable objects to override __format__

### LINKED ISSUE #465: Method resolution on .NET Types does not work as expected
```Python
import System
dt = System.DateTime(2008, 10, 26)
assert format(dt, 'MM-dd') == '10-26'
```
This calls the `__format__` of `object` instead of the one for `IFormattable`.

gfmcknight:
Do not resolve methods on Object other forms of member resolution have
been exhausted.

Resolves #465.

Root cause:
> values that live on object need to run before the StandardResolver
> which will pick up values off of object.

There are two things I'm trying to address in this pull request:
- We want to allow types to override object's `__format__` by being IFormattable, but we also want a type to be able to define `__format__` in order to override its IFormattable behavior.
- This issue is likely to happen again whenever we add anything else to ObjectOps.

One potential issue here is that this changes the order of member resolution for `__str__` `__new__` `__repr__` `__hash__` and `__iter__` slightly. While it doesn't seem to have caused any failures on currently enabled tests, if we have dead code for methods with these names, the methods are now called with this change.

### DIFF
--- a/Src/IronPython/Runtime/Types/PythonTypeInfo.cs
+++ b/Src/IronPython/Runtime/Types/PythonTypeInfo.cs
@@ -206,7 +206,7 @@ namespace IronPython.Runtime.Types {
                 MemberGroup res;
 
                 foreach (Type curType in binder.GetContributingTypes(type)) {
-                    res = FilterSpecialNames(binder.GetMember(curType, name), name, action);
+                    res = FilterObjectMembers(FilterSpecialNames(binder.GetMember(curType, name), name, action));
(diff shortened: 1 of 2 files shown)


gfmcknight:
My bad -- I got excited when the only failure was intentional and didn't even read the next line.

slozier:
Actually, it's my bad since I added that return statement when enabling the test! 😃 

Overall this looks good to me, just running some tests to see if I can break things. Will likely merge this later today.

PR #606, opened 2019-04-22
Implement PEP-263 and PEP-3120 (WIP 2)

BCSharp:
This PR contains mostly error message improvements (to align them more closely with CPython's) and `eval` changes for the bytes-like overload. Perfect alignment will probably not be practical as CPython may generate different error messages depending whether the code is imported, compiled inline, or run from a file given on the command line.

With these changes, `test.test_source_encoding` from StdLib passes (excluding two minor implementation-dependent incompatibilities with CPython).


slozier:
Thanks for the PR!

Issue #245, opened 2017-02-24
Exception context issue


slozier:
The meaning of "handling an exception" in `sys.exc_info` changed in Python 3. It changed from (see https://docs.python.org/3/library/sys.html#sys.exc_info):
> executing or having executed an except clause

to:

> executing an except clause

Because of this change the tracebacks may return incorrect outputs, for example:
```Python
def test4():
    try:
        raise Exception(1)
    except:
        pass    
    raise Exception(2)

test4()
```
outputs (we should only see the Exception: 2 traceback):
```
Traceback (most recent call last):
  File "<stdin>", line 3, in test4
Exception: 1

During handing of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 6, in test4
Exception: 2
```

Here are a few test cases which should all pass:
```Python
import sys

def test1():
    try:
        raise Exception
    except:
        pass
    assert sys.exc_info() == (None, None, None)

def test2():
    def gen():
        try:
            raise Exception
        except:
            exc_info = sys.exc_info()
            yield exc_info
            yield sys.exc_info()
        yield sys.exc_info()
    x = gen()
    exc_info = next(x)
    assert sys.exc_info() == (None, None, None)
    assert next(x) == exc_info
    assert next(x) == (None, None, None)

def test3():
    try:
        try:
            raise Exception(1)
        except:
            pass    
        raise Exception(2)
    except Exception as e:
        assert e.__context__ is None

test1()
test2()
test3()
```

PR #607, opened 2019-04-24
Get test_math passing

gfmcknight:
Apply division before mutliplication to avoid getting infinity for
large values passed to math.log1p().

Root cause: 
The term log(v1) * v0 is too large to represent in a double for very large v0.

This affects the results of many inputs to log1p:

```
import math
import random 
log = math.log
for _ in range (100):
    v0 = random.random() * 1024
    v1 = v0 + 1.0
    logA = log(v1) * v0 / (v1 - 1.0)
    logB = log(v1) * (v0 / (v1 - 1.0))
    if (logA != logB):
        print('Differing results on input', v0, '---', logA, '!=', logB)
```

Yields several results, but it seems to be within the acceptable margin for test_math. CPython appears to use `log1p()` from the C standard library, but I'm not completely certain of this claim. I tried some of the results from the above in CPython, and got results that matched neither, so from my perspective, this comes down to a matter of implementation.

Explanation of math domain error:
- The error is validly thrown previously
- This error shows up due to #245

### DIFF
--- a/Src/IronPython.Modules/math.cs
+++ b/Src/IronPython.Modules/math.cs
@@ -307,7 +307,7 @@ namespace IronPython.Modules {
             }
 
             // Apply correction factor
-            return log(v1) * v0 / (v1 - 1.0);
+            return log(v1) * (v0 / (v1 - 1.0));
         }
 
         public static double log1p(BigInteger value) {
--- a/Src/IronPython.Modules/pyexpat.cs
+++ b/Src/IronPython.Modules/pyexpat.cs
@@ -619,7 +619,12 @@ namespace IronPython.Modules {
                 if (!PythonOps.TryGetBoundAttr(context, file, "read", out object _readMethod))
                     throw PythonOps.TypeError("argument must have 'read' attribute");
 
-                if (!(PythonOps.CallWithContext(context, _readMethod) is string data)) throw PythonOps.TypeError("read() did not return a string object");
+                object readResult = PythonOps.CallWithContext(context, _readMethod);
+                if (readResult is Bytes byteData)
+                    readResult = byteData.MakeString();
+
+                if (!(readResult is string data))
+                    throw PythonOps.TypeError("read() did not return a string object");
 
                 using (var reader = new StringReader(data)) {
                     var settings = new XmlReaderSettings() { DtdProcessing = DtdProcessing.Parse, XmlResolver = null };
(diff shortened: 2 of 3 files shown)


slozier:
Looks like there's a failure on mac due to `plistlib` (although the underlying issue appears to be in `pyexpat`). Was this something you were interested in looking into? Otherwise I'm fine with filing an issue and disabling `test_math` on mac (or commenting out `@support.requires_mac_ver(10, 5)`).

gfmcknight:
I'll do some investigation into what's happening with the test. It'll be a little limited because I don't have a Mac that I'm free to install things on, so if I can't figure it out I'll come back here and say so.

slozier:
You should be able to replicate the (first?) failure on Windows with:
```Python
plistlib.loads(plistlib.dumps(""))
```

gfmcknight:
Just checking if this alone makes test_math work on Mac. I have yet to investigate what actually changed in Python3 for xmlparser (other than the fact that parse() can now take bytes). Also, I should probably be doing a decode here instead of MakeString().

slozier:
I believe the change that lead to this breaking is that `open(filename, "rb").read()` now returns `bytes` instead of the `str/unicode` of ipy2. I believe the proper behavior for `ParseFile` is to only accept a read that returns `bytes`.
> TypeError: read() did not return a bytes object (type=str)

 Although changing this might expose up a bunch of other issues... Maybe something like this (untested)?
```C#
                object readResult = PythonOps.CallWithContext(context, _readMethod);
                if (readResult is Bytes b) {
                    using (var stream = new MemoryStream(b.GetUnsafeByteArray())) {
                        var settings = new XmlReaderSettings() { DtdProcessing = DtdProcessing.Parse, XmlResolver = null };
                        using (xmlReader = XmlReader.Create(stream, settings)) {
                            parse(context);
                            _parsing_done = true;
                        }
                    }
                }
                else if (readResult is string s) {
                    // TODO: don't accept string!
                    using (var reader = new StringReader(s)) {
                        var settings = new XmlReaderSettings() { DtdProcessing = DtdProcessing.Parse, XmlResolver = null };
                        using (xmlReader = XmlReader.Create(reader, settings)) {
                            parse(context);
                            _parsing_done = true;
                        }
                    }
                }
                else {
                    throw PythonOps.TypeError("read() did not return a bytes object (type={0})", DynamicHelpers.GetPythonType(readResult).Name);
                }
```

slozier:
CI is having trouble picking up pull requests, but I ran your PR though the CI and it did seem to solve the test_math issue.

gfmcknight:
Thanks! I'll test out the above, and I'll look at how many things break when we don't accept strings.

gfmcknight:
Not accepting strings doesn't seem to have broken anything in the active tests, and it doesn't seem to have changed the result of test_pyexapat. This is likely because there are very few places in the code or tests that use ParseFile().

PR #610, opened 2019-04-28
Implement PEP-263 and PEP-3120 (final)

BCSharp:
This PR mainly contains corrected handling of source files starting with a Unicode BOM (UTF-8 and other) and additional tests for PEP-263 compliance. I label it "final" because it is the last from the series of #603 and #606, completing the work I planned to do on PEP-3120 (and Issue #22). There may still be bugs that I am not aware of. I am aware of the following though:

1. Some exception error messages are not the same as CPython's. This is because CPython generates sometimes different messages depending whether the file has been imported or run from the command line. Perhaps this can be resolved by some clever remapping higher up the stack than the parser, but IMHO it is not worth the effort. Where I had to make the choice, I chose the more informative error message. The tests in `IronPython.modules.io_related.test_codecs` document the differences. I was considering putting a condition `if is_ironpython:` around them, but I never got those tests running under CPython so I don't know whether this convention is still being used.

2. The location of the decoding error reported in the exception is approximate. This is partly because of a [CoreFX bug](https://github.com/dotnet/corefx/issues/29898), but more significantly because error handling in incremental decoding is not fully implemented yet. On top of that, CPython does not provide any meaningful (or correct) location in its exceptions, so it is clearly not a priority for the CPython developers. So I am inclined to call it "good enough", or at least put it on a low priority. After all, from the user's point of view, it is easy to get an accurate location of the decoding error by simply reading in the contents as bytes and decoding it to string.

3. The tests for this work are in `AllCPython.test_source_encoding` ([`Src/StdLib/Lib/test/test_source_encoding.py`](https://github.com/IronLanguages/ironpython3/blob/master/Src/StdLib/Lib/test/test_source_encoding.py)). I did not enable it because `test_isue2301` is failing due to a small incompatibility with CPython and I don't know yet whether it is a CPython bug, IronPython bug, or simply an implementation specific difference.


slozier:
1. I'm generally fine with not throwing exactly the same error message as long as the type of the error is correct.
2. Sounds good.
3. Were you going to look into enabling the test (`AllCPython.test_source_encoding`)? Would be good to have something working to avoid regressions. Also it looks like `AllCPython.test_pep3120` is passing so we should probably enable that as well.

Src/IronPython/Runtime/PythonContext.cs
```diff
@@ -936,125 +936,79 @@ private sealed class OptimizedBuiltinHasher {
...
-            // also StreamReader will automatically switch to proper Unicode encoding (non-throwing) when BOM present
-            using (StreamReader sr = new StreamReader(stream, Encoding.ASCII, detectEncodingFromByteOrderMarks: true, bufferSize: 1024, leaveOpen: true)) {
+            // Default system ASCII encoding will not throw exceptions, converting bad bytes to '?' instead
+            // which is sufficient to correctly parse the magic comments.
+            // Also StreamReader will automatically switch to proper Unicode encoding (non-throwing) when BOM present.
+            Encoding bootstrapEncoding = Encoding.ASCII;
+            using (StreamReader sr = new StreamReader(stream, bootstrapEncoding, detectEncodingFromByteOrderMarks: true, bufferSize: 1024, leaveOpen: true)) {
                 string line;
-                line = ReadOneLine(sr, ref bytesRead);
+                line = sr.ReadLine();
```
slozier: CPython allows arbitrarily long lines in it's coding check?

Src/IronPython/Runtime/PythonContext.cs
```diff
@@ -936,125 +936,79 @@ private sealed class OptimizedBuiltinHasher {
...
                 linesRead++;
 
                 // magic encoding must be on line 1 or 2
                 // if there is any encoding specified on line 1 (even an invalid one), line 2 is ignored
-                if (line != null && (encodingName = Tokenizer.GetEncodingNameFromComment(line)) == null) {
+                if ((encodingName = Tokenizer.GetEncodingNameFromComment(line)) == null) {
```
slozier: Any specific reason for remove the `line != null`? I guess it's an insignificant optimization...

BCSharp: IMHO simpler code and moves the check to `GetEncodingNameFromComment`, which becomes more robust to input.

BCSharp:
1. The error types are correct, it's just the messages.
1.
1. Yes, I was planning to post a question in the CPython repository or on their mailing list to get an idea what the required Python behaviour is. In the meantime, I could commit a small update to the test that removes the ambiguity and makes the test passing. This will allow us to enable the test and prevent the regressions. When the issue becomes clear, I would undo the patch and commit a proper fix. OK?
I will also enable `test_pep3120`, which interestingly tests the content of one of the error messages that is not exactly like CPython's but luckily it tests it in such a lax way that the check passes nevertheless.

Src/IronPython/Runtime/PythonContext.cs
BCSharp: Never tested whether there is a limit on a comment line. But there is no limit on any other line, as `test_source_encoding.test_file_parse` attests.

BCSharp: I haven't found it specified anywhere, but tests with CPython show that the lines are truncated after 511 characters. I guess we better follow this convention. This has to be done through a `StreamReader` though.

BCSharp: This may result in a truncated encoding name, for instance `# ... coding: latin-15` becomes `latin-1` if `5` happens to be in the 512-th column.

Src/IronPython/Runtime/PythonContext.cs
BCSharp: In CPython, the second line is not inspected if there is no `\n` or `\r` within the first 512 characters of the first line.
Perhaps this is an optimization to avoid scanning large binary files.
So the test `line != null` has to stay. I will add a test case to cover it.

Src/IronPython/Runtime/PythonContext.cs
BCSharp: Correction: not 512 characters but 512 bytes. However, the BOM bytes do not count to the limit.

BCSharp: What do you think of dropping support for widechar Unicode in source files (utf-16*, utf-32)? The code will become simpler (I won't have to use `StreamReader`). AFAIK, CPython does not support widechar Unicode.

slozier: I would be fine with dropping wide character encodings since PEP263 seems to talk about requiring ASCII compatible encodings. Assuming it hasn't been added in later versions (e.g. 3.8) of Python that is.

We don't have to match the 512-byte limit exactly, I was just thinking that some sort of limit would be useful since we're supposed to be quickly probing the file for an encoding.

slozier:
3. Yes, please update the test file to get it working, preferably add a comment with `IronPython` in it so that it's easy to spot where a change was done.

Src/IronPython/Runtime/PythonContext.cs
BCSharp: Great. I could set the limit to 512 characters per line, which will be 512 bytes for non-unicode encodings, and 2K for UTF-32* at the high end (or an extreme and unlikely worst case for UTF-8).

However, I am thinking that if CPython ever adds support for widechar Unicode source code, the encoding detection will probably not be PEP263 but maybe BOM. PEP263 is an old scheme that does not really scale well to non-ASCII compatible encodings. On the other hand, CPython already recognizes UTF-8 BOM so it is natural to extend this to other Unicode encodings as well. Another possibility here is how CPython 3.8 detects [widechar Unicode in JSON](https://bugs.python.org/issue17909) (building on [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt)). The assumptions there that the first character is ASCII or a quote are still true for a valid Python program. So in the end there is no need to extend support for PEP263 to widechar Unicode, especially that we already require BOM be present for widechar Unicode detection so PEP263 does not add anything useful.

Which approach do you prefer?

slozier: If CPython ever does add support then who knows what detection mechanism they'll pick, so I would just go with whatever implementation is simpler (presumably this is checking the BOM?).

BCSharp: Yes, checking the BOM is simpler. I will add a note to the "Differences with CPython" document stating that IronPython does support widechar Unicode based on BOM and that PEP263 does not apply in such case.

slozier:
Thanks for all your work on this "good first issue"!

BCSharp:
It's been a pleasure.

PR #612, opened 2019-04-30
Make test_source_encoding.test_issue2301 implementation independent

BCSharp:
This change is needed to get `test_source_encoding` working for IronPython.
It is a temporary fix until a proper solution to the compatibility issue is known and implemented.


slozier:
Thanks, CI failures are unrelated since I believe the test isn't enabled anyway?

BCSharp:
This change (python/cpython#13639) has been accepted upstream ([bpo-36919](https://bugs.python.org/issue36919)) and landed in Python 3.7 (python/cpython#15953) and 3.8 (python/cpython#15952) so it can be undone here once IronPython reaches v. 3.7.

PR #613, opened 2019-05-01
Fix some OSError subtype handling in the nt module

BCSharp:
This by no means fixes all `OSError` incompatibilites with CPython 3. It was meant to fix primarily `FileNotFoundError` on file/directory delete. This is order to get `AllCPython.test_source_encoding` passing.

Ideally, proper subclass creation should happen in a more generic place (maybe in `CreatePythonThrowable`?) so that e.g. `OSError(2, 'xxx')` in Python automatically creates `FileNotFoundError(2, 'xxx')` instead.


slozier:
Thanks for the PR. I find the `IOExceptionToPythonException` method a bit odd. What about a something like:
```C#
private static bool TryGetExceptionForOSError(int error, string filename, out Exception exception)
```
Anyway I'm not doing to block the PR on it. Will merge later today.

BCSharp:
I agree with you on the function signature, this idea crossed my mind as well. The reason I left it as it is, is that the switch statement is clearly incomplete and I didn’t know whether in some other cases the original CLR exception might not be needed. Then I realized that this is probably not the best place for it anyway so this whole method will be redesigned later. 

There are a few more points like this one where I stopped short from providing a satisfying solution as I had limited time available. I will have little time available for this project in the coming weeks and for now simply want to complete #610. If you want, you are more than welcome to continue working on #613, it is open for commits from the maintainers, or just merge it for now and we will working on `OSError` later.

BCSharp:
Thanks!

PR #616, opened 2019-05-02
Enable AllCPython.test_source_encoding

BCSharp:
Tests for PEP-263 and PEP-3120 (#22).

### DIFF
--- a/Src/IronPython/Runtime/ByteArray.cs
+++ b/Src/IronPython/Runtime/ByteArray.cs
@@ -75,7 +75,7 @@ namespace IronPython.Runtime {
             throw PythonOps.TypeError("string argument without an encoding");
         }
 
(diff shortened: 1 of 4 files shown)


BCSharp:
The reason that `PROCESS` isolation level is required is because `CaseGenerator` reads the content of the test file with `File.ReadAllText` and then `CaseExecutor` handles it with `CreateScriptSourceFromString`. This bypasses the Python coding detection machinery. If `CreateScriptSourceFromString` is replaced by `CreateScriptSourceFromFile`, `PROCESS` isolation level is not required. If performance is an issue, this can be done only to `ENGINE` isolation level. Or we can leave is at `PROCESS`, if it is not significant.

As for the failure on .NET Core, I am still investigating.

slozier:
Thanks!

PR #617, opened 2019-05-12
Implement cast() on memoryview

gfmcknight:
Refactor MemoryView to refer to the beginning and ending byte of the
underlying buffer rather than the index of the underlying buffer.
Implement cast() to change the shape and format of the memoryview. Add
tests in IronPython.test_memoryview for this behavior.

Hi, sorry for showing up with such a large commit. I understand if this isn't a high priority since we don't have much by way of C buffers. In any case, I want to get some feedback on whether this change is warranted at the moment and whether the approach is sufficiently clear and/or acceptable before proceeding.

Motivating behavior: with cast() it's possible to have a memoryview that is not aligned with the underlying buffer.

General approach: make `_start` and `_end` refer to the actual byte locations in the buffer. When getting any item or setting any item, always convert the objects of the buffer/memoryview to bytes, and then convert those bytes into the format of the memoryview/buffer.

Here are the specific questions I have:
1) Is this behavior worth supporting at the moment?
2) What do you think about the approach of directing everything to these getByteRange()/setByteRange() methods?
3) As implemented, I have a ton of fields on the memoryview that are null at first and get set after slicing/casting. Is it better to just initialize these to the underlying or default values in the first constructor?
4) The itemsize is a BigInteger, but I treat it exclusively as an integer here, opting to cast to an int for everything. This simplified things for me, but also I think that for any size not in the int range, we will run up against C#'s maximum array size. Is this something that can stay if we do proceed with this change? 

I intend to enable test_memoryview (the CPythonTest) in this PR. I got it working except for some stuff having to do with `sys.getrefcount()` and a weakref that wouldn't release on net45 (in a way that didn't break the other active tests -- this change is currently on master in my fork). I just don't want to show up asking for you to review with an even larger, more disorganized changeset before getting your opinion on the approach for this.

### DIFF
--- a/Src/IronPython/Runtime/MemoryView.cs
+++ b/Src/IronPython/Runtime/MemoryView.cs
@@ -9,13 +9,20 @@ using Microsoft.Scripting.Runtime;
 
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
+using System.Collections.Generic;
 
 namespace IronPython.Runtime {
     [PythonType("memoryview")]
     public sealed class MemoryView : ICodeFormattable {
+        private const int MaximumDimensions = 64;
+
         private IBufferProtocol _buffer;
         private readonly int _start;
         private readonly int? _end;
+        private int? _step;
+        private readonly string _format;
+        private readonly PythonTuple _shape;
+        private readonly int? _itemsize;
 
         public MemoryView(IBufferProtocol obj) {
             _buffer = obj;
@@ -28,6 +35,16 @@ namespace IronPython.Runtime {
 
         public MemoryView(MemoryView obj) : this(obj._buffer, obj._start, obj._end) { }
 
+        internal MemoryView(IBufferProtocol @object, int start, int? end, int? step, string format, PythonTuple shape) : this(@object, start, end) {
+            _format = format;
+            _shape = shape;
+            _step = step;
+
+            if (TypecodeOps.IsTypecodeFormat(format)) {
+                _itemsize = TypecodeOps.GetTypecodeWidth(format[0]);
+            }
+        }
+
         private void CheckBuffer() {
             if (_buffer == null) throw PythonOps.ValueError("operation forbidden on released memoryview object");
         }
@@ -35,9 +52,9 @@ namespace IronPython.Runtime {
         public int __len__() {
             CheckBuffer();
             if (_end != null) {
-                return _end.Value - _start;
+                return (_end.Value - _start) / ((int)itemsize * (_step ?? 1));
             }
-            return _buffer.ItemCount;
+            return _buffer.ItemCount * (int)_buffer.ItemSize / ((int)itemsize * (_step ?? 1));
         }
 
         public object obj {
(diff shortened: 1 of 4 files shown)


Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -0,0 +1,105 @@
...
+                case "q": // signed long long
+                case "Q": // unsigned long long
+                case "d": // double
+                    return true;
+                default:
+                    return false;
+            }
+        }
+
+        public static int GetTypecodeWidth(char typecode) {
```
slozier: Looking at usage it seems like you mostly do things like `if(IsTypecodeFormat(typecode)) GetTypecodeWidth(typecode)`. I wonder if using the Try pattern would be better suited `bool TryGetTypecodeWidth(typecode, out int width)`? Same comment for the other methods.

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -0,0 +1,105 @@
...
+                    throw PythonOps.ValueError("Bad type code (expected one of 'c', 'b', 'B', 'u', 'H', 'h', 'i', 'I', 'l', 'L', 'f', 'd')");
+            }
+        }
+
+        public static object FromBytes(char typecode, byte[] bytes, int offset) {
+            switch (typecode) {
+                case 'c': return (char)bytes[offset];
+                case 'b': return (sbyte)bytes[offset];
+                case 'B': return bytes[offset];
+                case 'u': return BitConverter.ToChar(bytes, offset);
```
slozier: Do we need to worry about endianness?

gfmcknight:
Thanks for the great feedback! Not to worry about the full review yet -- this change still has a bit of a way to go.
2. I'll add a field named something like `matchesBuffer` and then will fast-track in those cases.
3. I'll switch to initialization to good default values because it'll make things cleaner, especially with `_step`.

Src/IronPython/Runtime/TypecodeOps.cs
gfmcknight: I'm not sure. BitConverter uses the endianness of the system, which, I believe, is what we want.

Endianness will for sure break the tests I added in test_memoryview.

Tests/test_memoryview.py
```diff
@@ -117,4 +124,108 @@ def test_equality(self):
...
+        a = array.array('b', range(16))
+        mv = memoryview(a).cast('b', (4,2,2))
+        mv2 = mv[2:]
+        self.assertEqual(len(mv2), 2)
+        for i in range(2):
+            for j in range(2):
+                for k in range(2):
+                    self.assertEqual(mv[(i + 2, j, k)], mv2[(i, j, k)])
+
+#run_test(__name__)
```
gfmcknight: This needs to be reinserted and the test.support version needs to be removed.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -122,19 +153,236 @@ public sealed class MemoryView : ICodeFormattable {
...
+
+            bool thisIsBytes = this.format == "B" || this.format == "b" || this.format == "c";
+            bool otherIsBytes = formatAsString == "B" || formatAsString == "b" || formatAsString == "c";
+
+            if (!thisIsBytes && !otherIsBytes) {
+                throw PythonOps.TypeError("memoryview: cannot cast between two non-byte formats");
+            }
+
+            int newItemsize;
+            if (!TypecodeOps.TryGetTypecodeWidth(formatAsString, out newItemsize)) {
```
gfmcknight: This check should probably happen way earlier.

slozier:
Looks good overall. :+1:

gfmcknight:
Looks like multidimensional memoryviews return the length of the first dimension in `__len__`.

In it's current state memoryview as a whole is missing a few things (some general non-cast-related things, and also index access using a tuple). There are tests for these in CPython.test_memoryview that don't yet pass. Should I make a separate pull request for this set of changes?

slozier:
Yes for the separate PR. I'll merge this one once the CI is done assuming there are no issues.

gfmcknight:
Sounds good -- the other changes will be much more atomic than this one.

slozier:
Thanks!

PR #618, opened 2019-05-14
Fix $Env:PATH handling

BCSharp:
I've finally found the cause for some of the tests failing on my local machine, while they were fine on CI. After a number of local test runs, the PATH variable became too long to successfully spawn subprocesses for tests at PROCCESS isolation level.


slozier:
Thanks!

PR #619, opened 2019-05-19
Make codecs tests runnable under CPython

BCSharp:
This is primarily making `IronPython.test_codecs` and `IronPython.modules.io_related.test_codecs` runnable under CPython, to make sure that the assumptions made in those tests actually hold. And indeed there was one case where the test was assuming wrong. No practical significance, but still it demonstrated the validity of checking with CPython. Perhaps it would be good to extend the CI to run IronPython tests under CPython on each pull request. The tests can use `@unittest.skipUnlessIronPython` or conditionals `is_ironpython` or `is_cpython` to differentiate behaviour if needed. Any such difference will then be explicit.

Further, I (re)enabled some tests for cases that are no longer problematic or could easily be fixed (fixes included).

### DIFF
--- a/Tests/modules/io_related/test_codecs.py
+++ b/Tests/modules/io_related/test_codecs.py
@@ -16,11 +16,12 @@ import codecs
 import os
 import re
 import shutil
+import string
 import subprocess
 import sys
 import unittest
 
-from iptest import IronPythonTestCase, is_cli, is_mono, is_netcoreapp21, is_posix, run_test, skipUnlessIronPython
+from iptest import IronPythonTestCase, is_cli, is_mono, is_netcoreapp21, is_posix, run_test, skipUnlessIronPython, is_netcoreapp
 from iptest.misc_util import ip_supported_encodings
 
 class CodecTest(IronPythonTestCase):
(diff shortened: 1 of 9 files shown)


Tests/modules/io_related/test_codecs.py
```diff
@@ -470,10 +473,18 @@ def run_python(filename):
 
         self.assertEqual(len(t_err_lines), 0)
         if is_cli:
-            print("CodePlex 11334")
-            self.assertEqual(t_out_lines[0].rstrip(), b"\xe6ble")
+            # CodePlex 11334: IronPython uses active console codepage for output
+            # The test below assumes cp850, which is default on Western European/Northern American Windows
```
slozier: I'm not sure that this is necessarily correct. On my system the terminals are cp437 and the test is skipped even if it's valid. I'm not sure what encoding the CI uses. Would be nice not to have something so specific.

`not is_netcoreapp` can probably be replaced with `import clr; clr.AddReference("System.Console")`?

Tests/test_codecs.py
```diff
@@ -9,11 +9,14 @@
 import unittest
 import codecs
 
-import System.Text
+from iptest import run_test, is_ironpython
```
slozier: I think `is_cli` is used throughout the tests. Might be better to use it here for consistency. Would be nice to eliminate one of them at some point.

BCSharp: Most tests use `is_cli`, a few tests use both or only `is_ironpython` (`test_class`, `test_excinfo`, `test-exec`). It seems that originally `is_ironpython` was defined as `is_silverlight or is_cli`. With the drop of silverlight, they are now equivalent. It is a good idea to eliminate one of them. I am OK either way: `is_ironpython` is more informative, `is_cli` is shorter. It will be less work to settle on `is_cli` as the changes will be limited. If you are OK with that, I can include those changes in this commit.

slozier: Ah that explains it. Yes I'm fine with removing `is_ironpython` in this PR. Eventually it might be nice to rename `is_cli` to `is_ironpython` but I'd wait until we diverge more from or stop supporting ipy2.

Tests/modules/io_related/test_codecs.py
BCSharp: My mistake. I've discovered it by chance and now try to reverse-engineer the behaviour. On my system the console uses cp65001 and `PYTHONIOENCODING` is set to `utf-8:surrogateescape`. As a result, both IronPython and CPython were failing this test. My intention was to have this test skipped on my system (unless I reconfigure a console for testing) but pass on CI. Also, it should not fail on any other system that somebody may check it out on. 

What I propose: replace the constant `b"\xe6ble"` with a `'µble'` encoded according to the actual codepage in use. 

On a side note: since the console codepage turns out to be a significant enviromental factor, maybe display it in the "Version Information" section on CI? I also notice that "Version Information" section is missing in Windows CI builds.

As for CPython, things are also not that simple as my simple test condition suggests, esp. on POSIX and higher versions of Python (interplay of `LC_ALL`, `LC_CTYPE`, `PYTHONUTF8`, `PYTHONIOENCODING`, `PYTHONCOERCECLOCALE`, and maybe more). Again, my intention was to make this test runnable at least on CI and in my environment. Luckily (?...) the test is skipped on POSIX for an unrelated reason. The test may have to be revised once `Popen` on POSIX becomes available.

> `clr.AddReference("System.Console")`

Good idea.

slozier: Yes, maybe something like:
```Python
self.assertEqual(t_out_lines[0].rstrip(), "\xb5ble".encode(System.Console.OutputEncoding))
```

slozier:
Thanks!

PR #623, opened 2019-05-24
Get matching encoding names

slozier:
@BCSharp See any issues with this change?


BCSharp:
Looks good to me. I think you've fixed a copy-paste bug in `PythonEncoding`.

PR #624, opened 2019-05-24
Update to preview5

slozier:


slozier:
@BCSharp I'm getting some failures with `TestUtf8` and `TestIncrementalWithUtf8` on .NET Core 3 preview5. Any ideas?

BCSharp:
I see nothing wrong with the tests themselves. Also nothing obvious in `PythonEncoding` as yet. Will need to investigate deeper, perhaps tomorrow.

BCSharp:
In my assessment, it is a bug in .NET Core. When using the incremental UTF-8 decoder, the index of the error byte is not always reported correctly. This is being checked by the `IronPythonTest.SurrogatePassTest.DecodingTests.TestUtf8` test. The `TestIncrementalWithUtf8` fails because 'surrogatepass' error handler depends on the index being correct to detect a valid surrogate.

I am going to submit an issue report in dotnet/corefx.

BCSharp:
dotnet/corefx#37969

slozier:
@BCSharp Thanks for looking into this.

slozier:
Merging this in even with the failures. I tried the failing tests with a daily build of .NET Core and the fix appears to work.

Issue #544, opened 2019-02-18
weakref not released when expected


slozier:
This causes `test_deque.test_container_iterator` to fail on Mono.

slozier:
Test disabled in https://github.com/IronLanguages/ironpython3/pull/540

BCSharp:
`gc,collect()` on Mono seems not to have everything collected on return. For this reason, `IronPython.test_memoryview.test_finalizer` is disabled for Mono.

slozier:
I didn't look at it in details, but I believe `CPython.test_set` has the same issue. https://github.com/IronLanguages/ironpython3/pull/849

BCSharp:
Test skipped in #1851

Issue #489, opened 2018-11-16
Some tests are failing via the test runner but not when run directly


slozier:
Both `AllCPython.test_dynamic` and `AllCPython.test_webbrowser` are passing when run directly using `ipy` but fail with the test runner. Figure out why.

slozier:
Note that these will pass with `IsolationLevel=PROCESS`.

slozier:
Also, `AllCPython.test_deque` (https://github.com/IronLanguages/ironpython3/pull/516).

slide:
Can we turn on redirection of stdout/stderr?

slozier:
For some reason I can't get test_webbrowser to fail locally, will try with a PR. Here are the other errors.

test_dynamic
---
```
Error Message:
 Traceback (most recent call last):
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\test_dynamic.py", line 143, in <module>
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\test_dynamic.py", line 139, in test_main
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\support\__init__.py", line 1812, in run_unittest
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\support\__init__.py", line 1787, in _run_suite
test.support.TestFailed: Traceback (most recent call last):
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\support\__init__.py", line 2029, in swap_item
    yield
NameError: name 'next' is not defined
```
```
Error Message:
 Traceback (most recent call last):
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\test_dynamic.py", line 143, in <module>
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\test_dynamic.py", line 139, in test_main
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\support\__init__.py", line 1812, in run_unittest
  File "R:\Documents\Programming-github\ironpython3\Src\StdLib\Lib\test\support\__init__.py", line 1787, in _run_suite
test.support.TestFailed: Traceback (most recent call last):
[...]

PR #625, opened 2019-06-06
Get test_memoryview passing

gfmcknight:
... Kind of.

I have a couple pieces of the implementation that I think need revision, namely:
1) When I added weakref support to memoryview, for some reason it wasn't releasing in time for the net4.5 target when run through make.ps1 tests. If I ran test_memoryview in any other circumstance the weakref seemed to release appropriately. I left it out for now, but it might be a good idea to investigate, given that we've seen something similar in #544.
2) The change to suboffsets (empty tuple instead of None) should probably be handled on the other side of the IBufferProtocol instead of in memoryview, right?


Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -147,18 +147,44 @@ public sealed class MemoryView : ICodeFormattable {
...
                 CheckBuffer();
-                return _buffer.SubOffsets;
+                return _buffer.SubOffsets ?? PythonOps.MakeTuple();
             }
         }
 
         public Bytes tobytes() {
             CheckBuffer();
-            return _buffer.ToBytes(_start, _end);
+            byte[] bytes = getByteRange(_start, numberOfElements() * (int)itemsize);
```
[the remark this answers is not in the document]
gfmcknight: Maybe -- I think that the way each of the IBufferProtocol implementations handle ToBytes() is probably better than what getByteRange() does.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -390,6 +416,12 @@ public sealed class MemoryView : ICodeFormattable {
                     throw PythonOps.NotImplementedError("multi-dimensional sub-views are not implemented");
                 }
 
+                // We expect to get a numeric value when setting a paricular
+                // element, which we will handle by breaking up into bytes
+                if (!(value is int)) {
```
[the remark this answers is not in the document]
gfmcknight: Ah I see; maybe I should be using something like `Converter.TryConvertToInt32()`.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -282,6 +282,16 @@ public static class StringOps {
             }
         }
 
+        [StaticExtensionMethod]
+        public static object __new__(CodeContext/*!*/ context, PythonType cls, [NotNull]MemoryView @object) {
```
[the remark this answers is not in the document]
gfmcknight: You're right; I had a default argument for the encoding on the other `__new__`, which should be there.

Src/StdLib/Lib/test/test_memoryview.py
```diff
@@ -26,7 +26,8 @@ def _types(self):
 
     def check_getitem_with_type(self, tp):
         b = tp(self._source)
-        oldrefcount = sys.getrefcount(b)
+        with self.assertRaises(AttributeError):
```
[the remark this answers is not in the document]
gfmcknight: Sounds good!

gfmcknight:
1. If adding `gc.collect()` works to fix the weakref, would you also be okay with adding that into the test in case the implementation is ironpython or do you think it's better to leave the test disabled?

slozier:
Yes, you can just add something like `gc.collect() # required by IronPython`, probably no need to conditionally exclude. I did something like that in `test_deque` (although I think it still fails once in a while).

Src/IronPython/Runtime/MemoryView.cs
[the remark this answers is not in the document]
gfmcknight: Or Int64 -- since the opcode can have 8 bytes. I need to add a test for this.

gfmcknight:
It still consistently failed when run from make.ps1. The test currrently calls `support.gc_collect()`. Maybe it has to do with Release vs Debug version?

Src/IronPython/Runtime/MemoryView.cs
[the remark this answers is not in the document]
gfmcknight: It looks like the diff didn't update here. What do you think about using `_matchesBuffer` in this case?
```
if (_matchesBuffer && _step == 1) {
    return _buffer.ToBytes(_start / (int)itemsize, _end / (int)itemsize);
}
```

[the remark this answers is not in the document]
slozier: That seems fine, minor tweak would be to use `_itemsize` instead of `(int)itemsize`. We should probably do that everywhere anyway.

slozier:
It looks like changing the IsolationLevel seems to make `test_weakref` work. I'm fine with either skipping it (as long as an issue is filed -- even if it just says the test if failing for unknown reasons -- and you add a link to it in a comment in the source) or changing the IsolationLevel in AllCPythonCasesManifest.ini:
```
IsolationLevel=PROCESS # https://github.com/IronLanguages/ironpython3/issues/489
```

gfmcknight:
Just a quick update since I left this hanging so long. Since test_file seems to be a case of a weakref not releasing as well, I think they should be in the same status so I think I'll mark set IsolationLevel=Process as above.

The remaining hangup that I'm working on is that my conversion to an int (line 421) doesn't work above for floats and the other types, and I'm trying to come up with a clean way of giving the correct responsibility of error handing to MemoryView and TypecodeOps.

slozier:
Sounds like you might need a switch based on format to do the proper conversions?

gfmcknight:
I guess I think it should be done in MemoryView. This is a little closer.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -265,12 +265,22 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
...
+                } else if (o is Bytes b) {
+                    return b._bytes; // CData returns a bytes object for its type
+                } else {
+                    throw PythonOps.NotImplementedError("No conversion for type {0} to byte array", PythonOps.GetPythonTypeName(o));
+                }
+            } catch (OverflowException) {
+                bool valueError = Converter.TryConvertToBigInteger(o, out _);
+
+                if (valueError) {
+                    throw PythonOps.ValueError("memoryview: invalid type for format '{0}'", format);
```
slozier: Should this read "invalid value"  instead of "invalid type"?

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -265,12 +265,22 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
         }
 
         private byte[] unpackBytes(string format, object o) {
-            if (TypecodeOps.TryGetBytes(format, o, out byte[] bytes)) {
-                return bytes;
-            } else if (o is Bytes b) {
-                return b._bytes; // CData returns a bytes object for its type
-            } else {
-                throw PythonOps.NotImplementedError("No conversion for type {0} to byte array", PythonOps.GetPythonTypeName(o));
+            try {
```
slozier: Looking at usage in `setByteRange`/`getByteRange` this is called with `unpackBytes(_buffer.Format, _buffer.GetItem(index)`. Do we even need all this error checking in this case?

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -385,15 +395,53 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
             int firstByteIndex = _start + index * _itemsize * _step;
             object result = packBytes(format, getByteRange(firstByteIndex, _itemsize), 0, (int)_buffer.ItemSize);
 
-            // TODO: BigInteger should also be merged into an integer once the int/long merge occurs
-            if (result is BigInteger || result is double || result is float || result is Bytes) {
+            // TODO: BigInteger/long/ulong should also be merged into an integer once the int/long merge occurs
+           if (result is BigInteger || result is double || result is float || result is Bytes) {
```
slozier: `float` probably needs to be cast to `double`.

I wonder if there should be a .NET primitive to Python type helper somewhere? Maybe there already is...

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -385,15 +395,53 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
...
                 return result;
+            } else if (result is long || result is ulong) {
+                return Converter.ConvertToBigInteger(result);
             } else {
                 return Convert.ToInt32(result);
             }
         }
 
         private void setAtFlatIndex(int index, object value) {
+            switch (format) {
```
slozier: Won't this cause issues with the `_matchesBuffer` below? For example underlying buffer is a byte array but you convert to a long and call `_buffer.SetItem`.

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -123,6 +123,8 @@ class TypecodeOps {
                     result = BitConverter.GetBytes(Convert.ToSingle(obj));
                     return true;
                 case "d":
+                    double objAsDouble = 0;
```
slozier: What is this for? Doesn't seem to be used.

gfmcknight:
Seems like a serious case of me not cleaning up the things I started but didn't like. Sorry for making you review all of this!

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: You're right; the error checking should probably be done in `setAtFlatIndex` because that's the place where all of the external element setting eventually calls into

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: Each of the buffers handle overflows but they each do so differently; maybe MemoryView needs to do its own overflow check as well.

slozier: Ah I see, it basically ends up calling the setter on the type which does its own error checking.

If you look at the CPython example below, based on the error message, it seems like the type checking is done at the memoryview level:
```
>>> m = memoryview(bytearray(b"aaa"))
>>> m[0] = 256
ValueError: memoryview: invalid value for format 'B'
```

Seems to me like the `IBufferProtocol` interface is a bit too high level. Would be nice to have lower level access (like Span/Memory) to avoid all the overhead. Anyway, that's a problem for another time...

slozier:
I'm fine with reviewing, helps me get more familiar with the codebase. :)

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -390,7 +423,14 @@ public sealed class MemoryView : ICodeFormattable {
                     throw PythonOps.NotImplementedError("multi-dimensional sub-views are not implemented");
                 }
 
-                setAtFlatIndex(index, value);
+                // We expect to get a numeric value when setting a paricular
+                // element, which we will handle by breaking up into bytes
+                long valueAsLong = 0;
+                if (!Converter.TryConvertToInt64(value, out valueAsLong)) {
```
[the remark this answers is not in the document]
gfmcknight: Add a test to confirm it works now.

Src/IronPython/Runtime/TypecodeOps.cs
gfmcknight: Removed.

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: I added an overflow check in MemoryView so that should be hit first, instead of the `_buffer`'s type check.

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: Changed.

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -78,6 +79,12 @@ class TypecodeOps {
                 case "d":
                     result = BitConverter.ToDouble(bytes, offset);
                     return true;
+                case "q":
+                    result = (BigInteger)(BitConverter.ToInt64(bytes, offset));
```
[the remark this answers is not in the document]
gfmcknight: I put it in the commit accidentally; I meant to remove it.

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: I believe `GetPythonType` works for primitive types. I'm not sure if that's the functionality you're referring to.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -349,24 +382,124 @@ public sealed class MemoryView : ICodeFormattable {
...
                 return Convert.ToInt32(result);
             }
         }
 
+        /// <summary>
+        /// Verifies that the value being set does not overflow the format
+        /// for this MemoryView.
+        /// </summary>
+        /// <param name="value">The value to be set.</param>
+        private void checkNumeric(BigInteger value) {
```
gfmcknight: `checkForOverflow`. Should this be in TypecodeOps?

Src/IronPython/Runtime/MemoryView.cs
slozier: I meant something that essentially does what you do here: `byte`/`short`->`int`, `float`->`double`, `long`->`BigInteger`, and so on for all C# primitive types. `GetPythonType` does not do this since it will return you the Python version of the C# types (e.g. it will give you `System.Int16` for shorts).

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: `q` and `Q` seem to be flipped here.

slozier: Probably.

Might be lost in the noise, but it seems like we're doing a lot of converting back and forth: object -> BigInteger -> object -> primitive type -> byte[] -> primitive type -> object (and then SetItem does a bunch of the same work). Anyway, I guess we should get it working first before worrying about optimizations.

gfmcknight: Yeah I think removing some of the conversions here is a good idea, for the moment this conversion to BigInteger makes the overflow check easier.

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -118,10 +122,73 @@ class TypecodeOps {
...
+                    return true;
                 default:
                     result = null;
                     return false;
             }
         }
+
+        /// <summary>
+        /// Verifies that the value being set does not overflow the format
+        /// for this MemoryView.
```
gfmcknight: Not in MemoryView anymore.

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: I wasn't able to find anything that does this, so I added one such method in `PythonOps`. I'm not entirely sure what the correct location for this would be.

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: I was able to remove the conversion to BigInteger. Still thinking about the conversions to `long` and `ulong`, which are probably only necessary when we get a BigInteger value, to allow BitConverter to do what it wants.

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -188,7 +188,7 @@ class TypecodeOps {
                     return false; // All non-numeric types will not cause overflow.
             }
 
-            return value < minValue || value > maxValue;
+            return PythonOps.Compare(value, minValue) < 0 || PythonOps.Compare(value, maxValue) > 0;
```
slozier: Unfortunately calling `Compare` is probably even worse (performance wise) than converting to a `BigInteger`. Also, since `Compare` is equivalent to calling `__cmp__` it will be removed from the codebase at some point.

slozier:
Looking at the CI, it looks like the failures on Mono are weakref related? I would be fine with just skipping testing on Mono (by adding RunCondition=NOT $(IS_MONO) with a comment to AllCPythonCasesManifest.ini).

Src/IronPython/Runtime/TypecodeOps.cs
gfmcknight: I tried benchmarking setting, and indeed Compare is much slower. The current change is only a little faster (~10%) than converting to BigInteger (possibly within the margin of error -- it wasn't a very empirical process), but I think it's a little cleaner than expecting a BigInteger in TypecodeOps. Here's what I did to benchmark:
```
import time
import array

def benchmark_memoryview(mv):
    start_time = time.time()
    for j in range(2000):
        for i in range(4000):
            mv[i] = i + j
    end_time = time.time()
    return end_time - start_time

mv = memoryview(array.array('i', [0]*4001)).cast('b')[1:16001].cast('i')
print('int unaligned', benchmark_memoryview(mv))

mv = memoryview(array.array('i', [0]*4000))
print('int aligned', benchmark_memoryview(mv))
```
Also, the aligned benchmark was super slow because `_shape` was null, so I initialized it in the constructor.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -144,21 +144,52 @@ public sealed class MemoryView : ICodeFormattable {
...
+                    stridedBytes[indexStrided + j] = bytes[indexInBytes + j];
+                }
+            }
+
+            return Bytes.Make(stridedBytes);
         }
 
         public PythonList tolist() {
             CheckBuffer();
-            return _buffer.ToList(_start, _end);
```
slozier: Is this a spot where we could have a fast path with `_matchesBuffer && _step == 1`?

gfmcknight:
Other than the fast track, I don't think anything comes to mind. The last piece is indexing at a tuple, but I think that should be another pull request, and I don't believe it's blocking anything for IronPython.

Thanks for the patience and the helpful reviews with this one!

I believe that the remaining things to do after this is merged are:

- Add a comment to #544 
- Add a comment to #489

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: Added, but only tested against the two `test_memoryview` tests.

slozier:
Thanks for the PR!

Issue #451, opened 2018-10-16
unbound variable: $localContext


slozier:
```Python
def test():
    class two(object):
        def __init__(self):
            super().__init__()
            self.cnt = 0
    return two().cnt

print(test())
```
fails with
> SystemError: unbound variable: $localContext

slozier:
Another busted scenario I encountered while looking into this:
```Python
class two(object):
    def test(self):
        return __class__

a = two
two = "z"
assert a().test() == a().__class__
```

PR #630, opened 2019-07-10
 Update exception handling to Python3 definition

### LINKED ISSUE #245: Exception context issue

gfmcknight:
After exiting an except block, no longer treat the latest exception as
being "handled." Maintain the exception context when executing via a
generator.

Resolves #245.

---
It's still missing some major parts, but this is a minimal change that seems to get the tests described in the issue passing.


Src/IronPython/Compiler/Ast/AstMethods.cs
```diff
@@ -48,6 +48,7 @@ internal static class AstMethods {
         public static readonly MethodInfo GetExceptionInfoLocal = GetMethod((Func<CodeContext, Exception, PythonTuple>)PythonOps.GetExceptionInfoLocal);
         public static readonly MethodInfo CheckException = GetMethod((Func<CodeContext, object, object, object>)PythonOps.CheckException);
         public static readonly MethodInfo SetCurrentException = GetMethod((Func<CodeContext, Exception, object>)PythonOps.SetCurrentException);
+        public static readonly MethodInfo PopCurrentException = GetMethod((Action)PythonOps.PopCurrentException);
```
gfmcknight: This is no longer needed -- to be removed.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -2036,6 +2036,15 @@ public static partial class PythonOps {
             RestoreCurrentException(null);
         }
 
+        public static void PopCurrentException() {
```
gfmcknight: To be removed

Src/IronPython/Runtime/Generator.cs
```diff
@@ -408,6 +409,10 @@ public sealed class PythonGenerator : IEnumerator, IEnumerator<object>, ICodeFor
 
             RestoreFunctionStack();
 
+            // The generator may be in the process of handling an exception
+            // and therefore the generator needs to save its state
+            RestoreCurrentException(_exceptionContext);
```
gfmcknight: This only covers `next()` coming from Python and does not cover those from `IEnumerable.MoveNext()`

Src/IronPython/Compiler/Ast/TryStatement.cs
```diff
@@ -164,20 +170,23 @@ public class TryStatement : Statement {
...
             return Ast.Block(
-                GetVariables(lineUpdated, runElse), 
+                GetVariables(lineUpdated, runElse, previousExceptionContext),
                 AddFinally(result),
                 AstUtils.Default(typeof(void))
             );
         }
 
-        private static ReadOnlyCollectionBuilder<MSAst.ParameterExpression> GetVariables(MSAst.ParameterExpression lineUpdated, MSAst.ParameterExpression runElse) {
+        private static ReadOnlyCollectionBuilder<MSAst.ParameterExpression> GetVariables(MSAst.ParameterExpression lineUpdated, MSAst.ParameterExpression runElse, MSAst.ParameterExpression previousExceptionContext) {
```
gfmcknight: There's gotta be utility for this...

Src/IronPython/Runtime/Generator.cs
```diff
@@ -24,6 +24,7 @@ public sealed class PythonGenerator : IEnumerator, IEnumerator<object>, ICodeFor
         private readonly PythonFunction _function;                  // the function which created the generator
         private readonly MutableTuple _data;                        // the closure data we need to pass into each iteration.  Item000 is the index, Item001 is the current value
         private readonly MutableTuple<int, object> _dataTuple;      // the tuple which has our current index and value
+        private Exception _exceptionContext;                        // the exception being handled in the generator itself
```
gfmcknight: Do I need to initialize this to the current exception at the start?

slozier:
Looks like I had a local branch for this issue as well. For comparison: https://github.com/slozier/ironpython3/commit/9f68deaf5cfb0b35b927c1d57adc1d682fbaff31

Src/IronPython/Runtime/Generator.cs
slozier: I think I did that in mine? https://github.com/slozier/ironpython3/commit/9f68deaf5cfb0b35b927c1d57adc1d682fbaff31

Src/IronPython/Compiler/Ast/TryStatement.cs
slozier: No idea, could be something like:
```C#
private static ReadOnlyCollectionBuilder<MSAst.ParameterExpression> GetVariables(params MSAst.ParameterExpression[] expr)
    => new ReadOnlyCollectionBuilder<MSAst.ParameterExpression>(expr.Where(e => e != null));

```

Src/IronPython/Runtime/Generator.cs
slozier: Hmm, maybe? Not sure if this example relates...
```Python
def test4():
    def gen():
        try:
            raise Exception
        except:
            pass
        yield sys.exc_info()
    x = gen()
    try:
        raise Exception
    except:
        exc_info = sys.exc_info()
        assert next(x) == exc_info

test4()
```

Src/IronPython/Runtime/Generator.cs
gfmcknight: Looks like it

Src/IronPython/Compiler/Ast/TryStatement.cs
gfmcknight: Would this belong in `AstUtils`?

slozier: I'm not really seeing a use case outside of `TryStatement` so probably not worth it.

Src/IronPython/Runtime/Generator.cs
gfmcknight: I did not expect that behavior -- we get the exception context from the first call to `next()`?

slozier: I don't think this it's getting the exception from the first call to `next()`. It's basically whatever exception is currently being handled:
```Python
def test5():
    def gen():
        yield sys.exc_info()
        try:
            raise Exception
        except:
            pass
        yield sys.exc_info()
    x = gen()
    assert next(x) == (None, None, None)
    try:
        raise Exception
    except:
        exc_info = sys.exc_info()
        assert next(x) == exc_info

test5()
```

gfmcknight: Ah I see -- it looks like the handling is mostly intuitive. One edge case we might need to look out for is these two:
```
def test6():
    def gen():
        try:
            raise Exception
        except:
            yield traceback.format_exc()
        try:
            raise Exception
        except:
            yield traceback.format_exc()
        
    x = gen()
    try:
        raise Exception('@Exception 1')
    except:
        assert '@Exception 1' in next(x)
    try:
        raise Exception('@Exception 2')
    except:
        res = next(x)
        assert '@Exception 1' in res
        assert '@Exception 2' not in res

test6()

def test7():
    def gen():
        try:
            raise Exception
        except:
            yield traceback.format_exc()
        yield "Hello"
        try:
            raise Exception
        except:
            yield traceback.format_exc()
        
    x = gen()
    try:
        raise Exception('@Exception 1')
    except:
        assert '@Exception 1' in next(x)
    next(x)
    try:
        raise Exception('@Exception 2')
    except:
        res = next(x)
        assert '@Exception 2' in res
        assert '@Exception 1' not in res

test7()
```

gfmcknight: But this probably just follows from the implementation of try/catch that we have.

slozier: Hmm, test6 is unexpected. It passes with 3.4 but fails with 3.7.3. Might have been a bug in the CPython implementation?

Personally I would prefer to match 3.7 on this one, but I wouldn't go out of my way to do it at this point in time.

gfmcknight: I'm using 3.6 so it looks like it was probably changed in 3.7. I'll try to get the error inheritance to work first.

gfmcknight:
Quick update about the state of this PR: it looks like in IronPython, `traceback.format_exc()` leaks a `StopIterationException` to `PythonOps`/`sys.exc_info()`, with or without this change. It's a little challenging to debug, but I want to take it on. I think it fits under this issue since the function is another generator/exception context interaction.

```
try:
    raise Exception
except:
    print(traceback.format_exc())
    print(traceback.format_exc())
```

gfmcknight:
It looks to me like `yield from` is causing this behavior:
```
def test8():
    def gen1():
        yield 1
        yield 2
        yield 3
    def gen2():
        yield 4
        yield from gen1()
        yield 5
    x = list(gen2())
    assert x == [4,1,2,3,5]
    assert sys.exc_info() == (None, None, None)

test8()
```

temp_tests.py
```diff
@@ -0,0 +1,224 @@
...
+    x = gen()
+    try:
+        raise Exception('@Exception 1')
+    except:
+        assert '@Exception 1' in next(x)
+    try:
+        raise Exception('@Exception 2')
+    except:
+        res = next(x)
+        assert '@Exception 1' in res
```
gfmcknight: This is the version of the test as it applies to pre-3.7

Src/IronPython/Compiler/Ast/TryStatement.cs
```diff
@@ -114,6 +115,7 @@ public class TryStatement : Statement {
                 result =
                     Ast.Block(
                         Ast.Assign(runElse, AstUtils.Constant(true)),
+                        Ast.Assign(previousExceptionContext, Ast.Call(AstMethods.SaveCurrentException)),
```
gfmcknight: I left this in the try unlike in your implementation. I think that in either case we can't save the exception context like this because the underlying exception can change. It needs some sort of callback that can be made to restore the exception correctly if in a generator...

gfmcknight:
Current status is that I need to look into line numbers and how they are managed because I haven't yet paid any attention to them. Also, I need to do a bit of reorganizing the changes to `TryStatement` and update the comments in the code and the tests. However, the behavior appears to be mostly what we're aiming for. I have a few questions now:

- I think that the exception save/restore in `FunctionDefinition` is now useless because the try/catch always handles the exception context. Should we try removing that and the `CanSetSysExcInfo` property?
- Do you have any opinions or concerns at the approach to generators and exceptions? Namely where I tacked on this extra "callback" (`ExceptionMetadata`).
- Do you have any suggestions for line number tests that don't break every test when a new line is added? My thought is to use `eval`, catch thrown exceptions, and assert the line number.

Src/IronPython/Compiler/Ast/TryStatement.cs
```diff
@@ -17,6 +17,7 @@
 using MSAst = System.Linq.Expressions;
 
 using AstUtils = Microsoft.Scripting.Ast.Utils;
+using IronPython.Runtime.Operations;
```
slozier: nit: move the using to follow `using IronPython.Runtime.Binding;`

Src/IronPython/Runtime/Generator.cs
```diff
@@ -24,10 +24,13 @@ public sealed class PythonGenerator : IEnumerator, IEnumerator<object>, ICodeFor
         private readonly PythonFunction _function;                  // the function which created the generator
         private readonly MutableTuple _data;                        // the closure data we need to pass into each iteration.  Item000 is the index, Item001 is the current value
         private readonly MutableTuple<int, object> _dataTuple;      // the tuple which has our current index and value
+        private ExceptionMetadata _exceptionContext;                // the exception being handled in the generator itself
+        private ExceptionMetadata _save;
         private GeneratorFlags _flags;                              // Flags capturing various state for the generator
 
         private GeneratorFinalizer _finalizer;                      // finalizer object
 
+
```
slozier: nit: remove blank line

Src/IronPython/Runtime/Generator.cs
```diff
@@ -337,11 +341,17 @@ public sealed class PythonGenerator : IEnumerator, IEnumerator<object>, ICodeFor
...
             } else {
-                Exception save = SaveCurrentException();
+                _save = SaveCurrentException();
+                RestoreCurrentException(_exceptionContext);
+
                 try {
                     return MoveNextWorker();
                 } finally {
-                    RestoreCurrentException(save);
+
```
slozier: nit: remove blank line

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -38,6 +38,11 @@
 
 namespace IronPython.Runtime.Operations {
 
+    public class ExceptionMetadata {
```
slozier: Maybe make this internal? Does this need to be mutable (i.e. can we get rid of the setters and use constructors instead)?

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -2027,23 +2032,38 @@ public static partial class PythonOps {
             }
 #endif
 
-            RawException = clrException;
+            CurrentException = new ExceptionMetadata { Exception = clrException };
+            return res;
+        }
+
+        public static object SetCurrentExceptionWithCallback(CodeContext/*!*/ context, Func<Exception>/*!*/ callback) {
```
slozier: Is this used anywhere?

temp_tests.py
```diff
@@ -0,0 +1,236 @@
+import traceback
```
slozier: Presumably this will be changed to use unittest and be moved to a proper test file so that it runs with the CI?

Tests/test_excinfo.py
```diff
@@ -7,7 +7,7 @@
 import sys
 import unittest
 
-from iptest import is_cli, run_test
+#from iptest import is_cli, run_test
```
slozier: Might as well just delete the code instead of commenting it out.

Tests/test_excinfo.py
```diff
@@ -69,7 +69,7 @@ def f():
             try:
                 raise ValueError(63)
             except t():  # not matching
-                Assert(False)
+                assert False
```
slozier: Use `self.fail()` instead? Same comment in other places where you have `assert False`.

Tests/test_excinfo.py
```diff
@@ -371,94 +370,27 @@ def f():
...
-            self.A(None)
         self.A(None)
-
+   
     #========================================================
     # With's Pep (http://www.python.org/dev/peps/pep-0343/) says the
     # __exit__ can be invoked by an except block,
     # but unlike a normal except, that shouldn't set sys.exc_info().
 
+    @unittest.skip('unbound variable: $localContext')
```
slozier: Presumably this is the same issue as https://github.com/IronLanguages/ironpython3/issues/451? Maybe add a link for searchability.

Tests/test_excinfo.py
```diff
@@ -538,10 +465,10 @@ def __exit__(self, t, v, tb):
...
-                self.A(34)
+            self.A(15) # TODO: is this a bug?
+
 
+#run_test(__name__)
 
-run_test(__name__)
\ No newline at end of file
+if __name__ == "__main__":
+    unittest.main()
```
slozier: Do these changes make the test pass? If so we should remove the ignore for `test_excinfo` in `IronPythonCasesManifest.ini`. Also, did you try it with CPython?

gfmcknight: Yes -- I ran it with Python 3.6 and that's how I knew what needed to change inside the test vs what needed to change in IronPython. The changes also get test_excinfo passing.

Src/IronPython/Runtime/Operations/PythonOps.cs
gfmcknight: Nope! Will remove.

Src/IronPython/Runtime/Operations/PythonOps.cs
gfmcknight: Does not need to mutable. In order to make it internal `SaveCurrentException` and `RestoreCurrentException` will need to be internal as well. That's probably what they should be anyways.

slozier: Ah right. I think those need to be public to work with compiled code.

Tests/test_excinfo.py
slozier: I'm ok with using unittest directly. Makes it easier to run against CPython. Not really sure what we gain by using `run_test`? If it's a problem then it's easy to change it back.

gfmcknight:
> While it does seem to work, I'm not completely sold on the ExceptionMetadata. Maybe I'm just having trouble wrapping my head around the approach that you used. Particularly the bit with a callback. I'll keep staring at it maybe it'll start to make sense to me.

I'm kinda on the fence about it, too. The problem I was trying to solve is that the exception "coming from outside" the generator could change at any time at `yield` statements, and sometimes that exception was saved inside of `MSAst.Parameter`s. I was having a little trouble seeing how the generator could go in and change this exception. So I made the `ExceptionMetadata` (that really needs a better name) in order to ask the generator whether the "exception from outside" has changed every time we need to get the actual exception.

Looking at it again though, I think it's over-engineered. Since the `ExceptionMetadata` is already a sort of box for the exception, the generator class could just change the contents of that box and I think the change would be reflected in the compiled code (of course I have to see if this actually works). I feel like this solution will also have a bit of a magical code feel to it, though.

slozier:
I took a look at the CPython source (see https://github.com/python/cpython/blob/master/Objects/genobject.c). If I understand correctly they just have an exception state for the generator which they swap with the exception state of the thread before running the generator code?

Their code looks like:
```C
gen->gi_exc_state.previous_item = tstate->exc_info;
tstate->exc_info = &gen->gi_exc_state;
result = PyEval_EvalFrameEx(f, exc);
tstate->exc_info = gen->gi_exc_state.previous_item;
gen->gi_exc_state.previous_item = NULL;
```
which would translate to something like:
```C#
_state.PrevException = PythonOps.CurrentException;
PythonOps.CurrentException = _state;
try {
    return MoveNextWorker();
} finally {
    PythonOps.CurrentException = _state.PrevException;
    _state.PrevException = null;
}
```

In my example I replaced `PythonOps.CurrentException` to be of type `ExceptionState`:
```C#
internal class ExceptionState {
    public Exception Exception { get; set; }
    public ExceptionState PrevException { get; set; }
}
```

I tweaked your PR to use this approach and ran it against test_excinfo which seems to work. If you want I can load the changes I did in git.

Seems similar to your `ExceptionMetadata` approach except without the callback? Maybe a bit simpler to understand.

gfmcknight:
I think that seeing the changes would be helpful -- I'm having a little trouble understanding how it gets a try/except to restore to the correct exception inside of a generator.

slozier:
Here you go: https://github.com/IronLanguages/ironpython3/commit/39dbf3b8070b9f5712285d4fb9068e1cf582a444

gfmcknight:
I like that approach... I think it does a good job of separating the exceptions that will change while the generator is running from the exceptions that will change while the generator is not. When I revive my laptop, I'll cherry-pick that commit.

Src/IronPython/Compiler/Ast/TryStatement.cs
```diff
@@ -160,24 +166,35 @@ public class TryStatement : Statement {
...
+                    Ast.Call(AstMethods.SetCurrentException, Parent.LocalContext, exception),
+                    Ast.Rethrow()
+                );
             }
-            
+
             return Ast.Block(
-                GetVariables(lineUpdated, runElse), 
+                GetVariables(lineUpdated, runElse, previousExceptionContext),
+                Ast.Assign(previousExceptionContext, Ast.Call(AstMethods.SaveCurrentException)),
```
slozier: Something seems off here, not quite sure what exactly. It looks like you're doing the assignment twice (once in `else if (@catch != null)` and another time here)?

Src/IronPython/Compiler/Ast/TryStatement.cs
```diff
@@ -396,18 +414,25 @@ public class TryStatement : Statement {
...
-                        exception
-                    )
+                    Ast.Call(AstMethods.SetCurrentException, Parent.LocalContext, exception)
+                ),
+                // It's possible that the except block itself throws an exception,
+                // so we will need to update the current exception in that case
+                AstUtils.Try(
+                    body
+                ).Catch(exception,
+                        Ast.Call(AstMethods.SetCurrentException, Parent.LocalContext, exception),
```
slozier: nit: extra indent?

Src/IronPython/Compiler/Ast/TryStatement.cs
```diff
@@ -75,6 +76,7 @@ public class TryStatement : Statement {
             // locals allocated during the body or except blocks.
             MSAst.ParameterExpression lineUpdated = null;
             MSAst.ParameterExpression runElse = null;
+            MSAst.ParameterExpression previousExceptionContext = Ast.Variable(typeof(Exception), "$previousException");
```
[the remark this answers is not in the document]
gfmcknight: The way this is currently implemented, the value of `PythonOps.CurrentExceptionState` is not guaranteed to be restored correctly after the body is run. The reason for this is that the finally block needs to have an updated current exception, but it's also possible to jump out of the finally block and skip restoring the exception (which was what was happening in `yield from` expressions).

Also, if the handlers throw exceptions themselves, they also dirty the exception state, but I think that can (and should) be fixed if the try/catch in the finally sets the exception context instead (I'm really unsure why I didn't do that in the first place -- it's already made to look at the exception in the first place).

gfmcknight:
On the line number front, the failing tests are `test_catch_miss_finally_reraise`, `test_everything_reraise`, and `test_finally_reraise`. I think this will just be a matter of not updating the line number inside of the finally block if it's the one that was already being thrown.

Src/IronPython/Compiler/Ast/TryStatement.cs
gfmcknight: Looks like I forgot to remove one of the `SaveCurrentException`s when I moved the line down here (it was initially called in both parts of the if/else statement).

Src/IronPython/Compiler/Ast/TryStatement.cs
```diff
@@ -160,24 +165,35 @@ public class TryStatement : Statement {
                         )
                     );
             } else {
-                result = body;
+                // If the body throws an exception and there's no catch block,
+                // we should still update the current exception
+                result = AstUtils.Try(
+                    body
+                ).Catch(exception,
+                    Ast.Call(AstMethods.SetCurrentException, Parent.LocalContext, exception),
```
gfmcknight: Maybe I shouldn't be doing this here, and should instead do this over in the try/catch inside of `AddFinally()`

Src/IronPython/Compiler/Ast/TryStatement.cs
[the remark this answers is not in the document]
gfmcknight: That did not answer your question at all. I changed it so that the variable is only created when there are handlers.

gfmcknight:
On the side of implementation, I don't have anything else to do.

It looks like there were 2 failures on CI -- one was `test_math` on Mac and one was `test_missing` on Linux. On Linux the failure I'm not entirely sure how to interpret yet:
```SystemError: Could not instantiate Microsoft.Scripting.Interpreter.FuncCallInstruction`2[T0,TRet] due to generic type cannot be instantiated with type 'System.IO.Enumeration.FileSystemEntry' assembly:<unknown assembly> type:<unknown type> member:(null)```

I'm hoping that the failure on Mac is the reason that you have retry count = 2?
```
2019-07-31T05:45:59.0164780Z ======================================================================
2019-07-31T05:45:59.0165070Z ERROR: testLog2Exact (__main__.MathTests)
2019-07-31T05:45:59.0166290Z ----------------------------------------------------------------------
2019-07-31T05:45:59.0166540Z Traceback (most recent call last):
2019-07-31T05:45:59.0166780Z   File "/Users/vsts/agent/2.155.1/work/1/s/Src/StdLib/Lib/test/support/__init__.py", line 557, in wrapper
2019-07-31T05:45:59.0167010Z     version_txt = platform.mac_ver()[0]
2019-07-31T05:45:59.0167230Z   File "/Users/vsts/agent/2.155.1/work/1/s/Src/StdLib/Lib/platform.py", line 630, in mac_ver
2019-07-31T05:45:59.0167430Z     info = _mac_ver_xml()
2019-07-31T05:45:59.0167630Z   File "/Users/vsts/agent/2.155.1/work/1/s/Src/StdLib/Lib/platform.py", line 607, in _mac_ver_xml
2019-07-31T05:45:59.0167850Z     pl = plistlib.load(f)
2019-07-31T05:45:59.0168050Z   File "/Users/vsts/agent/2.155.1/work/1/s/Src/StdLib/Lib/plistlib.py", line 1011, in load
2019-07-31T05:45:59.0168270Z     return p.parse(fp)
2019-07-31T05:45:59.0168470Z   File "/Users/vsts/agent/2.155.1/work/1/s/Src/StdLib/Lib/plistlib.py", line 325, in parse
2019-07-31T05:45:59.0168680Z     self.parser.ParseFile(fileobj)
2019-07-31T05:45:59.0168890Z   File "/Users/vsts/agent/2.155.1/work/1/s/Src/StdLib/Lib/plistlib.py", line 332, in handle_begin_element
2019-07-31T05:45:59.0169100Z     handler(attrs)
2019-07-31T05:45:59.0169310Z   File "/Users/vsts/agent/2.155.1/work/1/s/Src/StdLib/Lib/plistlib.py", line 366, in begin_dict
2019-07-31T05:45:59.0169510Z     d = self._dict_type()
2019-07-31T05:45:59.0169720Z TypeError: NoneType is not callable
```

gfmcknight:
I haven't looked into either failure very closely yet.

slozier:
Yes, the retry count is there for that failing Mac test in `test_math`. Don't have a Mac to debug... I don't remember seeing that `test_missing` failure before. Probably unrelated.

slozier:
Thanks for all the work on this one (and for being patient with all my change requests)!

As a side note, I did a bit of digging on the test failures:
- `test_missing` is likely because `FlieSystemEntry` is a `ref struct` so trying to use it with a generic will indeed fail. It's probably possible to check the `Type.IsByRefLike` property and skip those types?
- `test_operator` seems like a bug in Mono (https://github.com/mono/mono/issues/16010)
- `test_math` is intermittent and as mentioned above I'm not equipped to debug it.

PR #632, opened 2019-08-05
fix test_missing on Mono

slozier:


Tests/test_missing.py
```diff
@@ -30,6 +28,12 @@
     import System.Diagnostics as Diagnostics
     from System.Reflection.Emit import OpCodes
 
+    try:
+        System.Type.IsByRefLike
```
gfmcknight: What determines when `IsByRefLike` exists? Is it just when the C# version is 7.2 or higher?

slozier: It's not tied to the C# version but to the .NET implementation. It was first added to .NET Core 2.1, my guess is to support Span and friends? It was then added to .NET Standard 2.1 which Mono is in the process of adopting.

PR #636, opened 2019-08-12
Allow indexing of memoryviews with a tuple.

gfmcknight:


Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -543,6 +552,69 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
             }
         }
 
+        public object this[[NotNull]PythonTuple index] {
+            get {
+                CheckBuffer();
+
+                int[] tupleValues = FixTuple(index);
```
slozier: From a perf point of view would be nice to avoid the array allocation on every item get/set. Maybe `FixTuple` could be changed to `GetFlatIndex` or something like that?

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -543,6 +552,69 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
...
+
+                int actualIndex = 0;
+                for (int i = 0; i < index.Count; i++) {
+                    actualIndex *= (int)shape[i];
```
slozier: Tuple access has some overhead to it, maybe cache the value `var item = (int)shape[i];` instead of calling it twice? If you wanted to go further (probably not worth it), since I think we fully control shape we could directly access the underlying array.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -543,6 +552,69 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
...
+            }
+        }
+
+        private int[] FixTuple(PythonTuple tuple) {
+            int[] values = new int[tuple.Count];
+
+            bool sawSlice = false;
+            bool sawInt = false;
+            for (int i = 0; i < tuple.Count; i++) {
+                if (tuple[i] is Slice) {
```
slozier: Same as above, consider `var item = tuple[i];`

gfmcknight:
Do you think that trying to do it all in one pass overcomplicates the logic?

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -543,6 +552,95 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
...
+                    // to check the rest of the tuple in case it has a
+                    // non-int value
+                    if (i >= ndim || firstOutOfRangeIndex > -1) {
+                        continue;
+                    }
+
+                    int dimensionWidth = (int)shape[i];
+
+                    // If we have an out of range exception, that will only
+                    // be thrown if the dimensions are correct, so we have to
```
gfmcknight: Tuple length, not dimension

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -543,6 +552,95 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
...
+                    if (i >= ndim || firstOutOfRangeIndex > -1) {
+                        continue;
+                    }
+
+                    int dimensionWidth = (int)shape[i];
+
+                    // If we have an out of range exception, that will only
+                    // be thrown if the tuple length is correct, so we have to
+                    // defer throwing to later
+                    if (indexValue < -dimensionWidth || indexValue >= dimensionWidth) {
```
slozier: I wonder if we should add an internal `PythonOps.TryFixIndex`?
```C#
if (!PythonOps.TryFixIndex(indexValue, dimensionWidth, out indexValue)) {
    firstOutOfRangeIndex = i;
    continue;
}

flatIndex *= dimensionWidth;
flatIndex += indexValue;
```

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -543,6 +552,95 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable {
...
+            //    Because we do a range check while we go through the tuple,
+            //    we have to remember that we had something out of range
+            // 2) CPython checks for a multislice tuple, then for all ints,
+            //    and throws an invalid slice key otherwise. We again try to
+            //    do this in one pass, so we remember whether we've seen an int
+            //    and whether we've seen a slice. If we haven't seen both, we
+            //    it means it's entirely ints or entirely slices
+            for (int i = 0; i < tupleLength; i++) {
+                object indexObject = tuple[i];
+                if (Converter.TryConvertToInt32(indexObject, out int indexValue)) {
```
slozier: Not sure if it's worth it or if it makes it harder to understand, but you might be able able to replace `sawInt`/`sawTuple` with something like:
```C#
bool allInts = true;
for (int i = 0; i < tupleLength; i++) {
    object indexObject = tuple[i];
    if (allInts && Converter.TryConvertToInt32(...)) {
        // ...
    } else if (indexObject is Slice) {
        allInts = false;
    } else {
        throw PythonOps.TypeError("memoryview: invalid slice key");
    }
}

if (!allInts) {
    throw PythonOps.NotImplementedError("multi-dimensional slicing is not implemented");
}
```

gfmcknight:
I think I like the changes as you suggested. Got a seat on the train so I will make them now

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: The only thing here is if we have a mix of ints and slices we need to throw the invalid slice key, and they wrote tests to explicitly check that

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1303,6 +1303,20 @@ public static partial class PythonOps {
             return v;
         }
 
+        public static bool TryFixIndex(int v, int len, out int res) {
```
gfmcknight: Should we deduplicate the code in FixIndex with this?

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: Wait that does account for it

Src/IronPython/Runtime/Operations/PythonOps.cs
slozier: Might as well, would make it nicer. I would also make `TryFixIndex` `internal` unless we have a good reason not to. The less public API surface there is the easier the codebase is to maintain.

Src/IronPython/Runtime/MemoryView.cs
gfmcknight: I think it's the most clear with the two variables. With just `allInts` we need to catch the case in which we have an int, followed only by slices. It's possible to do it with only one variable like so:
```
else if (indexObject is Slice) {
    if (allInts && i > 0) {
        throw PythonOps.TypeError("memoryview: invalid slice key");
    }
    allInts = false;
}
```
but I don't think it provides a lot of benefit.

slozier:
Thanks for the PR!

PR #639, opened 2019-09-29
Fix https://github.com/IronLanguages/ironpython2/issues/654

gpetrou:
Port of https://github.com/IronLanguages/ironpython2/pull/657


Src/IronPython/Hosting/PythonCommandLine.cs
```diff
@@ -483,7 +484,7 @@ public class PythonCommandLine : CommandLine {
                 try {
                     return RunCommandWorker(command);
                 } catch (Exception e) {
-                    Console.WriteLine(Language.FormatException(e), Style.Error);
+                    UnhandledException(e);
```
slozier: Wrong method. The fix should go in `RunFile`.

PR #643, opened 2019-10-04
Add selected tests from stdlib codecs

BCSharp:
The IronPython test apparatus allows for inclusion/exclusion of test sets on a module basis. `test_codecs` from stdlib is a large module with over 200 tests, half of them are not passing yet, so this test module is skipped. But the other half of the tests in it provides invaluable coverage and prevents regression.

This PR adds test module `test_codecs_stdlib`, which selectively runs those tests from `test_codecs` that are currently passing. Once IronPython development progresses enough for the whole `test_codecs` to pass, `test_codecs_stdlib` can be removed.


slozier:
Thanks!

PR #646, opened 2019-10-16
Implement UTF-32 support in codecs module

BCSharp:
Some of the test from stdlib now pass. Additional tests added.

Something like this will have to be done for UTF-16 as well, but in a separate PR.

### DIFF
--- a/Src/IronPython.Modules/_codecs.cs
+++ b/Src/IronPython.Modules/_codecs.cs
@@ -406,63 +406,109 @@ namespace IronPython.Modules {
 
         #region Utf-32 Functions
 
-        public static PythonTuple utf_32_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors = "strict", bool final = false)
-            => DoDecode(context, "utf-32", Encoding.UTF32, input, errors, final).ToPythonTuple();
+        public static PythonTuple utf_32_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors = "strict", bool final = false) {
+            PythonTuple res = utf_32_ex_decode(context, input, errors, 0, final);
+            return PythonTuple.MakeTuple(res[0], res[1]);
+        }
 
         public static PythonTuple utf_32_encode(CodeContext context, string input, string errors = "strict")
-            => DoEncode(context, "utf-32", Encoding.UTF32, input, errors, true).ToPythonTuple();
+            => DoEncode(context, "utf-32", Utf32LeBomEncoding, input, errors, includePreamble: true).ToPythonTuple();
 
-        #endregion
+        public static PythonTuple utf_32_ex_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors, int byteorder = 0, bool final = false) {
 
-        public static PythonTuple utf_32_ex_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors = "strict")
-            => utf_32_ex_decode(context, input, errors, null, null);
+            object output;
+            object numBytes;
 
-        public static PythonTuple utf_32_ex_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors, object byteorder, object final) {
-            byte[] lePre = Encoding.UTF32.GetPreamble();
+            if (byteorder != 0) {
+                var res = (byteorder > 0) ?
+                    utf_32_be_decode(context, input, errors, final)
+                :
+                    utf_32_le_decode(context, input, errors, final);
+                output = res[0];
+                numBytes = res[1];
 
-            string instr = Converter.ConvertToString(input);
-            bool match = true;
-            if (instr.Length > lePre.Length) {
+            } else {
+                byteorder = Utf32DetectByteorder(input);
+                var res = (byteorder > 0) ?
+                    DoDecode(context, "utf-32-be", Utf32BeBomEncoding, input, errors, final)
+                :
+                    DoDecode(context, "utf-32-le", Utf32LeBomEncoding, input, errors, final);
+                output = res.Item1;
+                numBytes = res.Item2;
+            }
+
+            return PythonTuple.MakeTuple(output, numBytes, byteorder);
+        }
+
+        private static int Utf32DetectByteorder(IList<byte> input) {
+            byte[] lePre = BOM_UTF32_LE;
+            if (input.Count > lePre.Length) {
+                bool match = true;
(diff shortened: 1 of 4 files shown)


Src/IronPython.Modules/_codecs.cs
```diff
@@ -406,63 +406,109 @@ public static PythonTuple utf_7_encode(CodeContext context, string input, string
...
         public static PythonTuple utf_32_encode(CodeContext context, string input, string errors = "strict")
-            => DoEncode(context, "utf-32", Encoding.UTF32, input, errors, true).ToPythonTuple();
-
-        #endregion
-
-        public static PythonTuple utf_32_ex_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors = "strict")
-            => utf_32_ex_decode(context, input, errors, null, null);
+            => DoEncode(context, "utf-32", Utf32LeBomEncoding, input, errors, includePreamble: true).ToPythonTuple();
+
+        public static PythonTuple utf_32_ex_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors, int byteorder = 0, bool final = false) {
```
slozier: `errors` should be optional.

Src/IronPython.Modules/_codecs.cs
```diff
@@ -406,63 +406,109 @@ public static PythonTuple utf_7_encode(CodeContext context, string input, string
...
+
+            object output;
+            object numBytes;
+
+            if (byteorder != 0) {
```
slozier: Seems like essentially the same thing in both cases. Maybe simplifying to:
```C#
if (byteorder == 0) byteorder = Utf32DetectByteorder(input);
var res = (byteorder > 0) ? ...
```

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1719,12 +1719,18 @@ public static class StringOps {
...
+        internal static string DoDecode(CodeContext context, IList<byte> s, string errors, string encoding, Encoding e)
+            => DoDecode(context, s, errors, encoding, e, true, true, out _);
 
-        internal static string DoDecode(CodeContext context, IList<byte> s, string errors, string encoding, Encoding e, bool final, out int numBytes) {
+        internal static string DoDecode(CodeContext context, IList<byte> s, string errors, string encoding, Encoding e, bool final, out int numBytes)
+            => DoDecode(context, s, errors, encoding, e, true, final, out numBytes);
+
+        // If the beginning of input s is the same as the preamble in Encoding e, those bytes are never decoded or passed to output.
+        // However, if discountPreamble is false, the number of those bytes is included in total numBytes reported.
+        internal static string DoDecode(CodeContext context, IList<byte> s, string errors, string encoding, Encoding e, bool discountPreamble, bool final, out int numBytes) {
```
slozier: Is there a use case for having `discountPreamble`? Just looking at the usage, when `discountPreamble == true` we always end up discarding `numBytes`. Unless you have specific plans for `discountPreamble` I would just remove it and use the `false` behavior all the time.

Src/IronPython.Modules/_codecs.cs
BCSharp: Right.

Src/IronPython.Modules/_codecs.cs
BCSharp: It is similar, but not the same. When given `byteorder` is not 0, BOM (if present) is passed on to output. If `byteorder` is 0, BOM (if present) is skipped. I have added some tests in `Tests/modules/io_related/test_codecs.py` to catch it and verified the stest correctnets with running under CPython 3.4.

But I can replace calls to `utf_32_[lb]e_decode` by `DoDecode`. Apart from making the code more symmetrical (thus neater), it will make the return value of the call of the same type, hence simplifying unpacking of the tuple.

Src/IronPython/Runtime/Operations/StringOps.cs
BCSharp: I had no plans for further use of `discountPreamble`. Quite the opposite: I've made a task to to for myself to look into getting rid of it. I found it strange that while the functions I was working on require the preamble bytes to be counted, the default behaviour of `DoDecode` is to discount them. I thought it was either a bug or maybe those few cases where it is really needed can be handled locally. 

I did not look into the effect of always counting the preamble bytes, thinking this may be difficult to trace (and also because so many codecs tests are still disabled), but I am encouraged by your observation that by default we discard `numBytes` anyway. So I will do some checks on my side as well and remove `discountPreamble` in the next commit.

Src/IronPython.Modules/_codecs.cs
slozier: Ah right, missed the BOM Encoding bit. The difference is clearer with both that use `DoDecode`. 👍

Src/IronPython.Modules/_codecs.cs
```diff
@@ -406,63 +406,104 @@ public static PythonTuple utf_7_encode(CodeContext context, string input, string
...
+                    DoDecode(context, "utf-32-be", Utf32BeBomEncoding, input, errors, final)
+                :
+                    DoDecode(context, "utf-32-le", Utf32LeBomEncoding, input, errors, final);
+            }
+
+            return PythonTuple.MakeTuple(res.Item1, res.Item2, byteorder);
+        }
+
+        private static int Utf32DetectByteorder(IList<byte> input) {
+            byte[] lePre = BOM_UTF32_LE;
```
slozier: Hmm, I had assumed that this was correct since it matched what we had before, but it doesn't match CPython when `input == BOM`.
```Python
assert _codecs.utf_32_ex_decode(b"\x00\x00\xfe\xff") == ('', 4, 1)
```

I think a nicer implementation would be:
```C#
        private static int Utf32DetectByteorder(IList<byte> input) {
            if (input.StartsWith(BOM_UTF32_LE)) return -1;
            if (input.StartsWith(BOM_UTF32_BE)) return 1;
            return 0;
        }
```

BCSharp: Good catch. I will add a test for this case.

The condition '>' operator should have been '>='. But I like your succinct version better. I didin't know we had `IList<byte>.StartsWith`. BTW, it can also be applied in `StringOps.GetStartingOffset`.

I'm going to read `IListOfByteOps` to see what other goodies we have around.

Issue #504, opened 2018-12-02
Support pow of negative numbers with fractional powers


slozier:
Support pow of negative numbers with fractional powers (note that I cheat in the example below since `isclose` is only in Python 3.5)
```Python
assert cmath.isclose(pow(-1, 0.5), 1j)
assert cmath.isclose((-1)**0.5, 1j)
```

Covered by `test_pow` in `test_builtin`.

slozier:
Done in https://github.com/IronLanguages/ironpython3/pull/651.

PR #651, opened 2019-10-19
Implemented Support pow of negative numbers with fractional powers #504

hackf5:
Hopefully this is the right way to do it.

The float path is tested through test_clrnuminterop.py.


hackf5:
Makes sense. I've moved the tests into Tests/test_pow.py as you have suggested.

Issue #505, opened 2018-12-02
round of an integer should return an integer


slozier:
```Python
x = 111111111111111111111111111111
assert round(x) == x
assert round(x, 2) == x
```
See https://github.com/IronLanguages/ironpython3/blob/master/WhatsNewInPython31.md

Also `round` with no `ndigits` argument always returns an `int`:
```Python
x = 1e16
assert type(round(x)) == type(int(x)) # using type(int(x)) since we haven't unified int/long yet
```

rtzoeller:
There are a few other differences with `round` between Python 2 and 3 that I wasn't able to find tracked anywhere for IronPython, so I am adding them here. If you'd prefer separate issues, I am happy to create them.

- Incorrect message when `round` is called with `ndigits` as a float (e.g. `round(0, 0.)`)
  - Python 2 (and IronPython 3): `TypeError: 'float' object cannot be interpreted as an index`
  - Python 3: `TypeError: 'float' object cannot be interpreted as an integer`
- `round(myobject, ndigits)` should delegate to `myobject.__round__(ndigits)` for general Python objects
  - Current behavior results in `TypeError: expected float, got MyClassName`

slozier:
Also:
- Python 2: if two multiples are equally close, rounding is done away from 0
- Python 3: if two multiples are equally close, rounding is done toward the even choice

hackf5:
I'm having a go at this. So far seems fiddly, but doable. However I need to make an update to the DLR submodule to provide a new rounding method and this is causing me some level of misery as Git is telling me that I have a detached branch in the submodule. I have been working in a branch, not master, which might be the cause of the issues.

![image](https://user-images.githubusercontent.com/48881437/67167641-3ddb4380-f394-11e9-9b86-6ba0e7b6681d.png)

I've tried reading this https://stackoverflow.com/questions/18770545/why-is-my-git-submodule-head-detached-from-master but it isn't making much sense.

Does anyone have any pointers? I'm not very familiar with Git (too long in a big company).

slozier:
@hackf5 Hmm, not sure. The DLR is a separate repo and you need a separate branch inside it. I usually do all my git stuff from the command line. From the command line I would go in the ironpython3\Src\DLR folder then `git checkout -b my_branch`. I would then do a PR in the DLR repo with my changes, get it merged in, then checkout the master branch of the DLR repo (with my changes merged in) and `git add Src\DLR` to the ironpython3 repo. Not sure if any of this helps...

[...]

PR #656, opened 2019-10-21
Round of an integer should return an integer #505

hackf5:
Implemented:
  - even rounding for floats and ints, floats is partially handled by an update to the DLR submodule.
  - all of the various error messages I could find are identical to those in Python 3.
  - for types that implement __round__, round() delegates to this special member.
  - when `number` is an `int` it returns an `int` when it can. when it cannot a `BigInteger` is returned.
  - when `number` is a `BigInteger` a `BigInteger` is always returned.

The algorithm I have used for integral rounding is the one described here https://bugs.python.org/msg78141, which I believe is the algorithm that they use in CPython. There is a caveat that when `ndigits` is negative and cannot be converted to an `int`, then 0 is returned.  There is no concept of signed zero for `BigInteger`.

I've routed all of the two parameter integral rounding through a single `BigInteger` method to reduce the total number of code paths. Since there is a need to unify the int and long types and the `int` method can overflow I think the benefits of this outweigh any immediate performance concerns.

I think I've tested all of the code paths, but I wouldn't say that the numerical testing is bulletproof.


slozier:
This is what I would include in the DLR instead of adding `RoundToEven`:
```C#
        public static double Round(double value, int precision, MidpointRounding mode) {
            if (double.IsInfinity(value) || double.IsNaN(value)) {
                return value;
            }

            if (precision >= 0) {
                if (precision > 308) {
                    return value;
                }

                double num = GetPowerOf10(precision);
                return Math.Round(value * num, mode) / num;
            }

            if (precision >= -308) {
                // Note: this code path could be merged with the precision >= 0 path,
                // (by extending the cache to negative powers of 10)
                // but the results seem to be more precise if we do it this way
                double num = GetPowerOf10(-precision);
                return Math.Round(value / num, mode) * num;
            }

            // Preserve the sign of the input, including +/-0.0
            return value < 0.0 || 1.0 / value < 0.0 ? -0.0 : 0.0;
        }
```

slozier:
I'm fine with this as an initial implementation. A different approach to consider (if you're at all interested) would be adding `__round__` methods to `FloatOps`, `IntOps` and `LongOps` and then delegating to those methods based on the type.

hackf5:
I'd like to take a look at moving the rounding logic into `FloatOps`, `IntOps` and `LongOps`, that sounds like the way to go and will let me see a bit more of the codebase, but it's getting late now, so I will do this tomorrow.

Src/IronPython/Modules/Builtin.cs
```diff
@@ -1401,26 +1401,23 @@ public static object eval(CodeContext/*!*/ context, [NotNull]FunctionCode code,
...
-        public static double round(double number, int ndigits) {
-            return PythonOps.CheckMath(number, MathUtils.RoundAwayFromZero(number, ndigits));
-        }
 
-        public static double round(double number, BigInteger ndigits) {
-            int n;
-            if (ndigits.AsInt32(out n)) {
-                return round(number, n);
+        public static object round(CodeContext/*!*/ context, object number) {
+            if (!PythonOps.TryGetBoundAttr(context, number, "__round__", out var func)) {
```
slozier: Maybe consider adding fast paths for the common case. For example:
```C#
if (number is double d) return DoubleOps.__round__(d);
```
This will likely give a significant perf improvement.

Src/IronPython/Runtime/Operations/FloatOps.cs
```diff
@@ -1017,6 +1017,29 @@ public static partial class DoubleOps {
...
             }
         }
+
+        public static double __round__(double self) {
+            // as of Python 3 rounding is to the nearest even number, not away from zero
+            return MathUtils.Round(self, 0, MidpointRounding.ToEven);
+        }
+
+        public static double __round__(double self, int ndigits) {
+            return PythonOps.CheckMath(self, MathUtils.Round(self, ndigits, MidpointRounding.ToEven));
```
slozier: Interesting, is there a test covering the case where it overflows to infinity?

Src/IronPython/Runtime/Operations/FloatOps.cs
```diff
@@ -1017,6 +1017,29 @@ public static partial class DoubleOps {
...
+        public static double __round__(double self, BigInteger ndigits) {
+            int n;
+            if (ndigits.AsInt32(out n)) {
+                return __round__(self, n);
+            }
+
+            return ndigits > 0 ? self : 0.0;
+        }
+
+        public static double __round__(double self, object ndigits) {
```
slozier: Not that I necessarily care about supporting this (very) edge case, but the following is valid:
```Python
class foo(object):
  def __index__(self):
    return 0

assert float.__round__(1.2, foo()) == 1.0
```
Same comment in `IntOps` and `LongOps`.

Src/IronPython/Runtime/Operations/LongOps.cs
```diff
@@ -990,6 +990,46 @@ public static partial class BigIntegerOps {
...
+        }
+
+        public static BigInteger __round__(BigInteger number, int ndigits) {
+            return __round__(number, new BigInteger(ndigits));
+        }
+
+        public static BigInteger __round__(BigInteger number, BigInteger ndigits) {
+            // as of Python 3 rounding is to the nearest even number, not away from zero
+            if (ndigits >= 0) {
+
```
slozier: nit: blank line

hackf5:
I'm just going through the code review now, as I was doing so I found a fun one:

```
>>> float(1.7976931348623157E+308).__round__()
179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368
```

It seems that once the floating point value gets large enough (I'm not sure what the threshold is) the value gets converted to a BigInteger before rounding.

I guess that is because of situations like this in the IronPython implementation:

```
>>> float(1.797693134862315**7**E+300).__round__()
1.797693134862315**6**e+300
```

Maybe someone thought that looked wrong, so decided to bite the whale and convert to an int first.

What do you want to do about it? My vote would be to raise another issue and not pull it into this as it's already got quite big.

slozier:
@hackf5 I believe there's an issue about floats being cast to ints somewhere. Will try to find it.

slozier:
Ah I see. It's not the same thing I was thinking of. Sure, a separate issue is fine.

hackf5:
> Ah I see. It's not the same thing I was thinking of. Sure, a separate issue is fine.

will do

Src/IronPython/Modules/Builtin.cs
hackf5: Done.

Src/IronPython/Runtime/Operations/FloatOps.cs
hackf5: There is now. Obviously the previous behaviour was all wrong. Good catch!

Src/IronPython/Runtime/Operations/FloatOps.cs
hackf5: No this doesn't work. But it isn't clear to me what the best way to deal with this is.

I could explicitly look for an __index__ member and call it, but this seems like it could be a specific case of a more general phenomenon.

Is there a way within the C# code to attempt to cast a python object to an int / long using the same logic as the runtime would use?

hackf5: I saw your other post about error messages, so I tried commenting out the `object ndigits` overload, expecting that this would fix the issue, however it doesn't.

I would have expected this case to be handled by the binding mechanism. i.e. `PythonOps.CallWithContextAndThis` should understand that an object with a special `__index__` member can satisfy the int parameter and perform the conversion.

What do you think?

slozier: We can't just leave it to the binding mechanism since I don't think CPython always does the conversion, for example `int(foo())` is not valid.

Seems like having some sort of attribute (e.g. `[AllowIndex]int bla`) that the binding could pick up may be interesting, but I don't know how that might affect performance. Anyway, it's not necessarily a trivial change.

There are probably a bunch of different ways that it's done in the codebase. An option would be something like `ndigits = PythonOps.Index(ndigits)`. I believe it will return an int/bigint so you'd still need to check the type and dispatch properly...

Src/IronPython/Runtime/Operations/IntOps.cs
```diff
@@ -551,6 +551,30 @@ public static partial class Int32Ops {
...
+            if (result.AsInt32(out var ret)) {
+                return ret;
+            }
+
+            // this path can be hit when number is close to int.MaxValue and ndigits is negative,
+            // causing number to be rounded up and over int.MaxValue
+            return result;
+        }
+
+        public static object __round__(int number, int ndigits) {
```
slozier: Probably don't need this overload if all it does it cast `ndigits` to bigint.

Src/IronPython/Runtime/Operations/LongOps.cs
```diff
@@ -990,6 +990,45 @@ public static partial class BigIntegerOps {
             return Bytes.Make(res.ToArray());
         }
 
+        public static BigInteger __round__(BigInteger number) {
+            return number;
+        }
+
+        public static BigInteger __round__(BigInteger number, int ndigits) {
```
slozier: Overload is not needed.

Tests/test_builtinfunc.py
```diff
@@ -730,11 +731,208 @@ def test_long_int(self):
...
+        self.assertEqual(round(number=2.5), 2)
+        self.assertEqual(round(number=3.5), 4)
+
+        self.assertEqual(round(number=2.5, ndigits=0), 2)
+        self.assertEqual(round(number=3.5, ndigits=0), 4)
+
+        self.assertEqual(round(number=25.0, ndigits=-1), 20)
+        self.assertEqual(round(number=35.0, ndigits=-1), 40)
+
+        # try:
```
slozier: any reason this one is commented out?

Src/IronPython/Runtime/Operations/FloatOps.cs
hackf5: OK, this looks doable. Actually it explains the error message I was trying to replicate. In fact this seems to be exactly what Python 3 is doing.

Src/IronPython/Runtime/Operations/FloatOps.cs
```diff
@@ -1017,6 +1017,46 @@ public static partial class DoubleOps {
...
             }
         }
+
+        public static double __round__(double self) {
+            // as of Python 3 rounding is to the nearest even number, not away from zero
+            return __round__(self, 0);
+        }
+
+        public static double __round__(double self, int ndigits) {
+            if (double.IsNaN(self)) {
```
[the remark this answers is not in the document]
hackf5: That overload calls into this one so we don't need to replicate the checks. There should be tests for these cases.

Src/IronPython/Runtime/Operations/FloatOps.cs
```diff
@@ -1017,6 +1017,46 @@ public static partial class DoubleOps {
                     throw PythonOps.ValueError("__setformat__() argument 1 must be 'double' or 'float'");
             }
         }
+
+        public static double __round__(double self) {
```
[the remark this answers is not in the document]
hackf5: Yes, I've raised an issue for this. I will look at that after this is done.

Src/IronPython/Runtime/Operations/FloatOps.cs
```diff
@@ -1017,6 +1017,46 @@ public static partial class DoubleOps {
...
+                throw PythonOps.ValueError("cannot convert float NaN to integer");
+            }
+
+            if (double.IsNegativeInfinity(self) || double.IsPositiveInfinity(self)) {
+                throw PythonOps.OverflowError("cannot convert float infinity to integer");
+            }
+
+            var result = MathUtils.Round(self, ndigits, MidpointRounding.ToEven);
+            
+            if (double.IsNegativeInfinity(result) || double.IsPositiveInfinity(result)) {
```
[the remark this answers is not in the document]
hackf5: That makes sense, it did seem a bit verbose.

Src/IronPython/Runtime/Operations/FloatOps.cs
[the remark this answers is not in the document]
slozier: Having nan/inf in this overload is valid. `float('nan').__round__(0)`

Src/IronPython/Runtime/Operations/FloatOps.cs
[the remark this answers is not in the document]
slozier: Just for reference: https://github.com/IronLanguages/ironpython3/issues/659

Src/IronPython/Runtime/Operations/FloatOps.cs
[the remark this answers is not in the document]
hackf5: > Having nan/inf in this overload is valid. `float('nan').__round__(0)`

I don't quite follow you. Are you saying it's OK as it is but pointing out that I missed out the tests for the `ndigits` case? I've added them now.

Src/IronPython/Runtime/Operations/LongOps.cs
hackf5: agreed

Src/IronPython/Runtime/Operations/IntOps.cs
hackf5: agreed

hackf5:
There is a numerical discrepancy between CPython and IronPython when the float is of the form `X.555` and `ndigits = 2)

*CPython*
```
>>> round(1.5, 0)
2.0
>>> round(1.55, 1)
1.6
>>> round(1.555, 2)
1.55
```

*IronPython*
```
>>> round(1.5, 0)
2.0
>>> round(1.55, 1)
1.6000000000000001
>>> round (1.555, 2)
1.5600000000000001
>>>
```

IronPython looks right and I doubt there is any way of reconciling that.

What do you think?

Tests/test_builtinfunc.py
hackf5: no, i've uncommented it.

Src/IronPython/Runtime/Operations/FloatOps.cs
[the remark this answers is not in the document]
slozier: The `double.IsNaN` and `double.IsInfinity` checks are only valid in the case where `ndigits` is undefined since that overload is expected to return an `int` (as you noted in https://github.com/IronLanguages/ironpython3/issues/659). As soon as you specify the `ndigits` argument it returns a `double` so `nan` and `inf` become valid values.
```Python
try:
    float('nan').__round__()
    assert False
except ValueError:
    pass

assert math.isnan(float('nan').__round__(0)) # this one shouldn't throw
```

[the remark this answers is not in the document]
hackf5: OK, got you.

slozier:
Hmm, round(1.555, 2) is surprising since the representation of 1.555 is 1.5549999999999999378275106209912337362766265869140625 it should be rounding to 1.55...

slozier:
Ah, I see, rounding will multiply by 100 which changes the value to 155.5 which is exact then it rounds to even. I don't know how precise Math.Round(value, precision, mode) is, but we might be able to call it directly when precision is <= 15?

hackf5:
> Ah, I see, rounding will multiply by 100 which changes the value to 155.5 which is exact then it rounds to even. I don't know how precise Math.Round(value, precision, mode) is, but we might be able to call it directly when precision is <= 15?

I've just tried that in C# interactive
```
> var x = Math.Round(1.555, 2, MidpointRounding.ToEven);
> x
1.56
```

So it does exactly what IronPython does.

How did you calculate that exact representation by the way?

slozier:
Well that's unfortunate. Maybe I could file an issue on corefx to get it fixed in .NET Core (but to support .NET Framework we'd have to roll our own ☹️).

`decimal.Decimal(1.555)`

slozier:
Probably this bug: https://github.com/dotnet/corefx/issues/36982

hackf5:
> Probably this bug: [dotnet/corefx#36982](https://github.com/dotnet/corefx/issues/36982)

Yes, that looks like the fellow.

> generally incorrect result most of the time

That fills one with confidence.

(I think the example above is a bit better than theirs 😉)

hackf5:
> Thanks for being patient with all my comments! If you're happy with this I'll merge it in.

You're the boss! I'm learning a lot and new to the codebase, so all of the feedback is very useful.

PR #657, opened 2019-10-21
Modify tests in stdlib to account for UTF-16 strings in .NET

BCSharp:
IronPython's strings are in the UTF-16 form thus decoding a surrogate pair always results in two characters. Consequently, decoding a broken surrogate pair with a `replace` error handler results in **two** U+FFFD rather than one as it is in case of CPython.


slozier:
Thanks!

Issue #659, opened 2019-10-22
The special method float.__round(self, ndigits) returns an int


hackf5:
### Description
`float` defines two overloads of `__round__`: `__round__(self)` and `__round__(self, ndigits)`

Where `ndigits` determines the decimal places to round to.

The overload `__round__(self)` always returns an `int`.
The overload `__round__(self, ndigits)` always returns a `float`

Somewhat confusingly this means that `__round__(self) != __round__(self, 0)` when self is moderately large.

#### In Python 3.4
```
>>> type(round(2.3, 0))
<class 'float'>
>>> type(round(2.0))
<class 'int'>
```

#### In IronPython
```
>>> type(round(2.3, 0))
<class 'float'>
>>> type(round(2.0))
<class 'float'>
```

### Versions

Found in commit https://github.com/IronLanguages/ironpython3/commit/11ecefaac7f2cd17dad7aac9102843ee5c225a91

### Fix Notes
Something like the following in `FloatOps`

```
public static object __round__(double self) {
   var result = new BigInteger(__round__(self, 0));
   return result.AsInt32(out var intResult) ? intResult : result;
}
```

But this would need to be tested against the CPython.

PR #660, opened 2019-10-23
The special method float.__round(self, ndigits) returns an int #659

hackf5:


hackf5:
Ah, thanks, I'll remember that for next time.

Issue #649, opened 2019-10-18
_codecs.utf_7_decode does work properly with incremental decoding


slozier:
The following fails:
```Python
import _codecs
b = b"+IKw-"
assert _codecs.utf_7_decode(b) == ('\u20ac', 5)
for i in range(len(b)-1):
    assert _codecs.utf_7_decode(b[:i]) == ('', 0) # these all fail with IronPython
```

There appears to be two issues:
- When it fails to decode, the number of decoded bytes is not 0
- It decodes `b"+IKw"` even if `final = False`. To support incremental decoding I assume it would need to at least check that the next byte is not `-` (or have `final = True`).

@BCSharp Any ideas? Is it an issue with the .NET UTF-7 decoder?

Encountered this issue when trying to figure out why `test_fileinput` started failing in ipy2 when I tried to backport https://github.com/IronLanguages/ironpython3/pull/430.

BCSharp:
I know that .NET UTF-7 encoder is does not behave exactly like CPython encoder, but so far have not noticed any discrepancies on the decoder side. This particular issue also does not seem to be caused by .NET.

The `_codecs` module, however, is unfinished and buggy, so I suspect there may be the cause of the problem. I haven't looked at UTF-7 support on `_codecs` at all yet. My rough plan was to get finished: UTF-32, UTF-16, UTF-8, UTF-7, and then all other special codecs in `_codecs`, in this order.

My assumption for this plan was that on .NET, UTF-16 and UTF-8 are the most important, and since UTF-32 was barely supported, I could work on it without concern of breaking anything and and the same time the implementation is easily applicable to UTF-16. But if you see other priorities (like some important tests being blocked) I am happy to refocus.

slozier:
Alright, UTF-7 is low priority so I wouldn't worry about it. I was able to avoid the issue by fixing `read1`.

BCSharp:
Alright, GTN. Anyway, it piqued my curiosity so I started to look into it, just to understand what is happening, even if we postpone the fix.

BCSharp:
Here is what I've found so far. `_codecs.utf_7_decode` underneath uses `StringOps.DoDecode` with `bool final = false` (correctly). 

[...]

PR #666, opened 2019-10-28
Incremental decoding for the UTF family of encodings

BCSharp:
This is a change in the incremental decoding approach. Instead of relying on exception to detect an incomplete sequence at the end of the increment, it now inspects the increment data to detect anomalies. The old approach was not working correctly:

* True invalid data was being detected as incomplete data and pushed to the next increment; CPython reports invalid data in the same increment in which it occurs (if possible).
* Some incomplete but otherwise valid data (completed in the next increment) was being reported as invalid, when the decoder divided unknown bytes into several calls to the fallback. For instance, uneven UTF-16 surrogate pair sequence is being reported as 2 + 1 bytes; unpaired UTF-8 surrogate is being reported as 2 + 1 bytes (and starting from .NET Core 3.0 as 1 + 1 + 1 bytes). 
* In case of UTF-7, some incomplete data was not being detected at all, with the decoder caching data without invoking any fallback (Issue #649)
* It only worked with the 'strict' error handler but all error handlers need to be supported (including user-registered ones).

The problem reported in #649 is resolved now and any workaround introduced in #430 can be removed.

Note that incremental decoding still does not work correctly with stateful error handlers (that is, `surrogateescape`). This is for another PR.


Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -2604,67 +2597,31 @@ private class UnicodeEscapeEncoding : Encoding {
...
-            }
 
             public override bool Fallback(byte[] bytesUnknown, int index) {
-                if (index > 0 && index + bytesUnknown.Length != length) {
-                    throw PythonOps.UnicodeDecodeError($"failed to decode bytes at index: {index}", bytesUnknown, index);
-                }
-                // just some bad bytes at the end
-                badBytes = bytesUnknown;
-                return false;
+                throw PythonOps.UnicodeDecodeError($"invaild bytes at index: {index}", bytesUnknown, index);
```
slozier: typo: invalid

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -2604,67 +2597,31 @@ private class UnicodeEscapeEncoding : Encoding {
...
-
-                public override bool Fallback(byte[] bytesUnknown, int index) {
-                    if (index < 0) throw new Exception();
-                    return false;
-                }
-
-                public override char GetNextChar() => (char)0;
+        // This is an equivalent of System.Text.DecoderExceptionFallback
+        // except for the custom error message and a workaround for a UTF-8 bug.
+        // It is **not** suitable for use if the following Encoding methods are being employed:
```
slozier: typo: use in

Src/IronPython.Modules/_codecs.cs
```diff
@@ -387,24 +417,68 @@ public static PythonTuple utf_16_be_encode(CodeContext context, string input, st
...
+        public static PythonTuple utf_8_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors = "strict", bool final = false)
+            => DoDecode(context, "utf-8", Utf8Encoding, input, errors, NumEligibleUtf8Bytes(input, final)).ToPythonTuple();
+
+        public static PythonTuple utf_8_encode(CodeContext context, string input, string errors = "strict")
+            => DoEncode(context, "utf-8", Encoding.UTF8, input, errors).ToPythonTuple();
+
+        private static int NumEligibleUtf8Bytes(IList<byte> input, bool final) {
+            int numBytes = input.Count;
+            if (!final) {
+                // scan for incomplete but vaild sequence at the end
```
slozier: typo: valid

Src/IronPython.Modules/_codecs.cs
```diff
@@ -387,24 +417,68 @@ public static PythonTuple utf_16_be_encode(CodeContext context, string input, st
...
+                for (int i = 1; i < 4; i++) { // 4 is the max length of a vaild sequence
```
slozier: typo: valid

Src/IronPython.Modules/_codecs.cs
```diff
@@ -387,24 +417,68 @@ public static PythonTuple utf_16_be_encode(CodeContext context, string input, st
...
 
         public static PythonTuple utf_7_decode(CodeContext context, [BytesConversion]IList<byte> input, string errors = "strict", bool final = false)
-            => DoDecode(context, "utf-7", Encoding.UTF7, input, errors, final).ToPythonTuple();
+            => DoDecode(context, "utf-7", Encoding.UTF7, input, errors, NumEligibleUtf7Bytes(input, final)).ToPythonTuple();
 
         public static PythonTuple utf_7_encode(CodeContext context, string input, string errors = "strict")
             => DoEncode(context, "utf-7", Encoding.UTF7, input, errors).ToPythonTuple();
 
+        private static int NumEligibleUtf7Bytes(IList<byte> input, bool final) {
+            byte[] bytes = input as byte[] ?? ((input is Bytes bt) ? bt.GetUnsafeByteArray() : input.ToArray());
```
slozier: Do we need to convert to an array?

Src/IronPython.Modules/_codecs.cs
```diff
@@ -387,24 +417,68 @@ public static PythonTuple utf_16_be_encode(CodeContext context, string input, st
...
+                    byte b = input[pos];
+                    if ((b & 0b10000000) == 0) return numBytes; // ASCII
+                    if ((b & 0b11000000) == 0b11000000) { // start byte
+                        if ((b | 0b00011111) == 0b11011111 && i < 2) return pos; // 2-byte seq start
+                        if ((b | 0b00001111) == 0b11101111 && i < 3) return pos; // 3-byte seq start
+                        if ((b | 0b00000111) == 0b11110111) { // 4-byte seq start
+                            if (b < 0b11110100) return pos; // chars up to U+FFFFF
+                            if ((b == 0b11110100) && (i == 1 || input[numBytes - i + 1] < 0x90)) return pos; // U+100000 to U+10FFFF
+                        }
+                        return numBytes; // invalid sequence or vaild but complete
```
slozier: typo: valid

Src/IronPython/Runtime/Operations/StringOps.cs
BCSharp: OK

Src/IronPython.Modules/_codecs.cs
BCSharp: OK

Src/IronPython.Modules/_codecs.cs
BCSharp: OK

Src/IronPython.Modules/_codecs.cs
BCSharp: OK

Src/IronPython/Runtime/Operations/StringOps.cs
BCSharp: sorry, I fail to see the typo...

Src/IronPython.Modules/_codecs.cs
BCSharp: No. It was meant as performance optimization. Since the whole array has to be scanned, I thought converting to byte[] will be beneficial. But some tests I did failed to show a benefit so I will revert to enumerating IList.

BCSharp:
I am also not quite happy with this -1 business. It was meant to be an optimization: since input is going to be converted to `byte[]` it will be cheap to get `Count`. However, although `ICollection.Count` is expensive, but `IList.Count` should be OK.

Even then, the constant defined for -1 (`AllBytes`) is in a wrong place. It is defined in `_codecs` but the meaning is established in `StringOps.DoDecode`. I'm going to get rid of -1.

Src/IronPython/Runtime/Operations/StringOps.cs
slozier: You're right, not sure what I was thinking.

PR #679, opened 2019-11-15
Normalize keys in StringOps.CodecsInfo

BCSharp:


slozier:
Guess I approved too quickly! Interesting how this seemingly small change leads to a test failure. Looks like there are multiple .NET encodings that have `WindowsCodePage == 1252` and one of them is clobbering the one we actually want (cp1252?)

Also, it looks like `windows-1252` didn't get normalized in the `#if NETCOREAPP` section which leads to a `Debug.Assert` failure.

slozier:
A possible fix might be to only publish with the `windows_xxxx` alias if `encInfo.CodePage == encInfo.GetEncoding().WindowsCodePage`?

BCSharp:
Guess I committed too quickly! I thought I had done all tests locally but now I realize it was only a partial test. The motivation for the change was to get "unicode-1-1-utf-7" to be found in `CodecsInfo`. "unicode-1-1-utf-7" was working before but doing the round trip through a Python object and stdlib. The "windows-XXX" entries got exposed when I added the debug assert. 

I think we can safely omit this particular publishing rule altogether:

* windows-1250 to windows-1258 and windows-874 will get published anyway because of their name.
* CPython does not recognize other windows-XXX names anyway.
* IronPython currently does not recognize such names either, and I would not encourage it. Users better stick with codepages (cpXXX), if needed.

slozier:
Since it's always been broken I'm fine with removing it from the `Codecs` dict!

slozier:
Thanks!

PR #690, opened 2019-12-01
Fix codecs.escape_decode

BCSharp:
And here [we have it](https://github.com/IronLanguages/ironpython3/pull/676#discussion_r346542744): `LiterslParser.ParseBytes` with `normalizeLineEndings: false`.

This is an RFC at this moment because of a regression: the old code supported `replace` error handler, which is temporarily gone. In the new code, error handlers will be implemented differently and I'd rather have this submission reviewed and agreed on before moving forward.

### DIFF
--- a/Src/IronPython.Modules/_codecs.cs
+++ b/Src/IronPython.Modules/_codecs.cs
@@ -175,11 +175,27 @@ namespace IronPython.Modules {
             => escape_decode(StringOps.DoEncodeUtf8(context, data), errors);
 
         public static PythonTuple escape_decode([BytesConversion]IList<byte> data, string errors = "strict") {
-            var res1 = LiteralParser.ParseBytes(data, 0, data.Count, isRaw: false, normalizeLineEndings: false);
-            return PythonTuple.MakeTuple(Bytes.Make(res1.ToArray()), data.Count);
-        }
+            var res = LiteralParser.ParseBytes(data, 0, data.Count, isRaw: false, normalizeLineEndings: false, errorHandler);
 
-        public static PythonTuple escape_encode([BytesConversion]IList<byte> text, string errors = "strict") {
+            return PythonTuple.MakeTuple(Bytes.Make(res.ToArray()), data.Count);
+
+            IReadOnlyList<byte> errorHandler(IList<byte> data, int start, int end) {
+                switch (errors) {
+                    case null:
+                    case "strict":
+                        throw PythonOps.ValueError("invalid \\x escape at position {0}", start);
+                    case "replace":
+                        return _replacementMarker ??= new[] { (byte)'?' };
+                    case "ignore":
+                        return null; ;
+                    default:
+                        throw PythonOps.ValueError("decoding error; unknown error handling code: " + errors);
+                }
+            }
+        }
+        private static byte[] _replacementMarker;
+
+        public static PythonTuple/*!*/ escape_encode([BytesConversion]IList<byte> text, string errors = "strict") {
             StringBuilder res = new StringBuilder();
             for (int i = 0; i < text.Count; i++) {
                 switch (text[i]) {
--- a/Src/IronPython.Modules/_codecs.cs
+++ b/Src/IronPython.Modules/_codecs.cs
@@ -175,22 +175,21 @@ namespace IronPython.Modules {
             => escape_decode(StringOps.DoEncodeUtf8(context, data), errors);
 
         public static PythonTuple escape_decode([BytesConversion]IList<byte> data, string errors = "strict") {
-            var res = LiteralParser.ParseBytes(data, 0, data.Count, isRaw: false, normalizeLineEndings: false, errorHandler);
+            Func<int, IReadOnlyList<byte>> eh = null;
+            var res = LiteralParser.ParseBytes(data, 0, data.Count, isRaw: false, normalizeLineEndings: false,
+                errorHandler: (errors != null ? errorHandler : (LiteralParser.ParseBytesErrorHandler<byte>)null));
 
             return PythonTuple.MakeTuple(Bytes.Make(res.ToArray()), data.Count);
 
             IReadOnlyList<byte> errorHandler(IList<byte> data, int start, int end) {
-                switch (errors) {
-                    case null:
-                    case "strict":
-                        throw PythonOps.ValueError("invalid \\x escape at position {0}", start);
-                    case "replace":
-                        return _replacementMarker ??= new[] { (byte)'?' };
-                    case "ignore":
-                        return null; ;
-                    default:
-                        throw PythonOps.ValueError("decoding error; unknown error handling code: " + errors);
-                }
+                eh ??= errors switch
+                {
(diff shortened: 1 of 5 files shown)


Src/IronPython.Modules/_codecs.cs
```diff
@@ -172,69 +172,14 @@ public static PythonTuple ascii_encode(CodeContext context, string input, string
...
-            }
-
-            val = 0;
-            return false;
+            var res1 = LiteralParser.ParseBytes(data, 0, data.Count, isRaw: false, normalizeLineEndings: false);
+            return PythonTuple.MakeTuple(Bytes.Make(res1.ToArray()), data.Count);
         }
 
-        public static PythonTuple/*!*/ escape_encode([BytesConversion]IList<byte> text, string errors = "strict") {
+        public static PythonTuple escape_encode([BytesConversion]IList<byte> text, string errors = "strict") {
```
slozier: Just in case you didn't know already. The `/*!*/` comments are non-null annotations. Need to look into turning on C#8 nullable.

BCSharp: I knew they were meant to indicate non-nullness, but I thought they are obsolete. The tool that was checking them is defunct and the project developing the tool seems abandoned. I removed the marker here for the sake of consistency: only `escape_encode` return type was marked with `/*!*/`, but all of the encoding and decoding methods in `_codecs` return the same non-null tuple and are not marked. Nevertheless, since you have plans for the marker, I will add it back in the next commit. Moving forward, I will leave them alone but also not introduce any new ones.

The idea of employing non-nullable reference types sounds appealing. Is there an "Issue" for it? We also have the `[NotNull]` attribute serving a similar (?) purpose.

slozier: I'm fine with the leave it alone approach. At this point they're only useful to help with future nullable annotations.

I have not filed an issue for it. I was planning on trying to annotate a few files before opening it up (although if someone else beats me to it then great).

As for replacing the `NotNullAttribute` it might be possible, although I think the C#8 `NullableAttribute` is embedded and internal? I have not given it any thought at this point.

BCSharp:
You may notice that the signature of the delegate `ParseBytesErrorHandler` is wider than strictly necessary: `data` and `end` are not used. I had several reasons in choosing this signature:

1. It carries no less information than `DecoderFallbackException` but in a more pythonic way, like `UnicodeDecodeError`. How it is being consumed is not the business of `LiteralParser`, but belongs to `_codecs`.
2. As it is now in CPython, user-defined error handlers are not recognized by `escape_decode`. If they were, all parameters would have been needed. Again, whether the caller of `LiteralParser.ParseBytes` supports used-defined error handlers or not is none of `LiteralParser` business.
3. I intend to use the same pattern with `LiteralParser.ParseString`, for supporting error handlers in `(raw-)unicode-escape` codecs, which, being fully-fledged codecs, do support user-defined error handlers, and I would like to keep the signature the same (except for the return type). Although as I am writing this now I realize this may be not so simple: Python's user-defined decoding error handlers have the ability to reset the decoding cursor, but .NET fallback mechanism does not allow it. So it will be fun trying to find a working solution. Whatever it will be, it may propagate back to `LiteralParser`.

If `data` were an array, I would probably have used `ArraySegment`.

Src/IronPython.Modules/_codecs.cs
```diff
@@ -172,67 +172,28 @@ public static PythonTuple ascii_encode(CodeContext context, string input, string
...
-                            res.Append("\\" + (char)data[i]);
-                            break;
-                    }
-                } else {
-                    res.Append((char)data[i]);
+            var res = LiteralParser.ParseBytes(data, 0, data.Count, isRaw: false, normalizeLineEndings: false, errorHandler);
+
+            return PythonTuple.MakeTuple(Bytes.Make(res.ToArray()), data.Count);
+
+            IReadOnlyList<byte> errorHandler(IList<byte> data, int start, int end) {
```
slozier: Not sure how I feel about having to do a string switch statement to check `errors` every time this is called. I guess we can leave it for now since you'll probably end up re-working the pattern if you want to do something similar in `ParseString`...

Src/IronPython.Modules/_codecs.cs
```diff
@@ -172,67 +172,28 @@ public static PythonTuple ascii_encode(CodeContext context, string input, string
...
+                switch (errors) {
+                    case null:
+                    case "strict":
+                        throw PythonOps.ValueError("invalid \\x escape at position {0}", start);
+                    case "replace":
+                        return _replacementMarker ??= new[] { (byte)'?' };
+                    case "ignore":
+                        return null; ;
```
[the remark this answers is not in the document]
BCSharp: OK

Src/IronPython/Runtime/LiteralParser.cs
```diff
@@ -234,22 +234,24 @@ public static class LiteralParser {
...
             int val;
             while (i < l) {
-                char ch = text[i++];
+                char ch = data[i++].ToChar(null);
                 if (!isRaw && ch == '\\') {
                     if (i >= l) {
                         throw PythonOps.ValueError("Trailing \\ in string");
                     }
-                    ch = text[i++];
+                    ch = data[i++].ToChar(null); ;
```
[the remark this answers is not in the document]
BCSharp: OK

Src/IronPython.Modules/_codecs.cs
BCSharp: Thank you for the comment. Yes, `ParseString` with `unicode-decode` will be more demanding and I will probably need to modify the pattern. But since this case is simpler, I'd rather have a starting solution here that we both feel good about. In the next commit I will propose a variant that does the string switch at most only once.

BCSharp:
A side note: Visual Studio insists that the opening brace of a switch expression be on a separate line. I am relying on `.editorconfig` guiding proper formatting but I do notice that it is different than for a switch statement.

Src/IronPython.Modules/_codecs.cs
slozier: Maybe something like this would make it slightly easier to follow the logic?
```C#
LiteralParser.ParseBytesErrorHandler<byte> getErrorHandler(string errors) {
    if (errors == null) return null;
    Func<int, IReadOnlyList<byte>> eh = null;
    return errorHandler;

    IReadOnlyList<byte> errorHandler(IList<byte> data, int start, int end) {
        eh ??= errors switch
        {
            "strict" => idx => throw PythonOps.ValueError("invalid \\x escape at position {0}", idx),
            "replace" => idx => _replacementMarker ??= new[] { (byte)'?' },
            "ignore" => idx => null,
            _ => idx => throw PythonOps.ValueError("decoding error; unknown error handling code: " + errors),
        };
        return eh(start);
    }
}
```

BCSharp: Good idea. Especially that for `unicode-escape`, `getErrorHandler` will only get more complex.

PR #694, opened 2019-12-05
Implement error handling for (raw-)unicode-escape codec

BCSharp:
Similarities to `escape_decode` turned out to be smaller than I expected.

### DIFF
--- a/Src/IronPython/Runtime/LiteralParser.cs
+++ b/Src/IronPython/Runtime/LiteralParser.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
+using System.Linq;
 using System.Numerics;
 using System.Text;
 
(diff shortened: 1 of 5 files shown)


Src/IronPython/Runtime/LiteralParser.cs
```diff
@@ -212,19 +186,31 @@ public static class LiteralParser {
                 }
             }
             return buf?.ToString();
+
+            void handleError(int start, int end, string reason) {
+                bytesData ??= data is byte[] ba ? new Bytes(ba) : Bytes.Empty;
+
+                if (errorHandler == null || start >= bytesData.Count) {
```
slozier: If `data` is not `byte[]` it will ignore `errorHandler` and always throw (since `bytesData.Count == 0`). Is this the expected behavior? If we expect `errorHandler` to always be `null` with `char[]` then I would add a guard (`Debug.Assert` or `NotImplementedException`) at the top of the method.

BCSharp: Yes, it was the intended behaviour. The current implementation assumes `errorHandler` to be always `default` with `char[]`, since currently there is no need for a non-default handler in such case. If in the future this assumption is no longer valid, `handleError` has to be updated. I agree that a guard with `NotImplementedException` is better than a silent ignore.

Issue #33, opened 2014-03-21
What's New in Python 3.0


jdhardy:
Go over http://docs.python.org/3.0/whatsnew/3.0.html for anything that doesn't have a PEP and create an issue for it.

slide:
Tracking changes in https://github.com/IronLanguages/ironpython3/pull/113

slide:
Views And Iterators Instead Of Lists
=======================
- [ ] dict methods dict.keys(), dict.items() and dict.values() return “views” instead of lists. For example, this no longer works: k = d.keys(); k.sort(). Use k = sorted(d) instead (this works in Python 2.5 too and is just as efficient).
- [ ] Also, the dict.iterkeys(), dict.iteritems() and dict.itervalues() methods are no longer supported.
- [ ] map() and filter() return iterators. If you really need a list, a quick fix is e.g. list(map(...)), but a better fix is often to use a list comprehension (especially when the original code uses lambda), or rewriting the code so it doesn’t need a list at all. Particularly tricky is map() invoked for the side effects of the function; the correct transformation is to use a regular for loop (since creating a list would just be wasteful).
- [ ] zip() now returns an iterator.

Ordering Comparisons
==============
- [ ] The ordering comparison operators (<, <=, >=, >) raise a TypeError exception when the operands don’t have a meaningful natural ordering. Thus, expressions like 1 < '', 0 > None or len <= len are no longer valid, and e.g. None < None raises TypeError instead of returning False. A corollary is that sorting a heterogeneous list no longer makes sense – all the elements must be comparable to each other. Note that this does not apply to the == and != operators: objects of different incomparable types always compare unequal to each other.
- [ ] builtin.sorted() and list.sort() no longer accept the cmp argument providing a comparison function. Use the key argument instead. N.B. the key and reverse arguments are now “keyword-only”
- [ ] The cmp() function should be treated as gone, and the __cmp__() special method is no longer supported. Use __lt__() for sorting, __eq__() with __hash__(), and other rich comparisons as needed. (If you really need the cmp() functionality, you could use the expression (a > b) - (a < b) as the equivalent for cmp(a, b).)

Integers
=====
- [ ] PEP 0237: Essentially, long renamed to int. That is, there is only one built-in integral type, named int; but it behaves mostly like the old long type.
- [ ] PEP 0238: An expression like 1/2 returns a float. Use 1//2 to get the truncating behavior. (The latter syntax has existed for years, at least since Python 2.2.)
[...]

PR #695, opened 2019-12-05
Remove comparison operator fallbacks

gfmcknight:
Remove type comparison fallbacks for the '<', '>', '<=', and '>='
operators and throw a TypeError instead when no suitable comparator is
found between two types.

Listed in #33.

### DIFF
--- a/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
+++ b/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
@@ -1286,12 +1286,17 @@ namespace IronPython.Runtime.Binding {
         }
 
         private static Expression/*!*/ MakeFallbackCompare(DynamicMetaObjectBinder/*!*/ binder, PythonOperationKind op, DynamicMetaObject[] types) {
-            return Ast.Call(
-                GetComparisonFallbackMethod(op),
-                PythonContext.GetCodeContext(binder),
-                AstUtils.Convert(types[0].Expression, typeof(object)),
-                AstUtils.Convert(types[1].Expression, typeof(object))
-            );
+            if (op == PythonOperationKind.Equal || op == PythonOperationKind.NotEqual ||
+                    op == PythonOperationKind.Compare) {
+                return Ast.Call(
+                    GetComparisonFallbackMethod(op),
+                    PythonContext.GetCodeContext(binder),
+                    AstUtils.Convert(types[0].Expression, typeof(object)),
+                    AstUtils.Convert(types[1].Expression, typeof(object))
+                );
+            }
+
+            return MakeBinaryThrow(binder, op, types).Expression;
         }
 
         private static Expression GetCompareTest(PythonOperationKind op, Expression expr, bool reverse) {
@@ -1881,10 +1886,6 @@ namespace IronPython.Runtime.Binding {
             switch (op) {
                 case PythonOperationKind.Equal: name = nameof(PythonOps.CompareTypesEqual); break;
                 case PythonOperationKind.NotEqual: name = nameof(PythonOps.CompareTypesNotEqual); break;
-                case PythonOperationKind.GreaterThan: name = nameof(PythonOps.CompareTypesGreaterThan); break;
-                case PythonOperationKind.LessThan: name = nameof(PythonOps.CompareTypesLessThan); break;
-                case PythonOperationKind.GreaterThanOrEqual: name = nameof(PythonOps.CompareTypesGreaterThanOrEqual); break;
-                case PythonOperationKind.LessThanOrEqual: name = nameof(PythonOps.CompareTypesLessThanOrEqual); break;
                 case PythonOperationKind.Compare: name = nameof(PythonOps.CompareTypes); break;
                 default: throw new InvalidOperationException();
             }
@@ -1990,7 +1991,7 @@ namespace IronPython.Runtime.Binding {
                     action.Throw(
                         Ast.Call(
                             typeof(PythonOps).GetMethod(nameof(PythonOps.TypeErrorForBinaryOp)),
-                            AstUtils.Constant(Symbols.OperatorToSymbol(op)),
+                            AstUtils.Constant(GetOperatorDisplay(op)),
                             AstUtils.Convert(args[0].Expression, typeof(object)),
                             AstUtils.Convert(args[1].Expression, typeof(object))
                         ),
--- a/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
+++ b/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
@@ -1297,6 +1297,17 @@ namespace IronPython.Runtime.Binding {
             }
 
             return MakeBinaryThrow(binder, op, types).Expression;
+
+            static MethodInfo/*!*/ GetComparisonFallbackMethod(PythonOperationKind op) {
+                string name;
+                switch (op) {
+                    case PythonOperationKind.Equal: name = nameof(PythonOps.CompareTypesEqual); break;
+                    case PythonOperationKind.NotEqual: name = nameof(PythonOps.CompareTypesNotEqual); break;
+                    case PythonOperationKind.Compare: name = nameof(PythonOps.CompareTypes); break;
+                    default: throw new InvalidOperationException();
+                }
+                return typeof(PythonOps).GetMethod(name);
+            }
         }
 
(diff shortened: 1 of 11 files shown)


Tests/test_attr.py
```diff
@@ -56,7 +56,7 @@ def CheckModule(mod):
             CheckDictionary(mod.__dict__)
 
             mod.__dict__[1] = '1'
-            self.assertEqual(dir(mod).__contains__(1), True)
+            self.assertRaises(TypeError, lambda: dir(mod).__contains__(1))
```
gfmcknight: This one is interesting and kinda makes me wonder whether this is intended behavior, but in Python 3.6, I got:
```
>>> import fractions
>>> fractions.__dict__[1] = '1'
>>> dir(fractions)
TypeError: '<' not supported between instances of 'int' and 'str'
```

Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
```diff
@@ -1881,10 +1886,6 @@ public ItemBuilder(DynamicMetaObject/*!*/[]/*!*/ types, Callable/*!*/ callable)
             switch (op) {
                 case PythonOperationKind.Equal: name = nameof(PythonOps.CompareTypesEqual); break;
                 case PythonOperationKind.NotEqual: name = nameof(PythonOps.CompareTypesNotEqual); break;
-                case PythonOperationKind.GreaterThan: name = nameof(PythonOps.CompareTypesGreaterThan); break;
```
slozier: Since you're playing with this code, consider making this function to a static local function in `MakeFallbackCompare`?

Tests/modules/misc/test__weakref.py
```diff
@@ -159,8 +159,15 @@ def %s(self, *args, **kwargs):
                 self.assertEqual(eval('x ' + op + ' 3'), op == '!='); self.assertEqual(called, None)
                 self.assertEqual(eval('3 ' + op + ' x'), op == '!='); self.assertEqual(called, None)
             else:
-                res1, res2 = eval('x ' + op + ' 3'), eval('3 ' + op + ' x')
-                self.assertEqual(called, None)
-                self.assertTrue((res1 == True and res2 == False) or (res1 == False and res2 == True))
+                try:
```
slozier: Can we use this format instead?
```Python
with self.assertRaises(TypeError):
    eval('x ' + op + ' 3')
```

Also, I assume CPython also throws?

Tests/test_attr.py
slozier: Presumably `dir` does a sort so this would be the expected behavior (e.g. `["a", 1].sort()`).

Tests/modules/system_related/test_nt.py
```diff
@@ -586,43 +586,27 @@ def test_stat_result(self):
         self.assertEqual(x + x, tuple(range(10))*2)
 
         #> (list/object)
-        if is_cli:
```
slozier: I wonder if it's worth making these `assertRaises`?

Tests/test_dict.py
```diff
@@ -472,23 +472,13 @@ def test_dict_to_idict(self):
...
-        class KNewDerived(KNew): pass
 
         test_dicts = [
                         {},
                         {1:100},
-                        {None:None},
                         {object:object},
                         {1:100, 2:200},
                         {1:100, 2:200, 3:300, 4:400},
-                        MyDict.__dict__,
```
slozier: It's too bad we're losing the check since `MyDict.__dict__` is actually `mappingproxy` and not a `dict` like the remaining values. Maybe we could change `DictConversion.ToIDictionary` to return tuples instead? Then we'd be able to sort and compare.

Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
```diff
@@ -1286,12 +1286,17 @@ internal static partial class PythonProtocol {
...
+                    op == PythonOperationKind.Compare) {
+                return Ast.Call(
+                    GetComparisonFallbackMethod(op),
+                    PythonContext.GetCodeContext(binder),
+                    AstUtils.Convert(types[0].Expression, typeof(object)),
+                    AstUtils.Convert(types[1].Expression, typeof(object))
+                );
+            }
+
+            return MakeBinaryThrow(binder, op, types).Expression;
```
slozier: Any idea in what other cases `MakeBinaryThrow` is used? Just wondering if changing the error message to match CPython will break other error messages.

gfmcknight: MakeBinaryThrow is called for all binary operators I think; it looks like those also use the symbol in Python 3, so I just changed it here.

slozier: Ok, looks like the error message is slightly different (CPython 3.8) for the comparison operators:
```
>>> 1 ^ None
TypeError: unsupported operand type(s) for ^: 'int' and 'NoneType'
>>> 1 < None
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

Since the error messages basically mean the same thing I'm fine with diverging from CPython as long as there are no StdLib tests wanting to checking the actual message.

Tests/modules/misc/test__weakref.py
gfmcknight: Ah; this doesn't throw in CPython, and I think that the whole premise of the test doesn't apply anymore?
```
>>> global called
>>> class C:
...     for method, op in [('__eq__', '=='), ('__gt__', '>'), ('__lt__', '<'), ('__ge__', '>='), ('__le__', '<='), ('__ne__', '!=')]:
...         exec("""
... def %s(self, *args, **kwargs):
...     global called
...     called = '%s'
...     return True
... """ % (method, op))
...
>>> a = C()
>>> import _weakref
>>> x = _weakref.proxy(a)
>>> eval('x < 3')
True
>>> called
'<'
>>> eval('x == 3')
True
>>> called
'=='
```

slozier: Indeed. It seems like it calls all the underlying operators with Python 3? I guess the test should be inverted.

gfmcknight: Then let's change the test to look for the new behavior an add an expected failure annotation instead? This looks like a pretty major feature -- I wonder it's under any of the What's new in Python 3.x updates.

Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
gfmcknight: It doesn't seem to show as updated, but this should now be a local method.

Tests/modules/misc/test__weakref.py
slozier: Agreed, let's update the test to the new behavior. We should file an issue about it and reference it in the test if possible.

I just tried with 3.0 and it looks like the operators "flow through" the proxy. Kind of makes sense since the proxy is supposed to be a stand-in for the actual value. So if it's in any of the release notes it would be in 3.0.

Src/IronPythonTest/Conversions.cs
```diff
@@ -69,11 +70,11 @@ public class DictConversion {
         public static IList<object> ToIDictionary(IDictionary dict) {
             List<object> res = new List<object>();
             foreach (DictionaryEntry de in dict) {
-                res.Add(de.Key);
+                res.Add(PythonTuple.MakeTuple(de.Key));
```
slozier: Not quite what I had in mind. I was thinking more like `PythonTuple.MakeTuple(de.Key, de.Value)`. I'm surprised the test is passing! I guess tuple comparison needs work (e.g. `(1,) < (None,)` should throw).

gfmcknight: I guess because we use comparison in `CompareArrays()`:
https://github.com/IronLanguages/ironpython3/blob/fa2d3972aaecc83283b205b47feb1a51883becc6/Src/IronPython/Runtime/Operations/PythonOps.cs#L688-L696

slozier: Yeah, something for a separate PR I think.

Tests/modules/misc/test__weakref.py
gfmcknight: I couldn't find it in any release notes, but I believe that it was discussed in this bug: https://bugs.python.org/issue9658 and the one linked at the bottom of it as well. It looks to me like operators in general have been added incrementally to the proxy implementation.

PR #696, opened 2019-12-08
Implement codecs.strict_errors

BCSharp:
Current `codecs` module lacks a number of standalone error handling functions (or, to be exact, the exported symbols are not callable):

```
>>> import codecs
>>> [ f for f in dir(codecs) if f.endswith("errors")]
['backslashreplace_errors', 'ignore_errors', 'replace_errors', 'strict_errors', 'xmlcharrefreplace_errors']
```

This PR implements `strict_errors` as a pilot; if the approach is agreed upon, I will implement the remaining ones in a similar fashion.

I am not quite sure whether using a delegate to put into the error handlers dictionary is the right way; perhaps there is a better, more ironpythonic way of doing it. As it is now, there are some subtle differences between IronPython and CPython regarding the type:

_CPython_:

```
>>> codecs.strict_errors
<built-in function strict_errors>
>>> type(codecs.strict_errors)
<class 'builtin_function_or_method'>
>>> codecs.strict_errors()
TypeError: strict_errors() takes exactly one argument (0 given)
```

_IronPython_:

```
>>> codecs.strict_errors
<IronPython.Runtime.Operations.StringOps+CodecsInfo+ErrorHandler object at 0x000000000000008E [IronPython.Runtime.Operations.StringOps+CodecsInfo+ErrorHandler]>
>>> type(codecs.strict_errors)
<class 'ErrorHandler'>
>>> codecs.strict_errors()    
TypeError: Invoke() takes exactly 1 argument (0 given)
```

Functionality-wise, they behave the same.

### DIFF
--- a/Src/IronPython/Runtime/Operations/StringOps.cs
+++ b/Src/IronPython/Runtime/Operations/StringOps.cs
@@ -1869,8 +1869,8 @@ namespace IronPython.Runtime.Operations {
         }
 
 #if FEATURE_ENCODING
-        private static class CodecsInfo {
-            public static readonly Dictionary<string, Lazy<Encoding>> Codecs = MakeCodecsDict();
+        internal static class CodecsInfo {
+            internal static readonly Dictionary<string, Lazy<Encoding>> Codecs = MakeCodecsDict();
 
             private static Dictionary<string, Lazy<Encoding>> MakeCodecsDict() {
                 Dictionary<string, Lazy<Encoding>> d = new Dictionary<string, Lazy<Encoding>>();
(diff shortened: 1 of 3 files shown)


Src/IronPython/Runtime/PythonContext.cs
```diff
@@ -1887,7 +1887,11 @@ public string GetDefaultEncodingName()
         internal Dictionary<string, object> ErrorHandlers {
             get {
                 if (_errorHandlers == null) {
+#if FEATURE_ENCODING
+                    _errorHandlers = StringOps.CodecsInfo.ErrorHandlers.Value;
```
slozier: `CodecsInfo` is static which means that this dict would apply to all contexts in an app which is likely undesirable. I would be fine with having something like `Dictionary<string, object> CodecsInfo.CreateErrorHandlers()` and using that (as long as it returned a new dictionary on every call).

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1939,6 +1939,23 @@ private static class CodecsInfo {
 #endif
                 return d;
             }
+
+            internal static readonly Lazy<Dictionary<string, object>> ErrorHandlers = new Lazy<Dictionary<string, object>>(MakeErrorHandlersDict);
```
slozier: See my comment in `PythonContext.cs`.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1939,6 +1939,23 @@ private static class CodecsInfo {
...
+
+            private delegate object ErrorHandler(object unicodeError);
+
+            private static Dictionary<string, object> MakeErrorHandlersDict() {
+                var d = new Dictionary<string, object>();
+
+                d["strict"] = new ErrorHandler(StrictErrors);
```
slozier: You can wrap the method in `BuiltinFunction` instead of creating delegates. Something like the following:
```C#
d["strict"] = BuiltinFunction.MakeFunction("strict_errors", ReflectionUtils.GetMethodInfos(typeof(StringOps).GetMember(nameof(StrictErrors), BindingFlags.Static | BindingFlags.NonPublic)), typeof(StringOps));
```
Note that I'm super familiar with this part of the codebase (`BuiltinFunction`) so I may be misusing it (e.g. what is the proper `declaringType`).

Src/IronPython/Runtime/PythonContext.cs
BCSharp: It is undesirable indeed; I forgot about the contexts.

PR #699, opened 2019-12-10
Enable testing with CPython on Linux

BCSharp:
The value of `sys.platform` in CPython on Linux is `linux`, while IronPython reports `posix`. Many tests that are not meant to be run on Linux are guarded by `is_posix`, so I have extended the definition of it to include what CPython reports. In this way I am able to verify the tests with CPython while working on Linux.

CPython is apparently more detailed in `sys.platform` than IronPython (which only differentiates `darwin`), and adding only `'linux'` here is not exhaustive for all POSIX systems. CPython will report `posix` as `os.name` (which includes the `darwin` case) and `sys.platform` as `unknown` if not recognized.

Note that the value of `sys.platform` is printed out on REPL start, right after `on`:

<pre><code>Python 3.4.4 (v3.4.4:737efcadf5a6, Dec 20 2015, 20:20:57) [MSC v.1600 64 bit (AMD64)] <strong>on win32</strong>
>>>

Python 3.6.9 (default, Nov  7 2019, 10:44:02)
[GCC 8.3.0] <strong>on linux</strong>
>>>

Python 3.7.5 (default, Nov  1 2019, 02:16:38)
[Clang 10.0.0 (clang-1000.11.45.5)] <strong>on darwin</strong>
>>>
</code></pre>

IronPython does not do it, maybe worth considering adding?

 * * *

The environment variable `PROCESSOR_ARCHITECTURE` is Windows-specific. I have changed the bitness detection for the CPython case to be more portable, based on pointer size, like the IronPython's bitness detection test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,5 +102,7 @@ launchSettings.json
 *.o
 /Src/ctypes_test/*.pyd
 /Src/ctypes_test/*.pyd
-/Src/StdLib/Lib/__pycache__
+/Src/StdLib/Lib/**/__pycache__
+/Src/IronPython/Lib/**/__pycache__
+/Tests/**/__pycache__
 /Tests/modules/misc/test_data.gz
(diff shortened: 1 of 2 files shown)


slozier:
We should change `sys.platform` to report `linux` on Linux and get rid of any uses of `sys.platform == "posix"`. As for adding the "on" bit in the repl that would be fine as well.

PR #701, opened 2019-12-11
Test codecs linux

BCSharp:
There seems to be a race condition in `__inport__` in CPython 3.4 on Linux, that a 10 ms delay solves.

### DIFF
--- a/Tests/modules/io_related/test_codecs.py
+++ b/Tests/modules/io_related/test_codecs.py
@@ -21,7 +21,7 @@ import subprocess
 import sys
 import unittest
 
-from iptest import IronPythonTestCase, is_cli, is_cpython, is_mono, is_netcoreapp, is_posix, is_windows, is_osx, run_test, skipUnlessIronPython
+from iptest import IronPythonTestCase, is_cli, is_cpython, is_mono, is_netcoreapp, is_posix, is_windows, run_test, skipUnlessIronPython
 from iptest.misc_util import ip_supported_encodings
 
 if is_cpython and is_posix:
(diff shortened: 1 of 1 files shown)


Tests/modules/io_related/test_codecs.py
```diff
@@ -21,9 +21,12 @@
 import sys
 import unittest
 
-from iptest import IronPythonTestCase, is_cli, is_cpython, is_mono, is_netcoreapp, is_posix, run_test, skipUnlessIronPython
+from iptest import IronPythonTestCase, is_cli, is_cpython, is_mono, is_netcoreapp, is_posix, is_windows, is_osx, run_test, skipUnlessIronPython
 from iptest.misc_util import ip_supported_encodings
 
+if is_cpython and is_posix:
```
slozier: I was a bit quick when merging in https://github.com/IronLanguages/ironpython3/pull/699. `is_posix` should actually mean mac or linux. I'll hold off on merging this until https://github.com/IronLanguages/ironpython3/issues/704 is done and amend this PR to use `is_linux` instead.

BCSharp: If `is_posix` means mac or linux, then the whole mac commit  e6a53913ce0677fd726394fa1adb17f43eaa1907 can/should be reverted.

slozier: I've merged in the changes, may be easier for you to roll back the commit. Otherwise I'll try to do it later today.

BCSharp: OK, I will do it now.

BCSharp:
CI will fail because changes in 02c5376e4dc04f61a4e568536552d803aac778d1 depend on d3c3de30e6e3b926f7f59cf55814c6113d638af4 from master (#705), but I have merged to workdir with master on all three platforms to verify and it works after merging.

BCSharp:
I suppose CI tests the merged version, after all it is called CI :blush:

slozier:
Thanks for the PR!

BCSharp:
My thanks.

PR #710, opened 2019-12-17
Make weakref proxies delegate comparison operators

### LINKED ISSUE #697: Weakref proxies should delegate comparison operators
### Prerequisites

* [X] Are you running the latest version?
* [X] Are you reporting to the correct repository?
* [X] Did you perform a cursory search?

### Description
`_weakref.proxy()` should delegate comparison and other operators to the proxied object.

### Steps to Reproduce
Run the following in IronPython
```
import _weakref

global called
called = None
class C:
    for method, op in [('__eq__', '=='), ('__gt__', '>'), ('__lt__', '<'), ('__ge__', '>='), ('__le__', '<='), ('__ne__', '!=')]:
        exec("""
def %s(self, *args, **kwargs):
    global called
    called = '%s'
    return True
""" % (method, op))

a = C()
x = _weakref.proxy(a)
print(x < 3)
print(called)
```

**Expected behavior:**
The operator gets called and
```
True
<
```
is printed.

**Actual behavior:**
The operator is not called and the following is printed instead:
```
False
None
```

### Versions
IronPython 3.4.0 3.4.0.0 on .NET 4.0.30319.42000

gfmcknight:
Add slots for all comparison operators which delegate to the object
being proxied.

Resolves #697

### DIFF
--- a/Src/IronPython.Modules/_weakref.Generated.cs
+++ b/Src/IronPython.Modules/_weakref.Generated.cs
@@ -57,7 +57,12 @@ namespace IronPython.Modules {
 
             #endregion
 
-            //[SlotField] public static PythonTypeSlot __cmp__ = new SlotWrapper(Symbols.Cmp, ProxyType);
+            [SlotField] public static PythonTypeSlot __eq__ = new SlotWrapper("__eq__", ProxyType);
+            [SlotField] public static PythonTypeSlot __ne__ = new SlotWrapper("__ne__", ProxyType);
+            [SlotField] public static PythonTypeSlot __lt__ = new SlotWrapper("__lt__", ProxyType);
+            [SlotField] public static PythonTypeSlot __gt__ = new SlotWrapper("__gt__", ProxyType);
+            [SlotField] public static PythonTypeSlot __le__ = new SlotWrapper("__le__", ProxyType);
+            [SlotField] public static PythonTypeSlot __ge__ = new SlotWrapper("__ge__", ProxyType);
             [SlotField] public static PythonTypeSlot __divmod__ = new SlotWrapper("__divmod__", ProxyType);
             [SlotField] public static PythonTypeSlot __float__ = new SlotWrapper("__float__", ProxyType);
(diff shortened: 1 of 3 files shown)


Src/IronPython.Modules/_weakref.cs
```diff
@@ -450,23 +450,6 @@ public sealed partial class weakproxy : IPythonObject, ICodeFormattable, IProxyO
...
             }
 
-            /// <summary>
-            /// Special equality function because IStructuralEquatable.Equals is not allowed to throw.
-            /// </summary>
-            [return: MaybeNotImplemented]
-            public object __eq__(object other) {
-                if (!(other is weakproxy)) return NotImplementedType.Value;
-
-                return ScriptingRuntimeHelpers.BooleanToObject(EqualsWorker((weakproxy)other));
```
slozier: Is `EqualsWorker` still used somewhere? If not we should get rid of it.

Src/IronPython.Modules/_weakref.Generated.cs
```diff
@@ -57,7 +57,12 @@ public sealed partial class weakproxy {
 
             #endregion
 
-            //[SlotField] public static PythonTypeSlot __cmp__ = new SlotWrapper(Symbols.Cmp, ProxyType);
+            [SlotField] public static PythonTypeSlot __eq__ = new SlotWrapper("__eq__", ProxyType);
```
slozier: Neat, didn't know about `SlotField`.

slozier:
I finally got around to running some tests, there's a minor issue when creating a proxy of a type that doesn't have rich comparison operators. However I think it should be resolved by https://github.com/IronLanguages/ironpython3/pull/709 since it'll add the operators to the `object` type. Will hold off on merging this PR until the other one is ready. In case you're interested here's the failure I ran into:
```
>>> import _weakref
>>> class A: pass
...
>>> a = A()
>>> p = _weakref.proxy(a)
>>> p.__lt__
SystemError: Operation is not valid due to the current state of the object.
```

slozier:
Thanks for the PR! Will merge it in now even though #709 isn't ready.

PR #711, opened 2019-12-18
Implement codecs.ignore_errors

BCSharp:
It works, but I feel uneasy about using a type which name starts with an underscore, although the type itself is public. Is it the right way of doing? If not, here is an alternative implementation of `ignore_errors`, which I can push in the next commit:

```csharp
private static object IgnoreErrors(CodeContext context, object unicodeError) {
    switch (unicodeError) {
        case PythonExceptions.BaseException _:
            var ptype = DynamicHelpers.GetPythonType(unicodeError);
            if (ptype.Name == "UnicodeDecodeError" || ptype.Name == "UnicodeEncodeError") {
                if (ptype.TryGetBoundAttr(context, unicodeError, "end", out object ret) && ret is int end) {
                    return PythonTuple.MakeTuple(string.Empty, end);
                }
            }
            goto default;
        case DecoderFallbackException dfe: 
            return PythonTuple.MakeTuple(string.Empty, dfe.Index + dfe.BytesUnknown.Length);
        case EncoderFallbackException efe: 
            return PythonTuple.MakeTuple(string.Empty, efe.Index + (efe.CharUnknownHigh != '\0' ? 2 : 1));
        default: throw PythonOps.TypeError("codec must pass exception instance");
    }
}
```

### DIFF
--- a/Src/IronPython/Runtime/Operations/StringOps.cs
+++ b/Src/IronPython/Runtime/Operations/StringOps.cs
@@ -1948,9 +1948,12 @@ namespace IronPython.Runtime.Operations {
                     ReflectionUtils.GetMethodInfos(typeof(StringOps).GetMember(nameof(StrictErrors), BindingFlags.Static | BindingFlags.NonPublic)), 
                     typeof(StringOps));
 
-                // TODO: Implement remaining error handlers
-                d["ignore"] = null;
+                d["ignore"] = BuiltinFunction.MakeFunction(
+                    "ignore_errors", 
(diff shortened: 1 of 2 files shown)


BCSharp:
I haven't noticed `test_codeccallbacks.py` until now. This is great. The tests I am writing are just enough to guide the implementation; they are by no means exhaustive. `test_codeccallbacks.py` is invaluable. Right now, almost all tests from that file fail, but looking at the failures I see they are expected. So there is still some way to go.

PR #713, opened 2019-12-31
Surrogateescape errors

BCSharp:
This PR implements the `codecs.surrogateescape_errors` builtin function with the accompanying tests. The same caveat applies as for the previous error handlers: in case of misconstructed `UnicodeError` the way how the range gets coerced may sometimes differ from CPython, but in no instance will it throw `IndexError`.

During testing I discovered that my original understanding of [PEP 383](https://www.python.org/dev/peps/pep-0383/) was wrong. I thought that PEP 383 forbids smuggling of ASCII characters, but it actually forbids smuggling of ASCII bytes. Consequently, `surrogateescape` cannot be meaningfully used with wide-char encodings, like UTF-16 or UTF-32. The Python documentation doesn't mention it, but the PEP does say it is intended for ASCII-compatible encodings only. On the bright side, the implementation becomes simpler.

For instance, in case of UTF-16, if a lone surrogate (ordinarily an error) has the LSB 0x80 or above, it will be escaped; if that byte is less than 0x80, it will not be escaped but treated as belonging to the next character, resulting in all subsequent bytes misaligned:

```python
>>> b"\xdd\x80".decode('utf-16-be', 'surrogateescape')
'\udcdd\udc80'
>>> b"\xdd\x20".decode('utf-16-be', 'surrogateescape')
  File "C:\Users\Pawel\.conda\envs\py37\lib\encodings\utf_16_be.py", line 16, in decode
    return codecs.utf_16_be_decode(input, errors, True)
UnicodeDecodeError: 'utf-16-be' codec can't decode byte 0x20 in position 1: truncated data
>>> b"\xdd\x20\xac".decode('utf-16-be', 'surrogateescape')
'\udcdd€'
```
In case of UTF-32, no surrogates can be escaped as they invariably contain two zero bytes.

Since it is a security issue, in the second commit of this PR I correct `PythonSurrogateEscapeEncoding` and modify the relevant tests. The change prevents smuggling of ASCII bytes, although the behaviour is still not exactly like CPython: the third statement from the example above will raise an exception rather than misalign the subsequent bytes. I leave it like this for now because I am considering getting rid of `PythonSurrogateEscapeEncoding` altogether at some point in the future.

### DIFF
--- a/Src/IronPython/Runtime/Operations/StringOps.cs
+++ b/Src/IronPython/Runtime/Operations/StringOps.cs
@@ -1973,6 +1973,11 @@ namespace IronPython.Runtime.Operations {
                     ReflectionUtils.GetMethodInfos(typeof(StringOps).GetMember(nameof(BackslashReplaceErrors), BindingFlags.Static | BindingFlags.NonPublic)),
                     typeof(StringOps));
 
+                d["surrogateescape"] = BuiltinFunction.MakeFunction(
+                    "surrogateescape_errors",
+                    ReflectionUtils.GetMethodInfos(typeof(StringOps).GetMember(nameof(SurrogateEscapeErrors), BindingFlags.Static | BindingFlags.NonPublic)),
+                    typeof(StringOps));
+
                 return d;
             }
         }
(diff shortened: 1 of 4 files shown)


slozier:
Thanks for the PR!

PR #718, opened 2020-01-09
Making `test_winreg` passes.

### LINKED ISSUE #410: test_winreg is blocking
The following code isn't terminating:
```Python
import threading

def test_thread():
    class Thread(threading.Thread):
        def run(self):
            print("run")

    thread = Thread()
    thread.start()
    print("join")
    thread.join()
    print("done")
    
test_thread()
```

There are some TODOs in `_thread.cs` so that might be the issue.

in-code-i-trust:
Hi, 
I wrote a patch to resolve #410.  

The code shown in #410 causes a deadlock because `_set_sentinel()` is not implemented as `threading.py` expects; an lock object returned by `_set_sentinel()` should be released on its thread exit.  
  
I also implemented a lacking feature, timeout for `acquire()` as well. 
According to [the documentation](https://docs.python.org/3/library/threading.html#threading.Lock.acquire), timeout is only supported on blocking acquire.

### DIFF
--- a/Src/IronPython.Modules/_thread.cs
+++ b/Src/IronPython.Modules/_thread.cs
@@ -23,6 +23,7 @@ namespace IronPython.Modules {
 
         private static readonly object _stackSizeKey = new object();
         private static object _threadCountKey = new object();
+        [ThreadStatic] private static List<@lock> _sentinelLocks;
 
         [SpecialName]
         public static void PerformModuleReload(PythonContext/*!*/ context, PythonDictionary/*!*/ dict) {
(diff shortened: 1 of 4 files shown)


Src/IronPython.Modules/_thread.cs
```diff
@@ -23,6 +24,7 @@ public static class PythonThread {
 
         private static readonly object _stackSizeKey = new object();
         private static object _threadCountKey = new object();
+        private static ConcurrentDictionary<int, Lazy<@lock>> _sentinelTable = new ConcurrentDictionary<int, Lazy<@lock>>();
```
slozier: Might be a good candidate for a `ThreadStatic`?
```C#
[ThreadStatic]
private static @lock _sentinel;
```

in-code-i-trust:
Okay..., this PR is completely messed up. I’m sorry. I will fix it. 
Thank you for rerunning tests and the review comments.

in-code-i-trust:
@slozier 
I've fixed the problems and enabled `test_winreg`.  As you mentioned, the test code at #410 is now in `test_regressions`.  
  
So here are what i've done in this PR.  
- Threads can now have multiple sentinel locks. 
  - This makes `test-all` passes.
- Implemented some lacking features in `winreg.cs`.
  - This makes `test_winreg` passes, so it's been enabled.
- Ported the codes shown in #410 to `test_regressions.py`.

Src/IronPython.Modules/_thread.cs
in-code-i-trust: Never knew that. Replaced with it in commit 972a7bd.

PR #729, opened 2020-01-19
Add support for struct.iter_unpack

### LINKED ISSUE #675: Add support for struct.iter_unpack
Should also implement `struct.Struct.iter_unpack` at the same time.

See https://docs.python.org/3/library/struct.html#struct.iter_unpack for details.

Tests in `CPython.test_struct`.

in-code-i-trust:
...and this should resolve #675.

### DIFF
--- a/Src/IronPython.Modules/_struct.cs
+++ b/Src/IronPython.Modules/_struct.cs
@@ -373,6 +373,15 @@ namespace IronPython.Modules {
                 return unpack_from(context, buffer.ToByteArray(), offset);
             }
 
+            [Documentation("iteratively unpack the current format from the specified array.")]
+            public PythonUnpackIterator iter_unpack(CodeContext/*!*/ context, [BytesConversion][NotNull]IList<byte>/*!*/ buffer, int offset = 0) {
+                return new PythonUnpackIterator(this, context, buffer, offset);
+            }
+
+            [Documentation("iteratively unpack the current format from the specified array.")]
+            public PythonUnpackIterator iter_unpack(CodeContext/*!*/ context, [NotNull]ArrayModule.array/*!*/ buffer, int offset = 0) {
+                return new PythonUnpackIterator(this, context, buffer, offset);
+            }
 
             [Documentation("gets the number of bytes that the serialized string will occupy or are required to deserialize the data")]
             public int size {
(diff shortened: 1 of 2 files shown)


in-code-i-trust:
The reason of CI failures was the test for `iter_unpack` had a dependency on 'n'/'N' formats.

Src/IronPython.Modules/_struct.cs
```diff
@@ -538,6 +547,72 @@ public class Struct : IWeakReferenceable {
...
+                _owner = owner;
+
+                Reset();
+            }
+
+            public System.Collections.IEnumerator GetEnumerator() {
+                return this;
+            }
+
+            public object Current => _iter_current;
```
slozier: Can you add a `PythonHidden` attribute here so the .NET members don't appear in `dir`? Similaly for `GetEnumerator`, `MoveNext` and `Reset`.
```
>>> dir(struct.iter_unpack("h", b""))
['Current', 'GetEnumerator', 'MoveNext', 'Reset', '__class__', ...
```

```C#
[PythonHidden]
public object Current => _iter_current;
```

Src/IronPython.Modules/_struct.cs
```diff
@@ -538,6 +547,72 @@ public class Struct : IWeakReferenceable {
             #endregion
         }
 
+        [PythonType("unpack_iterator"), Documentation("Represents an iterator returned by _struct.iter_unpack()")]
+        public class PythonUnpackIterator : System.Collections.IEnumerator, System.Collections.IEnumerable {
+            private object _iter_current;
+            private int _next_offset;
+
+            private CodeContext _context;
```
slozier: Maybe mark these as `readonly`?

in-code-i-trust:
> Looks good overall, we should probably check that the buffer is of a correct size like CPython does.
> 
> ```
> >>> import struct
> >>> struct.iter_unpack("h", b"\0")
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> struct.error: iterative unpacking requires a buffer of a multiple of 2 bytes
> ```

Thank you for reviewing and the comments, fixed in 8830349.

Src/IronPython.Modules/_struct.cs
in-code-i-trust: Added the attribute in 8830349.

PR #734, opened 2020-02-02
Get __class__ keyword works without AST modification

### LINKED ISSUE #723: __class__ is replaced with the name of the class in the AST
https://github.com/IronLanguages/ironpython3/pull/721 replaces `__class__` with the name of the class in the AST.

```Python
import ast
x = """
class test(object):
    def test(self):
        return __class__
"""
a = compile(x, "", "exec", flags=0x400)
assert a.body[0].body[0].body[0].value.id == "__class__"
```

Related to https://github.com/IronLanguages/ironpython3/issues/722

in-code-i-trust:
Hi, this resolves #723.    
As the subject says, the AST modification (introduced in #721) has been removed.  
  
Changes are...
1. Reverted #721 
2. Introduced `GetGrandParentContextFromFunction()` to make `__class__` handling code works. c.f.) FunctionDefinition.cs
3. Imported test from #723

### DIFF
--- a/Src/IronPython/Runtime/CodeContext.cs
+++ b/Src/IronPython/Runtime/CodeContext.cs
@@ -20,15 +20,17 @@ namespace IronPython.Runtime {
     public sealed class CodeContext {
         private readonly ModuleContext/*!*/ _modContext;
         private readonly PythonDictionary/*!*/ _dict;
+        private readonly CodeContext _maybegrandparent;// can be null.
 
         /// <summary>
         /// Creates a new CodeContext which is backed by the specified Python dictionary.
         /// </summary>
-        public CodeContext(PythonDictionary/*!*/ dict, ModuleContext/*!*/ moduleContext) {
+        public CodeContext(PythonDictionary/*!*/ dict, ModuleContext/*!*/ moduleContext, CodeContext maybegrandparent) {
             ContractUtils.RequiresNotNull(dict, nameof(dict));
             ContractUtils.RequiresNotNull(moduleContext, nameof(moduleContext));
             _dict = dict;
             _modContext = moduleContext;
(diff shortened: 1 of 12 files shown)


Src/IronPython/Runtime/CodeContext.cs
```diff
@@ -229,6 +231,8 @@ internal class DebugProxy {
             }
         }
 
+        internal CodeContext GrandParentContext => _maybegrandparent;
```
slozier: Maybe drop the `_maybegrandparent` field and use an auto-implemented property?
```C#
internal CodeContext GrandParentContext { get; }
```

Src/IronPython/Runtime/CodeContext.cs
```diff
@@ -20,15 +20,17 @@ namespace IronPython.Runtime {
     public sealed class CodeContext {
         private readonly ModuleContext/*!*/ _modContext;
         private readonly PythonDictionary/*!*/ _dict;
+        private readonly CodeContext _maybegrandparent;// can be null.
 
         /// <summary>
         /// Creates a new CodeContext which is backed by the specified Python dictionary.
         /// </summary>
-        public CodeContext(PythonDictionary/*!*/ dict, ModuleContext/*!*/ moduleContext) {
+        public CodeContext(PythonDictionary/*!*/ dict, ModuleContext/*!*/ moduleContext, CodeContext maybegrandparent) {
```
slozier: Is this a concept that should be exposed publicly? Maybe add an internal constructor overload instead?

slozier:
Thanks for the PR! Closing this since the issue has been resolved by https://github.com/IronLanguages/ironpython3/pull/1087

PR #737, opened 2020-02-15
Remove MakeRethrowExceptionWorker

slozier:
No longer necessary after https://github.com/IronLanguages/ironpython3/pull/735


slozier:
Resolves the issue, but breaks the traceback of:
```Python
def g():
    f()

def f():
    try:
        raise Exception
    except:
        raise

g()
```

PR #739, opened 2020-02-17
Get test_compare passing

gfmcknight:
Default to negating the result of `__eq__` when the `__ne__` slot is not
defined. Enable test_compare.py.

Looks like the earlier issues in this file have been fixed by other changes. According to the definition at `https://docs.python.org/3/reference/datamodel.html#object.__ne__`,

> By default, __ne__() delegates to __eq__() and inverts the result unless it is NotImplemented. There are no other implied relationships among the comparison operators, for example, the truth of (x<y or x==y) does not imply x<=y.

It doesn't look like the behavior the test is looking for has changed in later Python versions; in 3.8,
```
class C:
    def __init__(self):
        pass
    def __ne__(self, other):
        print('A')
        return NotImplemented
    def __eq__(self, other):
        print('B')
        return True

print(C() != 3)
```
Prints
```
A
True
```

### DIFF
--- a/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
+++ b/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
@@ -1011,7 +1011,6 @@ namespace IronPython.Runtime.Binding {
 
             SlotOrFunction fop = SlotOrFunction.GetSlotOrFunction(state, opSym, types);
             SlotOrFunction rop = SlotOrFunction.GetSlotOrFunction(state, ropSym, rTypes);
-
             SlotOrFunction cmp = SlotOrFunction.GetSlotOrFunction(state, "__cmp__", types);
             SlotOrFunction rcmp = SlotOrFunction.GetSlotOrFunction(state, "__cmp__", rTypes);
 
@@ -1031,7 +1030,6 @@ namespace IronPython.Runtime.Binding {
                 SlotOrFunction tmp = rop;
                 rop = fop;
                 fop = tmp;
-
                 opReverse = true;
             }
 
--- a/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
+++ b/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
@@ -857,11 +857,6 @@ namespace IronPython.Runtime.Binding {
             }
         }
 
-        private static void MakeNegatedCompareReturn(ConditionalBuilder/*!*/ bodyBuilder, Expression retCondition, Expression/*!*/ retValue, bool isReverse, Type retType) {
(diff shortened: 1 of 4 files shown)


Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
```diff
@@ -857,6 +857,11 @@ internal static partial class PythonProtocol {
             }
         }
 
+        private static void MakeNegatedCompareReturn(ConditionalBuilder/*!*/ bodyBuilder, Expression retCondition, Expression/*!*/ retValue, bool isReverse, Type retType) {
+            Expression newRetVal = Ast.Not(Ast.Call(Compiler.Ast.AstMethods.IsTrue, retValue));
```
gfmcknight: I think that the dynamic conversion is necessary here because we have methods that return an object type, which can be a boolean or NotImplemented.

slozier:
Thanks for the PR! I'm not entirely sure this is the right approach. I think this should probably be solved via a default implementation of `__ne__` on the object base type? See `ObjectOps.cs` in https://github.com/IronLanguages/ironpython3/pull/709 for example. Any thoughts? Is this something you'd like to look into?

I'm probably fine with merging this in to enable the test but if so we should add a comment to get rid of it as part of the `__cmp__` removal.

gfmcknight:
That makes sense -- the comments in the test suggest that it's implemented by having `__ne__`. I'll take a look if you don't mind me stepping on the PR you already have out.

slozier:
Go right ahead. I was going to scrap that PR and start over anyway. :smile:

gfmcknight:
This has an interesting side effect for overload resolution of operators that I'm still working through. A statement like: `System.Single(0) != System.Decimal(0)` now calls `ObjectOps.__ne__()` instead of the not equals operator defined for floats because that operator narrows the decimal to a float.

Then `TryInvokeBinaryOperator` for some reason doesn't find `op_Inequality`, so instead it falls back to comparing the types.

Src/IronPython/Runtime/Types/PythonTypeInfo.cs
```diff
@@ -1728,7 +1728,8 @@ public LookupBinder(PythonBinder/*!*/ binder)
 
             // numeric types in python don't define equality, just __cmp__
             if (t == typeof(bool) ||
-                (Converter.IsNumeric(t) && t != typeof(Complex) && t != typeof(double) && t != typeof(float))) {
+                (Converter.IsNumeric(t) && t != typeof(Complex) && t != typeof(double) && t != typeof(float)) &&
+                t != typeof(Decimal)) {
```
gfmcknight: Do we eventually want this constraint to disappear altogether, if `__cmp__` is going away?

slozier: Yeah, I think the whole method becomes obsolete once `__cmp__` is gone.

PR #761, opened 2020-03-04
Enable analyzer on IronPython.Modules

slozier:


Src/IronPython.Modules/_codecs.cs
```diff
@@ -112,7 +112,7 @@ public static PythonTuple ascii_encode(CodeContext context, [NotNull]string inpu
 
         #region Generic Encoding
 
-        public static object decode(CodeContext/*!*/ context, object obj, [NotNull]string? encoding = null, [NotNull]string errors = "strict") {
+        public static object decode(CodeContext/*!*/ context, object? obj, [NotNull, DisallowNull]string? encoding = null, [NotNull]string errors = "strict") {
```
BCSharp: `object obj` => `object? obj`: good catch! I haven't noticed that.

```c#
[NotNul, DisallowNull]string? encoding = null
```
I didn't know that `DisallowNull` could be applied on parameters with a default value `null`. Nice. The compiler rejects `DisallowNull` on fields initialized to `null`.

Another thing, about the coding style. I tend not to use spaces when listing multiple attributes on parameters for visual effect. Do you prefer it with spaces?

slozier: I didn't know about it either until I tried. Only reason I'm adding it is that my analyzer disallows `[NotNull]object?` and this seems like a reasonable use case.

As for the spaces when listing multiple attributes I don't have a specific preference. I tend to go with whatever Visual Studio 2019 thinks is best. In this case, when closing the end bracket of the method VS automatically inserts a a space between attributes which is why I use the space. If there's an editorconfig setting to drop the space I would be fine with switching to it.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -162,6 +164,8 @@ public static partial class PythonOps {
             }
         }
 
+#nullable disable
```
BCSharp: `#nullable restore`  ?

Src/IronPython.Modules/_codecs.cs
BCSharp: I could not find an appropriate editorconfig setting, I guess it has to be the VS way.

PR #764, opened 2020-03-09
Array cleanup

slozier:
Cleanup `array` and extract `ArrayData` to make it the backing field for `ByteArray`. This allows us to have access to the underlying array.

Related to the discussion on https://github.com/IronLanguages/ironpython3/pull/762


BCSharp:
:+1:

Since `ByteArray` and `ArrayData` implement `IList<byte>`/`IList<T>`, may I suggest that they also implement `IReadOnlyList`? It may facilitate adoption of `IReadOnlyList` throughout the codebase.

PR #765, opened 2020-03-10
Unsafe encoding

BCSharp:
This the first step of a feasibility study of using `Span` for bytes-like objects (triggered by #762). I address here my first concern for encodings not having access to the underlying `byte[]` hence needing to copy data to a new array every time on decode.

I did the experiments in net45; I didn't want to get sidetracked by the target upgrade. Once the project target is upgraded to .NET 4.6, we can use/override `Encoding.GetString(bytes*, ...)`, which will eliminate a few copy actions in my feasibility code. Since those changes are straightforward, there is no need to address them in this exercise.

My first concern was the feasibility of `Span` (i.e. byte pointer overloads) for `UnicodeEscapeEncoding`. This encoding depends on `LiteralParser`, which in its turn serves two masters, `UnicodeEscapeEncoding` (which used `Ilist<byte>`) and `Tokenizer` (which used `IList<char>`). In #671 we managed to cover both usecases by a generic method, which was not straightforward to get right so I wasn't sure how it would go here. I am happy to see that it actually went quite fine. Commit e0f7a580f0571ec289bcf5e0c2bdb304c2484f49 shows the result.

It was nice to see that the `IConvertible.ToChar(null)` on `Span<byte>` still works the same as for `IList<byte>`. In fact the whole expression `data[i].ToChar(null)` is somewhat simpler for `Span` than `IList`, probably because `get_Item` call site is simpler. Interesting observation: `get_Item` on `Span` is marginally faster (by one assembly instruction) than on `ReadOnlySpan`. I am guessing that this is because `Span` is being passed by value, while `ReadOnlySpan` by reference. In a loop like this it adds up and probably is more than the extra cost of passing by value. I tried to force the compiler to pass `ReadOnlySpan` by value, by **not** using the `in` parameter modifier, but it had no effect. The generated IL code still had the parameter tagged with `InAttribute`. Maybe because `ReadOnlySpan` as a whole is a `readonly` struct, so compiler "optimization" kicks in. In the end, not a big deal but it is strange to see that the compiler assumes it "knows better" and ignores what the programmer wrote.

In the next commit I go further along the feasibility of spans for decoding functions. Unfortunately, I have encountered some roadblocks there.

### DIFF
--- a/Src/IronPython.Modules/_codecs.cs
+++ b/Src/IronPython.Modules/_codecs.cs
@@ -54,8 +54,8 @@ namespace IronPython.Modules {
 
         #region ASCII Encoding
 
-        public static PythonTuple ascii_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null)
-            => DoDecode(context, "ascii", PythonAsciiEncoding.Instance, input, errors, input.Count).ToPythonTuple();
+        public static PythonTuple ascii_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null)
+            => DoDecode(context, "ascii", PythonAsciiEncoding.Instance, input, errors, input.Length).ToPythonTuple();
 
         public static PythonTuple ascii_encode(CodeContext context, [NotNull]string input, string? errors = null)
             => DoEncode(context, "ascii", PythonAsciiEncoding.Instance, input, errors).ToPythonTuple();
@@ -94,14 +94,14 @@ namespace IronPython.Modules {
         /// <summary>
         /// Decodes the input string using the provided string mapping.
         /// </summary>
-        public static PythonTuple charmap_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors, [NotNull]string map) {
+        public static PythonTuple charmap_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors, [NotNull]string map) {
             EncodingMap em = new EncodingMap(map, compileForDecoding: true, compileForEncoding: false);
-            return DoDecode(context, "charmap", new EncodingMapEncoding(em), input, errors, input.Count).ToPythonTuple();
+            return DoDecode(context, "charmap", new EncodingMapEncoding(em), input, errors, input.Length).ToPythonTuple();
         }
 
-        public static PythonTuple charmap_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null, IDictionary<object, object>? map = null) {
+        public static PythonTuple charmap_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null, IDictionary<object, object>? map = null) {
             if (map != null) {
-                return DoDecode(context, "charmap", new CharmapEncoding(map), input, errors, input.Count).ToPythonTuple();
+                return DoDecode(context, "charmap", new CharmapEncoding(map), input, errors, input.Length).ToPythonTuple();
             } else {
                 return latin_1_decode(context, input, errors);
             }
@@ -114,10 +114,11 @@ namespace IronPython.Modules {
 
         public static object decode(CodeContext/*!*/ context, object? obj, [NotNull, DisallowNull]string? encoding = null, [NotNull]string errors = "strict") {
             if (encoding == null) {
-                if (obj is IList<byte> bytesLikeObj) { // TODO: replace IList with IBufferProtocol
-                    var span = bytesLikeObj.ToArray().AsSpan(); // TODO: use IBufferProtocol.ToSpan
-                    PythonContext lc = context.LanguageContext;
-                    return StringOps.DoDecode(context, span, errors, lc.GetDefaultEncodingName(), lc.DefaultEncoding);
+                PythonContext lc = context.LanguageContext;
+                if (obj is ReadOnlyMemory<byte> bytesLikeObj) {
+                    return StringOps.DoDecode(context, bytesLikeObj, errors, lc.GetDefaultEncodingName(), lc.DefaultEncoding);
+                } else if (obj is IBufferProtocol bp) {
+                    return StringOps.DoDecode(context, bp.ToMemory(), errors, lc.GetDefaultEncodingName(), lc.DefaultEncoding);
                 } else {
                     throw PythonOps.TypeError("expected bytes-like object, got {0}", PythonTypeOps.GetName(obj));
                 }
@@ -152,13 +153,12 @@ namespace IronPython.Modules {
         #region Escape Encoding
 
         public static PythonTuple escape_decode(CodeContext/*!*/ context, [NotNull]string data, string? errors = null)
-            => escape_decode(StringOps.DoEncodeUtf8(context, data), errors);
+            => escape_decode(StringOps.DoEncodeUtf8(context, data).AsMemory(), errors);
 
-        public static PythonTuple escape_decode([BytesLike,NotNull]IList<byte> data, string? errors = null) {
-            var span = data.ToArray().AsSpan(); // TODO: remove when signature changed
-            var res = LiteralParser.ParseBytes(span, isRaw: false, normalizeLineEndings: false, getErrorHandler(errors));
+        public static PythonTuple escape_decode([BytesLike,NotNull]ReadOnlyMemory<byte> data, string? errors = null) {
+            var res = LiteralParser.ParseBytes(data.Span, isRaw: false, normalizeLineEndings: false, getErrorHandler(errors));
 
-            return PythonTuple.MakeTuple(Bytes.Make(res.ToArray()), data.Count);
+            return PythonTuple.MakeTuple(Bytes.Make(res.ToArray()), data.Length);
 
             static LiteralParser.ParseBytesErrorHandler<byte>? getErrorHandler(string? errors) {
                 if (errors == null) return default;
@@ -180,9 +180,11 @@ namespace IronPython.Modules {
         [DisallowNull]
         private static byte[]? _replacementMarker;
 
-        public static PythonTuple/*!*/ escape_encode([BytesLike,NotNull]IList<byte> data, string? errors = null) {
-            List<byte> buf = new List<byte>(data.Count);
-            foreach (byte b in data) {
+        public static PythonTuple/*!*/ escape_encode([BytesLike,NotNull]ReadOnlyMemory<byte> data, string? errors = null) {
+            var buf = new List<byte>(data.Length);
+            var span = data.Span;
+            for (int i = 0; i < span.Length; i++) {
+                byte b = span[i];
                 switch (b) {
                     case (byte)'\n': buf.Add((byte)'\\'); buf.Add((byte)'n'); break;
                     case (byte)'\r': buf.Add((byte)'\\'); buf.Add((byte)'r'); break;
@@ -198,15 +200,15 @@ namespace IronPython.Modules {
                         break;
                 }
             }
-            return PythonTuple.MakeTuple(Bytes.Make(buf.ToArray()), data.Count);
+            return PythonTuple.MakeTuple(Bytes.Make(buf.ToArray()), data.Length);
         }
 
         #endregion
 
         #region Latin-1 Functions
 
-        public static PythonTuple latin_1_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null)
-            => DoDecode(context, "latin-1", Encoding.GetEncoding("iso-8859-1"), input, errors, input.Count).ToPythonTuple();
+        public static PythonTuple latin_1_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null)
+            => DoDecode(context, "latin-1", Encoding.GetEncoding("iso-8859-1"), input, errors, input.Length).ToPythonTuple();
 
         public static PythonTuple latin_1_encode(CodeContext context, [NotNull]string input, string? errors = null)
             => DoEncode(context, "latin-1", Encoding.GetEncoding("iso-8859-1"), input, errors).ToPythonTuple();
@@ -217,8 +219,8 @@ namespace IronPython.Modules {
 #if FEATURE_ENCODING
 
         [PythonHidden(PlatformsAttribute.PlatformFamily.Unix)]
-        public static PythonTuple mbcs_decode(CodeContext/*!*/ context, [BytesLike,NotNull]IList<byte> input, string? errors = null, bool final = false)
-            => DoDecode(context, "mbcs", MbcsEncoding, input, errors, input.Count).ToPythonTuple();
+        public static PythonTuple mbcs_decode(CodeContext/*!*/ context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null, bool final = false)
+            => DoDecode(context, "mbcs", MbcsEncoding, input, errors, input.Length).ToPythonTuple();
 
         [PythonHidden(PlatformsAttribute.PlatformFamily.Unix)]
         public static PythonTuple mbcs_encode(CodeContext/*!*/ context, [NotNull]string input, string? errors = null)
@@ -231,11 +233,11 @@ namespace IronPython.Modules {
 #if FEATURE_ENCODING
 
         [PythonHidden(PlatformsAttribute.PlatformFamily.Unix)]
-        public static PythonTuple code_page_decode(CodeContext context, int codepage, [BytesLike, NotNull]IList<byte> input, string? errors = null, bool final = false) {
+        public static PythonTuple code_page_decode(CodeContext context, int codepage, [BytesLike, NotNull]ReadOnlyMemory<byte> input, string? errors = null, bool final = false) {
             // TODO: Use Win32 API MultiByteToWideChar https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar
             string encodingName = $"cp{codepage}";
             Encoding encoding = Encoding.GetEncoding(codepage);
-            return DoDecode(context, encodingName, encoding, input, errors, input.Count).ToPythonTuple();
+            return DoDecode(context, encodingName, encoding, input, errors, input.Length).ToPythonTuple();
         }
 
         [PythonHidden(PlatformsAttribute.PlatformFamily.Unix)]
@@ -255,14 +257,13 @@ namespace IronPython.Modules {
         public static PythonTuple raw_unicode_escape_decode(CodeContext/*!*/ context, [NotNull]string input, string? errors = null) {
             // Encoding with UTF-8 is probably a bug or at least a mistake, as it mutilates non-ASCII characters,
             // but this is what CPython does. Probably encoding with "raw-unicode-escape" would be more reasonable.
-            return raw_unicode_escape_decode(context, StringOps.DoEncodeUtf8(context, input), errors);
+            return raw_unicode_escape_decode(context, StringOps.DoEncodeUtf8(context, input).AsMemory(), errors);
         }
 
-        public static PythonTuple raw_unicode_escape_decode(CodeContext/*!*/ context, [BytesLike,NotNull]IList<byte> input, string? errors = null) {
-            var span = input.ToArray().AsSpan(); // TODO: remove when signature changed
+        public static PythonTuple raw_unicode_escape_decode(CodeContext/*!*/ context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null) {
             return PythonTuple.MakeTuple(
-                StringOps.DoDecode(context, span, errors, "raw-unicode-escape", StringOps.CodecsInfo.RawUnicodeEscapeEncoding),
-                input.Count
+                StringOps.DoDecode(context, input, errors, "raw-unicode-escape", StringOps.CodecsInfo.RawUnicodeEscapeEncoding),
+                input.Length
             );
         }
 
@@ -282,14 +283,13 @@ namespace IronPython.Modules {
         public static PythonTuple unicode_escape_decode(CodeContext/*!*/ context, [NotNull]string input, string? errors = null) {
             // Encoding with UTF-8 is probably a bug or at least a mistake, as it mutilates non-ASCII characters,
             // but this is what CPython does. Probably encoding with "unicode-escape" would be more reasonable.
-            return unicode_escape_decode(context, StringOps.DoEncodeUtf8(context, input), errors);
+            return unicode_escape_decode(context, StringOps.DoEncodeUtf8(context, input).AsMemory(), errors);
         }
 
-        public static PythonTuple unicode_escape_decode(CodeContext/*!*/ context, [BytesLike,NotNull]IList<byte> input, string? errors = null) {
-            var span = input.ToArray().AsSpan(); // TODO: remove when signature changed
+        public static PythonTuple unicode_escape_decode(CodeContext/*!*/ context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null) {
             return PythonTuple.MakeTuple(
-                StringOps.DoDecode(context, span, errors, "unicode-escape", StringOps.CodecsInfo.UnicodeEscapeEncoding),
-                input.Count
+                StringOps.DoDecode(context, input, errors, "unicode-escape", StringOps.CodecsInfo.UnicodeEscapeEncoding),
+                input.Length
             );
         }
 
@@ -306,10 +306,10 @@ namespace IronPython.Modules {
         #region Readbuffer Functions
 
         public static PythonTuple readbuffer_encode(CodeContext/*!*/ context, [NotNull]string input, string? errors = null)
-            => readbuffer_encode(DoEncode(context, "utf-8", Encoding.UTF8, input, "strict").Item1, errors);
+            => DoEncode(context, "utf-8", Encoding.UTF8, input, "strict").ToPythonTuple();
 
-        public static PythonTuple readbuffer_encode([BytesLike,NotNull]IList<byte> input, string? errors = null)
-            => PythonTuple.MakeTuple(new Bytes(input), input.Count);
+        public static PythonTuple readbuffer_encode([BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null)
+            => PythonTuple.MakeTuple(new Bytes(input), input.Length);
 
         #endregion
 
@@ -320,9 +320,9 @@ namespace IronPython.Modules {
             return PythonTuple.MakeTuple(input, input.Length);
         }
 
-        public static PythonTuple unicode_internal_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null) {
+        public static PythonTuple unicode_internal_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null) {
             PythonOps.Warn(context, PythonExceptions.DeprecationWarning, "unicode_internal codec has been deprecated");
-            return DoDecode(context, "unicode-internal", Encoding.Unicode, input, errors, input.Count).ToPythonTuple();
+            return DoDecode(context, "unicode-internal", Encoding.Unicode, input, errors, input.Length).ToPythonTuple();
         }
 
         public static PythonTuple unicode_internal_encode(CodeContext context, [NotNull]string input, string? errors = null) {
@@ -330,16 +330,16 @@ namespace IronPython.Modules {
             return DoEncode(context, "unicode-internal", Encoding.Unicode, input, errors, false).ToPythonTuple();
         }
 
-        public static PythonTuple unicode_internal_encode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null) {
+        public static PythonTuple unicode_internal_encode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null) {
             PythonOps.Warn(context, PythonExceptions.DeprecationWarning, "unicode_internal codec has been deprecated");
-            return PythonTuple.MakeTuple(new Bytes(input), input.Count);
+            return PythonTuple.MakeTuple(new Bytes(input), input.Length);
         }
 
         #endregion
 
         #region Utf-16 Functions
 
-        public static PythonTuple utf_16_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null, bool final = false) {
+        public static PythonTuple utf_16_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null, bool final = false) {
             PythonTuple res = utf_16_ex_decode(context, input, errors, 0, final);
             return PythonTuple.MakeTuple(res[0], res[1]);
         }
@@ -347,35 +347,36 @@ namespace IronPython.Modules {
         public static PythonTuple utf_16_encode(CodeContext context, [NotNull]string input, string? errors = null)
             => DoEncode(context, "utf-16", Utf16LeBomEncoding, input, errors, true).ToPythonTuple();
 
-        public static PythonTuple utf_16_ex_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null, int byteorder = 0, bool final = false) {
+        public static PythonTuple utf_16_ex_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null, int byteorder = 0, bool final = false) {
 
+            var span = input.Span;
             Tuple<string, int> res;
 
             if (byteorder != 0) {
                 res = (byteorder > 0) ?
-                    DoDecode(context, "utf-16-be", Utf16BeEncoding, input, errors, NumEligibleUtf16Bytes(input, final, false))
+                    DoDecode(context, "utf-16-be", Utf16BeEncoding, input, errors, NumEligibleUtf16Bytes(span, final, false))
                 :
-                    DoDecode(context, "utf-16-le", Utf16LeEncoding, input, errors, NumEligibleUtf16Bytes(input, final, true));
+                    DoDecode(context, "utf-16-le", Utf16LeEncoding, input, errors, NumEligibleUtf16Bytes(span, final, true));
 
             } else {
-                byteorder = Utf16DetectByteorder(input);
+                byteorder = Utf16DetectByteorder(span);
                 res = (byteorder > 0) ?
-                    DoDecode(context, "utf-16-be", Utf16BeBomEncoding, input, errors, NumEligibleUtf16Bytes(input, final, false))
+                    DoDecode(context, "utf-16-be", Utf16BeBomEncoding, input, errors, NumEligibleUtf16Bytes(span, final, false))
                 :
-                    DoDecode(context, "utf-16-le", Utf16LeBomEncoding, input, errors, NumEligibleUtf16Bytes(input, final, true));
+                    DoDecode(context, "utf-16-le", Utf16LeBomEncoding, input, errors, NumEligibleUtf16Bytes(span, final, true));
             }
 
             return PythonTuple.MakeTuple(res.Item1, res.Item2, byteorder);
         }
 
-        private static int Utf16DetectByteorder(IList<byte> input) {
+        private static int Utf16DetectByteorder(ReadOnlySpan<byte> input) {
             if (input.StartsWith(BOM_UTF16_LE)) return -1;
             if (input.StartsWith(BOM_UTF16_BE)) return 1;
             return 0;
         }
 
-        private static int NumEligibleUtf16Bytes(IList<byte> input, bool final, bool isLE) {
-            int numBytes = input.Count;
+        private static int NumEligibleUtf16Bytes(ReadOnlySpan<byte> input, bool final, bool isLE) {
+            int numBytes = input.Length;
             if (!final) {
                 numBytes -= numBytes % 2;
                 if (numBytes >= 2 && (input[numBytes - (isLE ? 1 : 2)] & 0xFC) == 0xD8) { // high surrogate
@@ -397,8 +398,8 @@ namespace IronPython.Modules {
         private static byte[] BOM_UTF16_LE => _bom_utf16_le ??= Utf16LeBomEncoding.GetPreamble();
         [DisallowNull] private static byte[]? _bom_utf16_le;
 
-        public static PythonTuple utf_16_le_decode(CodeContext context, [BytesLike,NotNull]IList<byte> input, string? errors = null, bool final = false)
-            => DoDecode(context, "utf-16-le", Utf16LeEncoding, input, errors, NumEligibleUtf16Bytes(input, final, isLE: true)).ToPythonTuple();
+        public static PythonTuple utf_16_le_decode(CodeContext context, [BytesLike,NotNull]ReadOnlyMemory<byte> input, string? errors = null, bool final = false)
+            => DoDecode(context, "utf-16-le", Utf16LeEncoding, input, errors, NumEligibleUtf16Bytes(input.Span, final, isLE: true)).ToPythonTuple();
 
         public static PythonTuple utf_16_le_encode(CodeContext context, [NotNull]string input, string? errors = null)
             => DoEncode(context, "utf-16-le", Utf16LeEncoding, input, errors).ToPythonTuple();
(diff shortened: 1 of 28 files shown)


Src/IronPython/IronPython.csproj
```diff
@@ -59,6 +60,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="System.Memory" Version="4.5.3" Condition="'$(TargetFramework)' == 'net45' or '$(TargetFramework)' == 'netstandard2.0'" />
```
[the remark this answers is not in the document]
BCSharp: Indeed. What about `PrivateAssets` and `IncludeAssets`? I'll leave it to you to update the project file on the master branch.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1764,9 +1764,24 @@ internal static string DoDecode(CodeContext context, IList<byte> s, string error
...
 #endif
 
-            string decoded;
+            string decoded = string.Empty;
             try {
-                decoded = e.GetString(bytes, start, length);
+                unsafe {
+                    ReadOnlySpan<byte> data = s.ToArray();
+                    fixed (byte* bp = data.Slice(start)) {
+                        if (bp != null) {
```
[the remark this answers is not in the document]
BCSharp: That's right.

Src/IronPython/Runtime/LiteralParser.cs
```diff
@@ -35,27 +35,41 @@ public static class LiteralParser {
...
             return result ?? bytes.MakeString(start, length);
         }
 
-        private static string DoParseString<T>(T[] data, int start, int length, bool isRaw, bool isUniEscape, bool normalizeLineEndings, ParseStringErrorHandler<T> errorHandler = default) where T : IConvertible {
-            Bytes bytesData = null;
+        internal static string ParseString(in ReadOnlySpan<byte> bytes, bool isRaw, ParseStringErrorHandler<byte> errorHandler) {
+            string result = DoParseString(bytes, isRaw, isUniEscape: true, normalizeLineEndings: false, errorHandler);
+
+            if (result == null) {
+                if (bytes.IsEmpty) {
```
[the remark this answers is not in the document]
BCSharp: I agree. I had more improvements in mind here. It seems that we use the Latin-1 codec a lot, so I was thinking of making somewhere a global static property `Encoding Latin1` rather than calling `Encoding.GetEncoding` every time or looping ourselves. Just like [.NET has it](https://referencesource.microsoft.com/#mscorlib/system/text/encoding.cs,0f9d38a1189e6308) (and [.NET Core too](https://source.dot.net/#System.Private.CoreLib/Encoding.cs,0f9d38a1189e6308)), unfortunately, it is `private`.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -2222,6 +2237,25 @@ public override int GetBytes(char[] chars, int charIndex, int charCount, byte[]
...
+            public override unsafe int GetCharCount(byte* bytes, int count) {
+                return GetChars(bytes, count, null, 0);
+            }
+
+            public override unsafe int GetChars(byte* bytes, int byteCount, char* chars, int charCount) {
+                var span = new Span<byte>(bytes, byteCount);
+                var res = LiteralParser.ParseString(span, _raw, GetErrorHandler());
+
+                if (chars != null) {
+                    int i;
```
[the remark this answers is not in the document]
BCSharp: Yes, it looks nicer and is faster. I am still getting used to spans. Though after upgrade to .NET 4.5 we will not be using this method.

Src/IronPython/Runtime/LiteralParser.cs
```diff
@@ -257,17 +256,18 @@ public static class LiteralParser {
...
-                            } else if (i < l && data[i].ToChar(null) == '\n') {
+                            } else if (i < length && data[i].ToChar(null) == '\n') {
                                 i++;
                             }
                             continue;
                         case 'x': //hex
                             if (!TryParseInt(data, i, 2, 16, out val, out int consumed)) {
-                                int pos = i - start - 2;
+                                int pos = i - 2;
+                                string message = $"invalid \\x escape at position {pos}";
```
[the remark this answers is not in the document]
BCSharp: We don't have to (at least, not yet). It is a left-over of an experiment with a change intended for another refactoring challenge I am thinking about: reducing functionality duplication by routing all error handling through the `StringOps.*Errors` methods. This has nothing to do with this feasibility study. I'm sorry for the confusion.

Src/IronPython/Runtime/LiteralParser.cs
```diff
@@ -187,58 +201,43 @@ public static class LiteralParser {
...
-                case byte[] bytes:
-                    for (int i = start; i < end; i++) {
-                        sb.Append((char)bytes[i]);
-                    }
-                    break;
+            if (sb != null) return;
 
-                default:
-                    throw new NotImplementedException("Only byte and char arrays supported");
+            sb = new StringBuilder(data.Length);
```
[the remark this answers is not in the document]
BCSharp: Yes, it is worth it here. Looping over span here is a stepback from master for the `char` specialization. Once we upgrade to .NET 4.6 I will port the changes to (hopefully) fully utilize all the 4.6 goodies.

BCSharp:
In this step I changed the signature of `StringOps.DoDecode` from accepting `IList<byte>` to `ReadOnlySpan<byte>` and looked for the ripple effects. I haven't changed signature of all decode functions in `_codecs`, to avoid failing tests or necessary but straightforward Binder updates. Instead of it, I convert `IList<byte>` to an array and create a span over it, to match changes in `StringOps.DoDecode`. This is a data copy, but will disappear after public function signature change. There are more places like that, so I will focus only on data copying that is problematic.

I will start with the major roadblock encountered in `DoDecode`. When the error handler is none of the builtin handler names, `PythonDecoderFallback` is used, which will possibly call any of the user-registered handlers. This fallback class requires a reference to the data being decoded, so I cannot pass `Span`. Hence a data copy to an array. The fact that a copy is being made by itself is acceptable — CPython will also wrap any input data into `bytes` before passing to an error handler — but the copy will happen always for any user-defined error handler, even when no errors are being generated. Even if this were acceptable, it would become a problem with my planned refactoring of error handling to always use `StringOps.*Errors` methods and abolish this switch-on-string we have now on every call. This would make all built-in error handlers be registered and called as any user-defined handler, which can be redefined/overriden. For instance, this works in Python:

```python
import codecs

def my_error_handler(err):
    if isinstance(err, UnicodeDecodeError):
        return ("<error>", err.end)
    else:
        raise err

# redefine every built-in error handler
for eh in ["test", "strict", "replace", "ignore", "xmlcharrefreplace", "backslashreplace", "surrogateescape", "surrogatepass"]:
    codecs.register_error(eh, my_error_handler)
    assert b"a\xffc".decode('utf-8', eh) == "a<error>c"

assert codecs.utf_8_decode(b"a\xffc", None) == ("a<error>c", 3)
```
I see two possibilities here: for `_codecs` functions to use either `IBufferProtocol` or `ReadOnlyMemory<byte>`. Simple functions accepting bytes-like input may still use `Span`. The former option is basically my proposal from #762 ("Choice D with a mixin from B"). It also plays well with `UserDecode` discussed below. Conversion from `IBufferProtocol` to `bytes` (or `Span` or `Memory`) can be delayed to when the data is actually needed. I am pretty confident this will work, so I do not need to test it out in actual code now.

Using `ReadOnlyMemory<byte>` should also be sufficient, though I would continue my experiment to get a better feeling for it. As far I can see now, it will force every object to create a `ReadOnlyMemory` right from the beginning, possibly done by the Binder. Getting `Memory` may be more expensive than `Span`, though it probably does not matter that much. For all builtin types like `Bytes`, this will be the same.

OK, here is the second interesting issue. `StringOps.RawDecode` does not use `Span`, it uses `IProtocolBuffer`. I have found that user decode functions in CPython are being given objects of type `memoryview`. If the input object was already `memoryview`, it is being passed as is to user decode. In case of any other buffer-protocol-compliant object, a `memoryview` wrapper is passed. I am not completely recreating this behavior in `UserDecode` because IronPython's `MemoryView` does not implement `IBufferProtocol`, but I anticipate it will, so I used `IBufferProtocol` as the input data type. @slozier, I remember you mentioned that adding `IBufferProtocol` to `MemoryView` is "technically not correct". Can you elaborate? All I see is that `memoryview` is being accepted by CPython wherever a buffer-protocol-compliant object is required and `dir` on an instance of `memoryview` shows all familiar `Py_Buffer` members.

On a side note, I do not understand why there are two identical in purpose, but different in implementation functions: `_codecs.decode` and `StringOps.RawDecode` (the same for encode). Perhaps this could be straightened out. I will need to investigate it further. It is a separate issue except that if `StringOps.RawDecode` uses `IBufferProtocol` so should `_codecs.decode`. And if so, then perhaps all other decoding functions in `_codecs`?

BCSharp:
I forgot to say: I looked into the performance of `LiteralParser` on .NET Core 3.0 only. I see that there are still [issues with JIT](https://github.com/dotnet/runtime/issues/10950) that impact performance of indexed access. The newest version of `Span` has a workaround for it. Maybe it will drip down to the NuGet package at some point.

BCSharp:
The CI test on Windows fails because of failing net45.IronPython.interop.net.dynamic.test_dynamic_regressions, which fails to load `System.Memory.dll` when executing `os.system("%TEMP%/IronPython/cp20519_vb.exe")`:

```
Unhandled Exception: System.Reflection.TargetInvocationException: Failed to load language 'IronPython 3.4.0': Could not load file or assembly 'System.Memory, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Memory, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
   at IronPython.Modules.PythonIOModule.BufferedReader.__init__(CodeContext context, Object raw, Int32 buffer_size)
   at IronPython.Modules.PythonIOModule.BufferedReader.Create(CodeContext context, Object raw, Int32 buffer_size)
   at IronPython.Modules.PythonIOModule.CreateConsole(PythonContext context, SharedIO io, ConsoleStreamType type, String name, FileIO& fio)
   at IronPython.Runtime.PythonContext.SetStandardIO()
   at IronPython.Runtime.PythonContext.InitializeSystemState()
   at IronPython.Runtime.PythonContext..ctor(ScriptDomainManager manager, IDictionary`2 options)
   --- End of inner exception stack trace ---
   at Microsoft.Scripting.Runtime.LanguageConfiguration.LoadLanguageContext(ScriptDomainManager domainManager, Boolean& alreadyLoaded)
   at Microsoft.Scripting.Runtime.DlrConfiguration.LoadLanguageContext(ScriptDomainManager manager, LanguageConfiguration config)
   at Microsoft.Scripting.Runtime.DlrConfiguration.TryLoadLanguage(ScriptDomainManager manager, String str, Boolean isExtension, LanguageContext& language)     
   at Microsoft.Scripting.Runtime.ScriptDomainManager.TryGetLanguage(String languageName, LanguageContext& language)
   at Microsoft.Scripting.Hosting.ScriptRuntime.TryGetEngine(String languageName, ScriptEngine& engine)
   at Microsoft.Scripting.Hosting.ScriptRuntime.GetEngine(String languageName)
   at CodePlex20519.Main()
-532462766
```

slozier:
> I remember you mentioned that adding IBufferProtocol to MemoryView is "technically not correct". Can you elaborate?

The issue isn't really that `MemoryView` shouldn't implement `IBufferProtocol`. It's that `IBufferProtocol` is all wrong. The protocol itself should consist of only two methods `GetPythonBuffer` and `ReleasePythonBuffer`. When you call `GetPythonBuffer` it should give you a `PythonBuffer` struct, which is equivalent to the `Py_buffer` object.

We could define an `IPythonBuffer` interface and leave it up to the `IBufferProtocol` objects to return something that implements the interface. We could also potentially drop the `ReleasePythonBuffer` if we select this option and make `IPythonBuffer` also implement `IDisposable`.

Anyway this would been prototyping. My goal is to be able to support the following:
```Python
ba = bytearray()
mv = memoryview(ba)
ba.append(0) # this should throw a BufferError since we're trying to resize the buffer
```

If you think implementing the current `IBufferProtocol` on `MemoryView` would be helpful in the short term then we can do that.

BCSharp:
I understand your point. I like the idea of separating `IPythonBuffer` from `IBufferProtocol`, with `IPythonBuffer` representing `Py_buffer`. `MemoryView` could implement both interfaces. In the short term, the following would be helpful:

 * A common interface across `Bytes`, `ByteArray`, `MemoryView`, and `array.array` that allows access to `Memory`/`ReadOnlyMemory` and `Span`/`ReadOnlySpan`, directly or indirectly (e.g. through `IPythonBuffer`), ideally without copying data. It doesn't have to be `IBufferProtocol` but I think it suits the purpose. The actual design of this interface can be refined later. So let's assume `MemoryView` implements `IBufferProtocol`

 * If we go ahead with `ReadOnlyMemory` on `_codecs` decoding functions, a `MemoryView` constructor accepting `ReadOnlyMemory`, obviously without data copying.

 * If we go ahead with `ReadOnlyMemory` on `_codecs` decoding functions, a `Bytes` constructor accepting `ReadOnlyMemory`, ideally without data copying. Unfortunately, the current implementation exposes the backing array through `UnsafeByteArray`, which turns out to be quite popular (37 usages). So for now a copy will be necessary. Maybe over time we can wean the code off `UnsafeByteArray` and replace it with `ReadOnlyMemory`/`ReadOnlySpan` access.

Currently `bytes(bytes(1))` will copy the underlying array, despite it being read-only. It would be nice if the array could be shared. This would be beneficial if we adopt `IBufferProtocol` on `_codecs` decoding functions. If we go ahead with `ReadOnlyMemory`, the previous point will take care of it — over time, that is.

Src/IronPython/Runtime/Bytes.cs
```diff
@@ -60,7 +64,17 @@ public class Bytes : IList<byte>, IEquatable<Bytes>, ICodeFormattable, IExpressi
             _bytes = StringOps.encode(context, unicode, encoding, "strict").UnsafeByteArray;
         }
 
-        private static readonly Bytes[] oneByteBytes = Enumerable.Range(0, 256).Select(i => new Bytes(new byte[] { (byte)i })).ToArray();
+        private static readonly IReadOnlyList<Bytes> oneByteBytes;
+
+        static Bytes() {
+            var barr256 = new Bytes[256];
```
slozier: Not exactly sure what's gained by using this instead of the one-liner above?

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -3357,38 +3357,19 @@ public static partial class PythonOps {
             return DoubleOps.Compare(self, other);
         }
 
-        [Obsolete("Use Bytes(IList<byte>) instead.")]
+        [Obsolete("Use Bytes(IList<byte>) or Bytes(ReadOnlySpan<byte>) instead.")]
         public static Bytes MakeBytes(byte[] bytes) {
-            return new Bytes(bytes);
+            return new Bytes(bytes.AsSpan());
```
slozier: Why add `AsSpan`? I think this should probably have been `Bytes.Make` anyway.

Src/IronPython/Runtime/Bytes.cs
```diff
@@ -40,6 +40,10 @@ public class Bytes : IList<byte>, IEquatable<Bytes>, ICodeFormattable, IExpressi
             _bytes = bytes.ToArray();
         }
 
+        public Bytes(in ReadOnlySpan<byte> bytes) {
```
slozier: I think this is what's causing all the test failures. Maybe change it to internal?

Src/IronPython/Runtime/IBufferProtocol.cs
```diff
@@ -31,6 +32,8 @@ public interface IBufferProtocol {
 
         Bytes ToBytes(int start, int? end);
 
+        Span<byte> ToSpan(int start, int? end);
```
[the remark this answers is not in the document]
BCSharp: I agree. Spans are easy to be sliced anyway. It is a strange signature, I just copied it from `ToBytes` etc. assuming that the whole interface will be redesigned at some point. Besides, it seems to me that we will end up with `ToMemory`.

If we don't want to expose writable access to data, I would simply return `ReadOnlySpan`/`ReadOnlyMemory` and keep one method.

Src/IronPython/Modules/_bytesio.cs
```diff
@@ -95,7 +95,7 @@ public BytesIO(CodeContext/*!*/ context, object initial_bytes=null)
             public MemoryView getbuffer() {
                 _checkClosed();
 
-                return new MemoryView(new Bytes(_data), 0, _length, 1, "B", PythonOps.MakeTuple(_length));
+                return new MemoryView(new Bytes(_data.AsSpan()), 0, _length, 1, "B", PythonOps.MakeTuple(_length));
```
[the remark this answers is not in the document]
BCSharp: `Span`, `ReadOnlySpan`, `Memory`, `ReadOnlyMemory` define an implicit cast operator from an array to themselves. With the addition of constructor `Bytes(ReadOnlySpan<byte>)`, there is an ambiguity which constructor to use when the argument is `byte[]`: this one or `Bytes(IList<byte>)`? To resolve it, I tried various approaches (more on that below), but in this commit chose an explicit conversion to `Span`. A cast to `IList<byte>`  would work too, but it was more typing and more parentheses.

Src/IronPython.Modules/_codecs.cs
```diff
@@ -256,15 +258,18 @@ public static PythonTuple mbcs_encode(CodeContext/*!*/ context, [NotNull]string
         }
 
         public static PythonTuple raw_unicode_escape_decode(CodeContext/*!*/ context, [BytesConversion,NotNull]IList<byte> input, string? errors = null) {
+            var span = input.ToArray().AsSpan(); // TODO: remove when signature changed
+            var encoding = StringOps.CodecsInfo.Codecs["raw_unicode_escape"].Value;
```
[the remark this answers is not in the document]
BCSharp: I like your suggestion.

Src/IronPython/Runtime/Bytes.cs
BCSharp: In my attempts to resolve the ambiguity calling the constructor with a `byte[]` type argument, one of the things I tried was to make `Bytes(byte[])` public. Then neither cast to `IList` nor to `Span` would be necessary.

But this would mean that this constructor had to copy the input, so I tried to eliminate the need for a non-copying constructor. This static constructor is a left-over form that, which I kept around to give myself more freedom to experiment further, but at this point I am ready to give it up. In the next commit I will restore the one-liner.

Src/IronPython/Runtime/Operations/PythonOps.cs
BCSharp: Yes, this should have been `Bytes.Make` here. And maybe everywhere else where `byte[]` is being passes to `Bytes` constructor, instead of casting to `IList` or `Span`, use `Bytes.Make` or `PythonOps.MakeBytes`, since `Bytes.Make` is internal. I haven't done it because `PythonOps.MakeBytes` is marked obsolete and is being used in only one place, but maybe it can find a new purpose?

Src/IronPython/Runtime/Bytes.cs
BCSharp: You are exactly right, it is causing all those test failures. Interestingly, I had no test failures locally, but since then was able to reproduce them, just needed to run `./make clean`. I'm not quite sure I understand why the failures are happening (besides knowing that this constructor is the trigger), perhaps related to/the same as  IronLanguages/ironpython2#347 you mentioned.

Strictly speaking, I don't need this constructor to "Utilize new overloads from .NET 4.6", but looking ahead I thought we will need something like that to handle _bytes-like_, so this was a "trial balloon". Boy, it popped. The good news is that `ReadOnlyMemory` instead of `ReadOnlySpan` works OK.

Src/IronPython/Runtime/Operations/PythonOps.cs
slozier: I think I was the one to mark `PythonOps.MakeBytes` as obsolete since I didn't want it to be used to create `Bytes` objects. Not sure I really want to (for the moment) have a public method to create `Bytes` from an array.

BCSharp: By the way, do you know how to create a `bytes[]` instance in IronPython (that is `System.Array` with elements of `System.Byte`)?

Src/IronPython/Runtime/IBufferProtocol.cs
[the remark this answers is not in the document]
slozier: Let's make it read-only for now. We can revisit when we change `IBufferProtocol` or when a reasonable use case comes up.

Src/IronPython/Runtime/Bytes.cs
slozier: You can use `Bytes.Make` to avoid any ambiguity.

Src/IronPython/Runtime/Operations/PythonOps.cs
slozier: `System.Array[System.Byte]([1, 2, 3])`

BCSharp: I see. I was trying to create an empty (zero-filled) array of a given size:
```
>>> System.Array[System.Byte](1000)
TypeError: expected object with __len__ function, got int
```

slozier: Presumably we could add an `int` overload to `ArrayOps` to allow this?

BCSharp: I think it would be good thing to implement. I would bring .NET arrays on par with `bytes` and `bytearray`, which both accept an int to specify a length of a zero-filled array.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -3357,9 +3357,9 @@ public static partial class PythonOps {
             return DoubleOps.Compare(self, other);
         }
 
-        [Obsolete("Use Bytes(IList<byte>) or Bytes(ReadOnlySpan<byte>) instead.")]
+        [Obsolete("Use Bytes(IList<byte>) instead.")]
         public static Bytes MakeBytes(byte[] bytes) {
-            return new Bytes(bytes.AsSpan());
+            return Bytes.Make(bytes.ToArray());
```
BCSharp: I use `ToArray()` here to maintain the semantics of the method as it was before. Although the usage shows that a non-copying construction would be OK, I am uneasy of providing it here as it is a public method. What about calling `Bytes.Make` directly in the expression tree and deleting `PythonOps.MakeBytes`?

Src/IronPython/Runtime/Operations/PythonOps.cs
BCSharp: I've found it:
```
System.Array.CreateInstance(System.Byte, 1000)
```
Nevertheless, I will submit a PR to support a more Pythonic syntax as well.

Src/IronPython/Runtime/Operations/PythonOps.cs
slozier: I believe the expression tree is used when compiling to an assembly. For example, looking at the compiled code of `b"a"` using ILSpy, I see that it ends up as
```C#
PythonOps.MakeBytes(new byte[4] { (byte)97 })
```
I'm guessing the method needs to be public so that it can be called from external DLLs so that probably rules out using `Bytes.Make`.

It seems like in this case it would be beneficial to allow it to construct the Bytes object without an extra array copy. I'm not sure what the solution is... Maybe add a description to the method and `[EditorBrowsable(EditorBrowsableState.Never)]` and if people choose to disregard it then too bad.

BCSharp: OK, let's keep using it as a no-copy construction option and pile up any usage discouraging attributes on it.

I did some tests in trying to use the internal method, but the only version that worked was using reflection. Since constructing a `bytes` object from a constant is such a basic operation, I don't like the extra performance hit of the reflection in this case.

Src/IronPython.Modules/_codecs.cs
```diff
@@ -248,51 +250,58 @@ public static PythonTuple mbcs_encode(CodeContext/*!*/ context, [NotNull]string
         #endregion
 
         #region Raw Unicode Escape Encoding Functions
+#if FEATURE_ENCODING
```
slozier: Unrelated to this PR, but I think we can drop the whole `FEATURE_ENCODING` from the codebase. It was meant for platforms such as Silverlight but I think we can safely stop supporting those.

Src/IronPython/Runtime/ByteArray.cs
```diff
@@ -269,13 +269,19 @@ public ByteArray center(int width, [BytesLike, NotNull]IList<byte> fillchar)
 
         public string decode(CodeContext context, [NotNull]string encoding = "utf-8", [NotNull]string errors = "strict") {
             lock (this) {
-                return StringOps.RawDecode(context, _bytes, encoding, errors);
+                var mv = new MemoryView(this);
+                // TODO: Move the rest outside lock when wrapping bytearray in memoryview turns byttearray readonly
```
slozier: Maybe we should make `MemoryView` implement `IDisposable`?

Wrapping  `bytearray` in a `memoryview` does not make it readonly, it should only prevent resizing of the buffer.

Src/IronPython.Modules/_codecs.cs
BCSharp: There are more compilation [features defined](https://github.com/IronLanguages/ironpython3/blob/master/Documentation/feature-symbols.md) than just `FEATURE_ENCODING`. Maybe it is a long shot but maybe one day IronPython will be used in a (relatively) resource constrained environment, like a mobile platform, Raspberry Pi, or IoT device, and the ability to exclude all that encoding stuff will be handy. It is already in place.

The whole list of feature symbols will need to be reviewed, updated and documented at some point. For encodings, I just need to get in a habit to build with this feature disabled, to catch compiler errors. How about defining a property, say `Bare` that would produce a build with all features disabled? No testing, just compilation.

Src/IronPython/Modules/array.cs
```diff
@@ -127,12 +127,17 @@ public class array : IEnumerable, IWeakReferenceable, ICollection, ICodeFormatta
...
+                ExtendIter(initializer);
+            }
+
             public array([NotNull]string type, object? initializer) : this(type) {
-                if (_typeCode != 'u') {
-                    if (initializer is string || initializer is Extensible<string>)
-                        throw PythonOps.TypeError("cannot use a str to initialize an array with typecode '{0}'", _typeCode);
-                    if (initializer is array arr && arr._typeCode == 'u')
-                        throw PythonOps.TypeError("cannot use a unicode array to initialize an array with typecode '{0}'", _typeCode);
+                if (_typeCode != 'u' && (initializer is string || initializer is Extensible<string>)) {
```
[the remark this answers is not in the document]
BCSharp: I'm not sure I understand your comment. The following produces identical result in in both CPython and IronPython:

```
>>> array.array("b", array.array("u"))
TypeError: cannot use a unicode array to initialize an array with typecode 'b'
>>> array.array("b", "abc")
TypeError: cannot use a str to initialize an array with typecode 'b'
```

Do you mean the case when somebody casts `array` to `object` and calls the constructor from C#?

Src/IronPython/Runtime/Binding/ConversionBinder.cs
```diff
@@ -170,7 +170,15 @@ internal class PythonConversionBinder : DynamicMetaObjectBinder, IPythonSite, IE
 
                         // Interface conversion helpers...
                         if (genTo == typeof(IList<>)) {
-                            res = TryToGenericInterfaceConversion(self, type, typeof(IList<object>), typeof(ListGenericWrapper<>));
+                            if (type.GenericTypeArguments[0] == typeof(byte) && self.Value is IBufferProtocol) {
```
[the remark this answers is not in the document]
BCSharp: I don't want #762 be merged, but I also don't want it to be rejected yet. It is basically a backup for this PR. In the (seemingly unlikely) case that we do not go ahead with this PR, #762 can then be merged instead.

Src/IronPython/Runtime/Binding/PythonOverloadResolver.cs
```diff
@@ -69,12 +69,26 @@ public PythonOverloadResolver(PythonBinder binder, IList<DynamicMetaObject> args
             if ((fromType == typeof(PythonList) || fromType.IsSubclassOf(typeof(PythonList)))) {
                 if (toParameter.Type.IsGenericType &&
                     toParameter.Type.GetGenericTypeDefinition() == typeof(IList<>) &&
-                    toParameter.ParameterInfo.IsDefined(typeof(BytesLikeAttribute), false)) {
+                    (toParameter.ParameterInfo?.IsDefined(typeof(BytesLikeAttribute), inherit: false) ?? false)) {
```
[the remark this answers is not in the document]
BCSharp: Yes, I did, although I don't have the sample Python code anymore that triggered it. But it got my attention.

Anyway, if you look into ParameterWrapper definition, it [explicitly guards itself](https://github.com/IronLanguages/dlr/blob/4d72535ae3109ec077168138723c59c6f8e1493a/Src/Microsoft.Dynamic/Actions/Calls/ParameterWrapper.cs#L83) against null `ParameterInfo`. I suppose nullable annotations would have exposed it.

Src/IronPython/Runtime/ByteArray.cs
BCSharp: I like the `IDisposable` idea, although this may change then the `IBufferProtocol` will be refactored to `IPythonBuffer`, which will likely be `IDisposable`. Unless `MemoryView` implements `IPythonBuffer` as well...

As for the readonly issue, I will do some more experiments tomorrow.

Src/IronPython/Modules/array.cs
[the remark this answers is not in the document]
slozier: Yes, I meant if someone calls it via C# or uses the object overload specifically in IronPython (`array.array.__new__.Overloads[str, object](array.array, "b", array.array("u", "abc"))`). Although, I have no idea why anyone would ever try that. 😄

Src/IronPython/Runtime/Binding/PythonOverloadResolver.cs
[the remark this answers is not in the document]
slozier: I see, interesting that we never had a check for that before.

Src/IronPython/Runtime/ByteArray.cs
slozier: Even if it changes `IDisposable` probably makes sense for interop. Anyway, this change can probably be done in a separate PR since it's a side issue.

```Python
import array
x = array.array("b", range(5))
with memoryview(x) as mv:
    mv[0] = 20
    assert x[0] == 20
    x[0] = 30
    assert mv[0] == 30
    try:
        x.append(1) # this throws
        assert False
    except BufferError:
        pass
try:
    mv.readonly # this throws
    assert False
except ValueError:
    pass
```

Src/IronPython.Modules/_codecs.cs
slozier: Indeed, there are many feature symbols defined, most of them are there due to missing APIs in one platform or another. I had done a first pass at eliminating some feature symbols https://github.com/IronLanguages/ironpython2/issues/200 although I had need very tolerant. For ipy3 I would lean towards eliminating any feature switch for features part of .NET Standard 2.0 (e.g. FEATURE_DBNULL, FEATURE_APPLICATION_EXCEPTION, etc).

I guess `FEATURE_ENCODING` could be open to discussion. According to the above criterion it could be eliminated, however limiting to "base" encodings could be an option. We should however review what gets excluded by this symbol, for example it seems like ignoring `errors` is not a good idea? Anyway, probably something for another PR.

"Constrained environments" is indeed a bit of a long shot, I just ran the test suite on a Raspberry pi. 😄

Src/IronPython/Modules/array.cs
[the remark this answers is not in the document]
BCSharp: I see. OK, I will revert the object constructor overload to its previous form.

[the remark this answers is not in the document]
slozier: No need to revert, just saying it's something to think about. I guess the `IList<byte>` overload may have a similar issue anyway.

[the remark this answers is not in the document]
BCSharp: Food for thought:

```
>>> ua = array.array("u", "abc")
>>> array.array.__new__.Overloads[str, System.Collections.Generic.IList[System.Byte]](array.array, 'b', ua)
array('b', [97, 0, 98, 0, 99, 0])
>>> array.array.__new__.Overloads[str, object](array.array, "b", ua)
OverflowError: couldn't convert str to SByte
```
It is not possible to accidentally call the `IList<byte>` overload from C# so if somebody goes though all that trouble to call it from IronPython, presumably they know what they are doing and I would give them a free pass.

The `object` overload still throws but the exception is less informative and does not match CPython's. Also, it is easier to land on this overload from C#, although if the array reference is kept in a `dynamic` variable, the proper `array` overload will be used.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -457,7 +460,7 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
             }
             set {
                 CheckBuffer();
-                if (_buffer.ReadOnly) {
+                if (@readonly) {
```
BCSharp: Accessing the `readonly` property already does `CheckBuffer`, so here (and in a few cases below) the buffer is unnecessarily checked twice. I can either remove the call to `CheckBuffer` here (more consise) or replace the property access with expression `_readonlyView || _buffer.ReadOnly` (more flexibility in ordering the exception conditions, although it seems good enough now).

@slozier, any preference?

slozier: I would prefer to keep the `CheckBuffer` since it reduces the chance of the check being accidentally removed. Is there a reason why we couldn't do something like `_isReadOnly = readonlyView || _buffer.ReadOnly` in the constructor and drop the `_readonlyView` field? That way we could just do `if (_isReadOnly)` throughout.

BCSharp: It assumes that a buffer that at one time reported readonly will never become writable. I don't know if it is a valid assumption, but if you can confirm it is so, then yes, `_isReadOnly` would be the best option.

slozier: I think it's a reasonable assumption. We're the ones defining how the buffer objects should behave, if someone chooses to disregard the contract then too bad.

In the (CPython) buffer protocol you just ask an object for a buffer to its underlying memory. If you ask for a writable buffer then:
> the exporter MUST provide a writable buffer or else report failure. Otherwise, the exporter MAY provide either a read-only or writable buffer, but the choice MUST be consistent for all consumers.

https://docs.python.org/3/c-api/buffer.html#buffer-request-types

Src/IronPython.Modules/_codecs.cs
BCSharp: After thinking about it for a few days, I now agree that `FEATURE_ENCODING` as it is can better be dropped. Something for a separate PR.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1732,13 +1721,11 @@ public static class StringOps {
...
 
-        internal static string DoDecode(CodeContext context, IList<byte> s, string errors, string encoding, Encoding e)
-            => DoDecode(context, s, errors, encoding, e, s.Count, out _);
-
-        internal static string DoDecode(CodeContext context, IList<byte> s, string errors, string encoding, Encoding e, int numBytes, out int numConsumed) {
-            byte[] bytes = s as byte[] ?? ((s is Bytes b) ? b.UnsafeByteArray : s.ToArray());
-            int start = GetStartingOffset(e, bytes);
-            int length = numBytes - start;
+        internal static string DoDecode(CodeContext context, ReadOnlyMemory<byte> data, string errors, string encoding, Encoding e) {
+            int start = GetStartingOffset(e, data.Span);
```
slozier: Since `start` is only used for slicing maybe it could just be a helper (static local) method `RemovePreamble` or something?
```C#
data = RemovePreamble(e, data);
```

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1870,13 +1868,15 @@ internal static Bytes DoEncodeUtf8(CodeContext context, string s)
             throw PythonOps.TypeError("'{0}' encoder returned '{1}' instead of 'bytes'; use codecs.encode() to encode to arbitrary types", encoding, DynamicHelpers.GetPythonType(res[0]).Name);
         }
 
-        private static string UserDecode(PythonTuple codecInfo, IList<byte> data, string errors) {
+        private static string UserDecode(PythonTuple codecInfo, object data, string errors) {
```
slozier: Maybe `IBufferProtocol` instead of `object`? Unless you plan to call it from somewhere else.

Tests/test_bytes.py
```diff
@@ -26,6 +27,35 @@ def __index__(self):
...
+            self.assertEqual(testType(array.array(b)), b)
+
+    @unittest.skipUnless(is_cli, "Interop with CLI")
+    def test_init_interop(self):
+        import System
+        from System import Byte, Array, ArraySegment
+
+        arr = Array[Byte](b"abc")
+        ars = ArraySegment[Byte](arr)
+        if is_netcoreapp:
```
slozier: Could we do `clr.AddReference("System.Memory")` to get it working with `net46`?

Src/IronPython/Runtime/Bytes.cs
```diff
@@ -48,8 +52,16 @@ public class Bytes : IList<byte>, IReadOnlyList<byte>, IEquatable<Bytes>, ICodeF
             _bytes = new byte[size];
         }
 
-        private Bytes(byte[] bytes) {
-            _bytes = bytes;
+        public Bytes([NotNull]byte[] bytes) {
```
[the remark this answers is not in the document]
BCSharp: Yes, they are, but `ReadOnlyMemory` defines [implicit conversions from those types](https://docs.microsoft.com/en-us/dotnet/api/system.readonlymemory-1.op_implicit?view=netcore-3.1). So we do need these overloads to break the abiguity between `IList<byte>` and `ReadOnlyMemory<byte>` when calling from C#. Try to delete the `byte[]` overload and compile the project to see what I mean.

Src/IronPython/Runtime/Operations/StringOps.cs
BCSharp: I was thinking of merging `StringOps.RawDecode` and `_codecs.decode`: they both seem to do the same, yet they have a different implementation. Right now, `UserDecode` is only used through `RawDecode`, which accepts only `IBufferProtocol`, but `_codecs.decode` accepts `object?`. So I wanted to keep it this way until it becomes clear what the final solution will be. I haven't looked into this issue deep enough yet.

Src/IronPython/Runtime/Bytes.cs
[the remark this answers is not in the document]
slozier: Ugh, that is super unfortunate. It means that any time we have an `IList<byte>` and `ReadOnlyMemory<byte>` overload we need to add overloads for `byte[]`, `ArraySegment<byte>` and whatever other type implements `IList<byte>` with an implicit `ReadOnlyMemory<byte>` cast?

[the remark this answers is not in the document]
BCSharp: Well, not really, we don't need to. 

First of all, this affects only calls from C#. For the calls from Python, the overload resolver seems not to take implicit casts into consideration. That's why currently a call with a `Memory` argument fails to resolve, rather than being bound to the `ReadOnlyMemory` overload. Which is a good thing in this case, because it means than in the absence of the `byte[]` overload, a `byte[]` argument call will simply bind to the `IList<byte>` overload, rather than trigger the ambiguous match error. I wouldn't even know how to explicitly "cast" an argument to a specific type in a dynamic language. 

So those overloads are there only for the sake of C# calls. Without them, every call with, say, `byte[] arr`, would either have to use `((IList<byte>)arr)` or `arr.ToMemory()` instead of simply `arr` as argument. But it is till workable. However, in this particular case, I thought that `Bytes` is such a fundamental Python type, basically a Python equivalent of `byte[]`, and initializing it with `byte[]` is such a basic operation that an explicit overload of `Bytes` constructor is warranted. 

I feel much less strongly about the `ArraySegment` overload. It is there just for completeness. I know of no other .NET library type than these two that implement `IList<byte>` and at the same time have an implicit conversion to `ReadOnlyMemory<byte>`.  The IronPython codebase does not use `ArraySegment`, so this overload is there just as courtesy for user C# code. If you think it goes too far, I am ready to remove it. There is always some benefit of having fewer overloads to consider for the overload resolver.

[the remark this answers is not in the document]
BCSharp: There is yet another possibility: remove the two new overloads **together** with the `IList<byte>` and `IEnumerable<object?>` overloads. Then use type tests in the `object?` overload to perform the most efficient conversion to array.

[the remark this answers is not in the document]
slozier: Yeah, I though of that, although it's not very .NET-like. I'd prefer to reserve the object overload for Python objects.

Anyway, we can keep the `byte[]` overload if you think it's a good idea.

I would also be fine with having the ambiguous call issue. Just looking at where it fails, in `_bytesio.cs` is shouldn't even be making a copy, the underlying buffer should be exposed directly. I guess the other failures in the `IronPython.SQLite` project could just be solved via a cast.

BCSharp:
ByteArray.cs, lines 1174:1176:
```csharp
            // always copy from a List object, even if it's a copy of some user defined enumerator.  This
            // makes it easy to hold the lock for the duration fo the copy.
            IList<byte> other = GetBytes(value);
```
I can't figure out the purpose of this line.  The comments suggests that a copy is required, but a copy is already made in the caller, `object? this[[NotNull]Slice slice]`. Besides, `value` here is always `IList<byte>` hence `GetBytes` is simply returning the same reference as the argument received. As much in the latest version as well in the very first version checked into the ironpython3 git archive.

Shouldn't this line simply be removed? Or am I missing something?

slozier:
Indeed, seems safe to remove. Probably just a copy/paste of the `PythonList.SliceNoStep` which had an `object` argument.

BCSharp:
I consider this exercise as completed. Any other improvement discussed here can be further done on the mainline, after this PR is merged, assuming it is going to be merged.

BCSharp:
Thanks!

PR #771, opened 2020-03-22
Fix failing tests on .NET Core 3.1 on macOS

BCSharp:
Unfortunately, .NET Core 3.1 on macOS does not produce a standalone executable, so the shell script is still needed.

### DIFF
--- a/Src/IronPythonConsole/IronPythonConsole.csproj
+++ b/Src/IronPythonConsole/IronPythonConsole.csproj
@@ -9,7 +9,6 @@
     <CodeAnalysisRuleSet>..\..\IronPython.ruleset</CodeAnalysisRuleSet>
     <DocumentationFile>$(OutputPath)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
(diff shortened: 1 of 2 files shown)


Src/IronPythonConsole/IronPythonConsole.csproj
```diff
@@ -9,6 +9,7 @@
     <CodeAnalysisRuleSet>..\..\IronPython.ruleset</CodeAnalysisRuleSet>
     <DocumentationFile>$(OutputPath)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <MacOS Condition="'$(OS)' == 'Unix' AND Exists('/Applications') AND Exists('/Library') AND Exists('/System') AND Exists('/Volumes')">true</MacOS>
```
slozier: Does this work? I have no mac to try it.
```suggestion
    <MacOS Condition="$([MSBuild]::IsOsPlatform('OSX'))">true</MacOS>
```

BCSharp: It works! I suppose we don't need `MacOS` anymore. Neither here nor in `Build.proj`.

slozier: Looks like the one in `Build.proj` is used by packaging.

PR #773, opened 2020-03-24
Fix intermittent test_shutil failures on Mono on macOS

BCSharp:


Tests/test_shutil.py
```diff
@@ -22,6 +23,7 @@ def test_copyfileobj(self):
         pass
 
     def test_copyfile(self):
+        return
```
slozier: Forgot to remove these?

PR #777, opened 2020-04-02
Fix list assignment using extended slice with big integers

BCSharp:
Was:
```
>>> plist = [1, 2]
>>> plist[0:None:2<<222] = [0]
ValueError: too many items in the enumerator need 0 have 1
>>> plist[0:None:2<<222] = [111, 222]
ValueError: too many items in the enumerator need 0 have 2
```
Now:
```
>>> plist = [1, 2]
>>> plist[0:None:2<<222] = [0]
>>> plist
[0, 2]
>>> plist[0:None:2<<222] = [111, 222]
ValueError: too many items in the enumerator need 1 have 2
```
Which is the same as CPython, except that the message for `ValueError` reads `attempt to assign sequence of size 2 to extended slice of size 1`. Do we care about the error message?

The patched code seems quite old. The same bug exists in IronPython 2.

Similar issues are with assigning slices of `bytesarray` and `memoryview`, but I am reluctant of submitting any updates of code in that area until the status of #765 is resolved, to avoid merge conflicts.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -1185,6 +1185,8 @@ namespace IronPython.Runtime.Operations {
             int length, object start, object stop, object step,
             out int ostart, out int ostop, out int ostep
         ) {
+            Debug.Assert(length >= 0);
+
             if (step == null) {
                 ostep = 1;
             } else {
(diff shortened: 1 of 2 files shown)


BCSharp:
I think you've made a good point. I added the assertion when I realized that the logic of `FixSlice` breaks down for negative length values beyond something that can be handled with a simple `Math.Min` and `Math.Max`, and negative length values are not a valid use case to justify extra work or complexity to support it. So it was meant as a remainder to the reader what the assumptions of these methods are. 

I completely overlooked the fact that they are public. In this case, `ValueError` would be more appropriate, but also both methods could be made internal instead of public. If you are planning to go through the whole `PythonOps` to reduce the public API then I'll leave it as is for now.

PR #778, opened 2020-04-03
Fix sequence assignments using extended slice with big integers

BCSharp:
This is the continuation of #777. Hopefully I have fixed all places. I see a lot of similar, but ultimately tricky calculations being done in different places, so I propose to move `GetSliceLength` and `CoalesceToInt32` from `MemoryView` to `PythonOps` and let the rest of the code to use those functions.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -1150,7 +1150,7 @@ namespace IronPython.Runtime.Operations {
             }
         }
 
-        public static int ClampToInt32(this long value) {
+        internal static int ClampToInt32(this long value) {
             return (int)Math.Max(Math.Min(value, int.MaxValue), int.MinValue);
         }
 
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -1150,6 +1150,10 @@ namespace IronPython.Runtime.Operations {
             }
         }
 
+        public static int ClampToInt32(this long value) {
+            return (int)Math.Max(Math.Min(value, int.MaxValue), int.MinValue);
+        }
+
         public static bool IsMappingType(CodeContext/*!*/ context, object o) {
             if (o is IDictionary || o is PythonDictionary || o is IDictionary<object, object>) {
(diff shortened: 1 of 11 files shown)


BCSharp:
...and perhaps rename `GetSliceLength` to `GetSliceCount` to keep with the convention that "count" is used with enumerables and sequences, while "length" is used more with contiguous types like arrays, memory, and span.

slozier:
I agree a lot of this could be consolidated. It seems like in most cases this is used following a `Slice.indices`. What about a `Slice.GetIndicesAndCount` method that would give you everything at once?

As for naming, I agree with the use of "count". As for `CoalesceToInt32` I wasn't quite sure what that meant until I looked at the code. Maybe `ClampToInt32` would be a better name?

BCSharp:
I wouldn't know what `ClampToInt32` meant unless I looked at the code 😉 . How about `ClipToInt32`? Or `ReduceToInt32`?

slozier:
Interesting, I thought clamp was common terminology. Maybe it's more common in math? I was proposing it since it's basically doing `Math.Clamp(value, int.MinValue, int.MaxValue)`. Another option would be `RoundToInt32`? Anyway, I guess I don't have a strong preference. :smile:

BCSharp:
Interesting indeed. I've never seen `Math.Clamp` before. In signal processing, what it does would be called clipping. Clamping would be more like going from signed to unsigned. But I don't have a strong preference either, and since `Math.Clamp` has set the precedent, I'm okay with `ClampToInt32`. BTW, I foresee a need for an overload acting on `BigInteger`. I think extension methods would be most convenient. 

Looking at the bigger picture, I would like to discuss the overall handling of `int` in IronPython3, but best done under a separate issue.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1150,6 +1150,10 @@ public static partial class PythonOps {
             }
         }
 
+        public static int ClampToInt32(this long value) {
```
slozier, line 1153: Any specific reason to have this one public?

BCSharp: No. Was meant to be `internal`.

PR #782, opened 2020-04-10
Fix find/rfind

BCSharp:
It was tempting to switch from `[BytesLike]IList<byte>` to `[BytesLike]ReadOnlyMemory<byte>`. A number of things could have been simplified.

PR #784, opened 2020-04-12
Patch test_unicode from StdLib to work with IronPython

BCSharp:


BCSharp:
It cannot be enabled as a whole yet, but I have a partial enable, which I am about to commit under #782 once this one has landed on master.

PR #786, opened 2020-04-14
Remedy intermittent failures of test__ssl

BCSharp:
Hopefully it will reduce a few false negatives from CI tests.


slozier:
Thanks!

PR #789, opened 2020-04-17
Add nullable annotations to builtins

slozier:


BCSharp:
Great! On my end, I have prepared a PR with nullable annotations in `StringOps`, which has exactly the same changes in `PythonCalls` as here. I will wait with the submission until this PR is merged and then rebase on top.

Issue #790, opened 2020-04-18
Review PythonOps.TryGetBoundAttr usage


slozier:
In a lot of cases, `PythonOps.TryGetBoundAttr` is used when it should not be. While the usages may have been correct for IronPython 2, it is probably no longer the case.

As an example, `math.trunc` calls:
```C#
if (PythonOps.TryGetBoundAttr(value, "__trunc__", out object func))
    return PythonOps.CallWithContext(context, func);
```
it should probably be:
```C#
if (PythonTypeOps.TryInvokeUnaryOperator(context, value, "__trunc__", out object res))
    return res;
```
The difference being that `TryGetBoundAttr` will resolve on the instance, but in this case `__trunc__` should resolve on the type.

PR #798, opened 2020-04-21
Support __bytes__ special method for bytes initialization

BCSharp:
Makes the following pass:
```python
class A:
    def __bytes__(self):
        return b'abc'

assert bytes(A()) == b'abc'
```

The following still does not work:

```python
class BytesSubclass(bytes): pass

class A5:
    def __bytes__(self):
        return BytesSubclass(b'abc')

assert bytes(A5()) == b'abc'  # passes
assert type(bytes(A5())) is BytesSubclass # fails
```


Src/IronPython/Runtime/Bytes.cs
```diff
@@ -30,7 +30,18 @@ public class Bytes : IList<byte>, IReadOnlyList<byte>, IEquatable<Bytes>, ICodeF
             _bytes = new byte[0];
         }
 
-        public Bytes(object? source) : this(ByteArray.GetBytes(source, useHint: true)) { }
+        public Bytes(CodeContext context, object? source) {
+            if (PythonOps.TryGetBoundAttr(context, source, "__bytes__", out object? func)) {
```
slozier, line 34: This should probably use `PythonTypeOps.TryInvokeUnaryMember`. See https://github.com/IronLanguages/ironpython3/issues/790. Otherwise the following (minus any typos) would work when it shouldn't.
```Python
class test: pass
t = test()
t.__bytes__ = lambda x: b"1"
bytes(t)
```

BCSharp: Hmm, I thought it should, but it shouldn't indeed. Funny, it still doesn't work, and with a strange error message:
````
TypeError: <lambda$175>() takes exactly 1 argument (1 given)
````

slozier: That would be a typo on my end (writing code on a phone usually doesn't work out well), should have been `t.__bytes__ = lambda: b"1"`.

BCSharp: But shoudn't the message read `<lambda$175>() takes exactly 1 argument (0 given)`?

slozier: Yes, the message is indeed messed up. I think it has to do with it counting "self" as an argument? Not the first time I've seen messed up argument counts but never bothered to investigate.

BCSharp:
I can't figure out how to support the case of returning a subclass from a constructor without changing a half of the .NET constructors to `__new__` constructors. 

Whatever the solution is, it will be for another PR, so I'm setting this one as ready for merging.

PR #801, opened 2020-04-22
Implement pickling protocol in PythonBytesIterator

BCSharp:
I suppose that `PythonBytesIterator` was not working with the default pickle handling because it does not have a public copy constructor and is not accessible as member of `__builtins__`.

This PR makes `bytes_iterator` behave more like CPython's.


Src/IronPython/Runtime/PythonBytesIterator.cs
```diff
@@ -24,6 +25,26 @@ public sealed class PythonBytesIterator : IEnumerable, IEnumerator<int> {
             _index = -1;
         }
 
+        #region Pickling Protocol
+
+        public PythonTuple __reduce__(CodeContext context) {
+            object? iter = context.GetBuiltinsDict()?[nameof(Builtin.iter)] ?? throw PythonOps.KeyError("'__builtins__'");
```
slozier, line 31: Any specific reason to want to use this pattern instead of `context.TryLookupBuiltin` used in other `__reduce__` methods? I have no idea if using `TryLookupBuiltin` is actually better...

Otherwise, my only issue with your approach is the following crazy example:
```Python
import builtins
__builtins__ = {"iter": None}
builtins.iter(b"").__reduce__()
```

BCSharp: Frankly, it did not occur to me to look at other `__reduce__` methods...

Playing with your suggestion now I see that using `context.TryLookupBuiltin` would make your "crazy" example to pass, producing `(None, (b'',), 0)`. However, CPython fails with `KeyError: '__import__'`. Currently IronPython fails with `KeyError: '__builtins__'`.

Here is another crazy example that would work OK in both implementations, but with different results:

```python
import builtins
__builtins__ = {"iter": None, "__import__": builtins.__import__}
rv = builtins.iter(b"").__reduce__()
print(rv)
```
CPython and current IronPython produces: `(<built-in function iter>, (b'',), 0)`
With `context.TryLookupBuiltin`, IronPython would produce: `(None, (b'',), 0)`. 

I was trying to mimic CPythons behaviour as much as possible, though it is still not exactly the same in some of those twilight cases. I'm guessing that CPython tries to import `__builtins__` first, then accesses `iter` through `getattr` on module. `GetBuildinsDict` fast-tracks directly to the dictionary, so if `iter` is not there there will be `KeyError`, while CPython will produce `AttributeError`:
```python
red = getattr(iter(b""), "__reduce__")
del __builtins__.__dict__["iter"]
print(red())
```
CPython: `AttributeError: 'module' object has no attribute 'iter'`
IronPython: `KeyError: iter`

I could use `TryGetValue` ISO indexed access and then throw the same error type, but I thought it was not worth the fuss. Such obscure scenarios, when somebody messes up builtins and then tries to reduce an iterator, can hardly be called a legitimate language use. I think it is more important to fail early and expose the bug.

FYI: I was also playing around with using `context.LanguageContext.BuiltinModuleDict` but found that `GetBuildinsDict()` got me closer to CPython's behaviour.

slozier: Interesting, it looks like CPython changed its behavior with 3.8 (which is the version I was playing with). With that version it produces `None`. Anyway, my goal was actually to seek consistency in our approach and trying to determine if your method was better.

Feel free to merge the PR as-is.

BCSharp: It also think that consistency is important. When I go through the code, I am often puzzled when seemingly the same operation is done in different ways. I find myself wondering: am I missing something here?

If other `__reduce__` methods follow what CPython 3.8 does, it is better to round up than down. I will change this PR to match it.

BCSharp: It looks like CPython changed its behaviour again in 3.8.2 (which I the version I have installed locally). With this version, the "crazy example" produces `<built-in function iter>`. Actually, this looks to me like the most sensible result. I wouldn't be surprised if the change was to some bug that got fixed.

I've found a way to get exactly the same behaviour as CPython 3.8.2 (if we disregard that some error messages may be slightly different) on all crazy examples I could think of.  This will be the code I will submit today — with a comment explaining why is has this form and not like other `__reduce__` methods. I leave it up to you to deal with other `__reduce__` cases (or this one) if you wish.

slozier: Hmm, interesting, I just upgraded and get `None`...
```
Python 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)] on win32
>>> import builtins
>>> __builtins__ = {"iter": None}
>>> builtins.iter(b"").__reduce__()
(None, (b'',), 0)
```

Anyway, I'll merge your updated version, there are more critical things to worry about. :smile:

BCSharp: I agree this is not important, but still odd...

I could reproduce your results when I used interactive prompt. But when the code is put in a `.py` file and run from command line, I get `<built-in function iter>`...

Issue #21, opened 2014-03-21
PEP 3118 -- Revising the buffer protocol


jdhardy:
http://www.python.org/dev/peps/pep-3118/

Issue #52, opened 2014-04-05
PEP 237 -- int/long Unification


jdhardy:
Python 3 gets rid of the distinction between int and long, with int now behaving as long. One way would be to switch everything to BigInteger, but it might be more efficient to try and use Int32 as much as possible behind the scenes, and switch to BigInteger if we have to while keeping the visible type int at all times.

Probably worth doing some benchmarks to see if it's worthwhile first.

markusschaber:
On modern 64 bit machines, we could use long (64 bit) instead of int for "simple" operations. (This is also what Python 2 did on 64 bit machines, as I just verified on an older ubuntu installation:
```
>>> sys.version
'2.7.17 (default, Sep 30 2020, 13:38:04) \n[GCC 7.5.0]'
>>> sys.maxint
9223372036854775807
```

BCSharp:
To get a feeling about performance of various options across frameworks, I have done some measurements of a few arithmetic operations executed in a long loop. Results of all operations fit in 32 bits so it is possible to compare to native `System.Int32`.

### .NET Performance Results

Here are best times for various types used for the variables in the loop, for both checked and unchecked code, on AMD Ryzen Threadripper 3960X.

<details>
<summary>Test loop template in C#</summary>

```csharp
    public void CheckedIntTest()
    {
        int n = int.MaxValue;
        int z = 0;
        checked
        {
            for (int i = 0; i < n; i++)
            {
                if (i % 2 == 0)
                {
                    z += i;
                }
                else
                {
                    z -= i;
                }
            }
        }
    }
```
</details>

|              Method    | .NET Fmwk | .NET 6.0 |
|----------------------- |----------:|---------:|
|      CheckedIntTest    |    1.71 s |   1.22 s |
|    UncheckedIntTest    |    1.24 s |   1.06 s |
|     CheckedLongTest    |    1.71 s |   1.22 s |
|   UncheckedLongTest    |    1.23 s |   1.05 s |
|      CheckedBoxTest    |   12.89 s |  12.50 s |
|    UncheckedBoxTest    |   13.08 s |  12.74 s |
|   CheckedBigIntTest[1] |  179.35 s |  62.35 s |
| UncheckedBigIntTest    |  179.55 s |  59.17 s |
|      CheckedGmpTest[2] |   95.81 s |    N/A   |
|    UncheckedGmpTest    |   94.14 s |    N/A   |

Comments:

[...]

PR #808, opened 2020-04-27
Factor out IPythonBuffer from IBufferProtocol

BCSharp:
Here is my exploration around the re-implementation of the buffer protocol (#21). I will try to make incremental changes that will keep IronPython working (so all tests should be passing).

Each step is a proposal of an idea that can be discussed and adjusted, so smaller steps are preferable, rather a large dive that goes nowhere.

The PR is writable to the maintainers, so I welcome commits on top in a collaborative effort.


BCSharp:
In the first commit I separated the buffer API from the buffer exporter API. This is how [PEP 3118](https://www.python.org/dev/peps/pep-3118/) handles it too.  The difference is how the buffer is being released. PEP 3118 [argues against an idea of a separate releaser object](https://www.python.org/dev/peps/pep-3118/#issues-and-details) (that is, different than the exported object), and that argument makes sense for C, but with OO C#, the `Dispose` pattern makes more sense. We have [discussed it before](https://github.com/IronLanguages/ironpython3/pull/765#issuecomment-598167466). For the record, here are some arguments for the `IDisposable` patterns.
1. It doesn't require for the consumer to hold on to the reference of the exporter. The buffer keeps that reference and GC will not collect it until the buffer is collected.
2. It fits the prevailing .NET usage pattern, which has some dedicated support from the C# compiler (`using`).
3. It is fairly simple to go from one form to the other in the interface adaptation layer, if a 3-rd party library expect a separate `PyBuffer_Release()`.

The changes here deliberately stop short from changing the existing behaviour in any big way. In particular, I think that several methods in `IPythonBuffer` should be removed. Also, generally speaking, exporters should not implement `IPythonBuffer` themselves, but spawn a separate object representing the buffer (with the exceptions of `MemoryView`, and maybe `CData`). This seems fairly straightforward to implement, but there is one pint that I do not yey how to handle: the automatic conversion of bytes-like objects to `IList` or `ReadOnlyMemory` in the Binder.

In CPython, a bytes-like handling function accepts the exporter object as argument (i.e. an object that has `getbufferproc` and `releasebufferproc`), and then requests and releases the buffer as needed. This corresponds to having bytes-like parameters typed `IBufferProtocol` (Option D). However, we opted for more convenient types `IList` and `ReadOnlyMemory`, with the conversion implemented in `PythonConversionBinder`. The issue is, that although the converted can easily call `GetBuffer` on `IBufferProtocol` to extract a compatible data access type, there no obvious way to call `Dispose`. It should be called after the method call returns, but the conversion expression tree does not know when it happens. Maybe there is a sort of de-convert mechanism available in the DLR that I am not aware of. Or maybe there is an option to provide conversion on the whole method level.

A workaround could be to release the buffer in the conversion ET right after the desired data object has been obtained, but this is not as the buffer protocol should work. The data object will be used after the buffer is released. It will work sort-of OK with IronPython built-in types that implement IBufferProtocol, but could be disastrous for unmanaged/unsafe 3-rd party libraries like NumPy.

slozier:
This is a good start. Hopefully I can find some time to experiment with this. Here are some of my initial thoughts:

> In particular, I think that several methods in IPythonBuffer should be removed.

Agreed, most of them can probably go.

> However, we opted for more convenient types IList and ReadOnlyMemory, with the conversion implemented in PythonConversionBinder.

Well, maybe we should revisit this. While these types are more from a .NET point of view it seems like they're not a great fit for the Python world. In particular `IList` is terrible because it would force us to make a copy the buffer? Another point which makes using the convenient types moot is that a lot of these methods are already a  annoying to invoke from .NET due to the `CodeContext` argument. Perhaps we should just stick to `IBufferProtocol` overloads. While we'd need to call `GetBuffer` every time it doesn't seem like such a big deal and we'd have more controls over the flags (otherwise we'd need some sort of attribute to specify them?).

We could have `IBufferProtocol` wrappers (for example `static IBufferProtocol ToBufferProtocol(this Array[byte] bytes)`) for the convenient .NET types. When invoking from the Python side of things the binder could easily use those wrappers to automatically convert the object to an `IBufferProtocol`.

BCSharp:
> Well, maybe we should revisit this. While these types are more from a .NET point of view it seems like they're not a great fit for the Python world. [...]  Perhaps we should just stick to `IBufferProtocol` overloads. 

We can revisit it, but I wouldn't dismiss the existing mechanism as yet. First of all, nothing needs to change to support `IBufferProtocol` as bytes-like input; it can already be used and there is no need for any special parameter attributes. So having support for `[BytesLike]` in the converter is a bonus, that may be useful or convenient in some cases (assuming that it works correctly). On the other hand, if the "case for those cases" is weak, then it does not warrant the extra complexity. I will try to sum up some pros and cons of both approaches, as I see it.

Using `IBufferProtocol` in the method signature:
1. It matches how CPython and offers the most flexibility on when, how (i.e. flags), and for how long a buffer is exported, unlike with the Binder mechanism. With the Binder mechanism, the buffer will be always requested at the entry to the method and released at the exit. This may not always match CPython. For instance, it is conceivable than for some corner cases created by values of other arguments, CPython never does request the buffer but fast-tracks to a default answer. If so, buffer error handling behaviour would be different. (Though I think this is an unlikely scenario).
2. It does not require any special handling code in the Binder, or any parameter attribute.
3. It makes it more difficult for interop, but in many cases methods that accept a bytes-like object are there just to provide an existing .NET functionality to Python. So there is little benefit for C# to call them. And indeed, we could provide a wrapper around `byte[]` or `Memory<byte>` for convenience.

Using `[BytesLike]` support in the Binder.
1. The change has less impact on the codebase, so the migration can be gradual. The common `GetBuffer` functionality is factored out to the Binder-generated code. The actual method code is cleaner.
2. If we settle on `Memory<byte>`/``ReadOnlyMemory<byte>`, interop is easier.
3. For advanced scenarios, a method can still use `IBufferProtocol` in its signature.

> In particular `IList` is terrible because it would force us to make a copy the buffer?

I agree that `IList` is not the best match for the buffer semantics. I would like to phase it out and replace with either `[BytesLike]ReadOnlyMemory` or `IBufferProtocol`, depending on the choice. Keeping support for `IList` as a possibility for buffer access will either require some exporters to make a copy of the data or we will need a wrapper around `Memory` that provides `IList`. The latter is functionally possible but will not offer efficient element access. One way or another, `IList` always underperforms, or is at best equal to other alternatives. It does offer an ability to pass discontinuous data, but as [you have indicated before](https://github.com/IronLanguages/ironpython3/pull/762#issuecomment-595947166), this is not a requirement of bytes-like objects, [which are expected to have contiguous byte data](https://docs.python.org/3/glossary.html#term-bytes-like-object).

So now how to phase it out. Since the full progression of types is `IBubberProtocol` -> `IPythonBuffer` -> `ReadOnlyMemory` -> `ReadOnlySpan` -> individual `byte`, I think the best first step is to replace `[BytesLike]IList<byte>` with `[BytesLike]ReadOnlyMemory<byte>`. These changes will be useful regardless the final mechanism of handling of bytes-like objects.

slozier:
>  I think the best first step is to replace [BytesLike]IList<byte> with [BytesLike]ReadOnlyMemory<byte>

This seems like a good first step, however, before jumping into it, I'd like to make sure this actually works. In particular, I'm trying to see how we would get from, for example, an `array('i')` (which is backed by an `Array<int>`) to a `Memory<byte>`. Right now we can write:
```C#
array a = new array("i", new byte[] {1,2,3,4,5});
using IPythonBuffer buffer = a.GetBuffer();
ReadOnlyMemory<byte> mem = buffer.ToMemory();
```
The issues which we would need to resolve are:
1. `buffer.ToMemory()` copies the data, it should be able to provide direct access to the memory.
2. `a` is backed by an `ArrayData<int>` which in turn is backed by an `int[]`. I know we can go from the array to a `Memory<int>`, but how do we go from there to a `Memory<byte>`? There's `MemoryMarshal.Cast` for `Span` but I don't see something like that for `Memory`.
3. `ToMemory` returns a `ReadOnlyMemory`, what do we do when we want write access?

I guess in all the `[BytesLike]` cases we only need read-only access, however, we need to support the array scenario where the data is not backed by a byte array.
```Python
bytes(array.array('i', [1, 2, 3, 4, 5]))
```

Any ideas?

BCSharp:
Here are my ideas (none of them tested out, though).

First of all, although `buffer.ToMemory()` copies data if `buffer` comes from an `array` so does converting `buffer` to `IList<byte>`. Eliminating `IList<byte>` use would allow to remove `IPythonBuffer.ToBytes()`. It is not a small task because `IList<byte>` is used in many methods of `Bytes` and `ByteArray`, but it will be needed regardless which final solution we settle on.

As for the `IPythonBuffer` interface, I wanted to remove `GetItem`, `SetItem`, `SetSlice`, `ToBytes`, and `ToList`. If we need `ToBytes` and `ToList` they can come as extension methods.

Yes, it would be easier with spans, but we can't use spans in public methods.

So I was thinking about replacing `ToMemory()`, which returns `ReadOnlyMemory<byte>` and possibly copying data, with two methods: `Memory<T>? AsMemory<T>()` and `ReadOnlyMemory<T>? AsReadOnlyMemory<T>()`. Neither of them would copy data. The former would only work for writable buffers (that is, `! IPythonBuffer.IsReadOnly`), throwing `InvalidOperationException` otherwise. The latter would work all the time. But: if `T` does not "match" `IPythonBuffer.Format`, `null` is returned.

An alternative: forget the generics and always return `Memory<byte>`/`ReadOnlyMemory<byte>`. The client then has to do proper data marshalling by dropping to `Span` and using `MemoryMarshal.Cast` if needed. This option is currently my favourite.

This does not solve the `ArrayData<int>` problem, which should be able to provide both `Memory<int>`, when `BufferFlags.Format` is used, and `Memory<byte>`, for `BufferFlags.Simple` requests. Maybe a solution is in changing the backing object from `int[]` to some unmanaged blob?

Yet another idea: use `IBufferProtocol` as the designated bytes-like type, and then each method does `GetBuffer(BufferFlags.Format)` -> `AsMemory<T>` -> `AsSpan<T>` -> `AsBytes<T>`. This could be an utility method in `PythonOps`, doing the job based on `IPythonBuffer.Format` Not ideal though, because CPython's `array` can satisfy simple (i.e. byte-typed) buffer requests too, so ideally the problem should be addressed in `array`.

BCSharp:
Here is another idea that I have tested out a little bit. It addresses the array problem locally, in `ArrayData<T>`. Instead of using `T[]` as the backing object, it can use `byte[]`. Providing `Memory<byte>` for `IPythonBuffer` would become trivial. Indexed access, on the other hand, would go something like this:
```C#
public T this[int index] {
    get => MemoryMarshal.Cast<byte, T>(_items.Span)[index];
    set => MemoryMarshal.Cast<byte, T>(_items.Span)[index] = value;
}
```
I would like to see it tied out. @slozier, you have been working on `ArrayData<T>` on the past, do you prefer to take it on?

slozier:
> As for the IPythonBuffer interface, I wanted to remove GetItem, SetItem, SetSlice, ToBytes, and ToList. If we need ToBytes and ToList they can come as extension methods.

Agreed, these are the methods I would also remove. I'm not sure that they are even needed as extension methods as their (only?) consumer is memoryview.

> Maybe a solution is in changing the backing object from int[] to some unmanaged blob?

I've thought about that as well, but while it solves the issue with `array`, it does not really solve the larger issue. For instance, I would really like to be able to use .NET arrays anywhere Python arrays are accepted (so with any bytes-like call site). We'd basically be running into the same issue.

> Yet another idea: use IBufferProtocol as the designated bytes-like type, and then each method does GetBuffer(BufferFlags.Format) -> AsMemory<T> -> AsSpan<T> -> AsBytes<T>

This is the idea I'm currently leaning towards. In fact, if we do this then I don't see why we couldn't just get rid of `Memory` and just have an `AsSpan<byte>` on `IPythonBuffer`?

BCSharp:
> I don't see why we couldn't just get rid of Memory and just have an `AsSpan<byte>` on `IPythonBuffer`?

I remember that having span as a method parameter of a public method was not working well. I never fully understood the root cause of this bug so I was avoiding spans anywhere public. But depending on the nature of the bug, maybe it is OK to have `Span<byte>` as return type of a method of a public interface? It will not be involved in the overload resolution process.

I am currently testing it.

slozier:
> I remember that having span as a method parameter of a public method was not working well.

Part of the issue with `Span` is that does not work well is that it can't be boxed or used as a generic argument so it causes a bunch of issues with call sites and the interpreter (which uses a stack of objects). If we're just using it "behind the scenes" via C# then there should be no issue with it.

BCSharp:
This was originally "Remove IPythonBuffer.SetItem(...)" but things turned out to be quite interdependent. I tried to keep changes to minimum, just to the level of all tests passing, and commit them to get early feedback on the direction it is going in. The code is not proofread or cleaned up, and is sprinkled with TODO comments.

There are some additional tests that are passing now (not included) but to my knowledge there are no regressions. 

It is strange that the build fails; does CI do something that a local `./make clean; ./make` doesn't? I've successfully tested the commit locally on Windows and macOS.

BCSharp:
I will try to merge in master and see if it helps.

slozier:
I like it. I added an `ArrayDataView` to test out the approach. I put the logic there instead of in `ByteArray` so that it could be reused with `array` but I'm not sure if it's worth it? I don't think my `ArrayDataView` is quite correct since it should probably throw whenever `Count` changes instead of when the array is replaced. Also there might be threading issues with just using a counter as I do?

BCSharp:
@slozier, thanks for the commit. I started working on `ByteArrayView`, leaving factoring out commonalities between it and `ArrayView` for later, but looking at your solution I see it can be done right away. Nevertheless, it was a worthwhile exercise because it got me thinking about multithreading early on. After some deliberation I came to a conclusion that it will be best not to make the Buffer Protocol views thread-safe, leaving any synchronization to the client (CPython also does not give any guarantees on thread-safety of buffer data). In the last commit, I wrote some rules of engagement as I see them currently.

 However, obtaining and releasing (and in the case of `ByteArray`, checking for exports) has to be thread-safe, as the rest of API of the exporting object. I added the modification to the code to ensure that. Perhaps they all can be moved to `ArrayData` for symmetry. It also means that all `CheckBuffer` calls have to happen in a locked critical section. This can be better done at `ByteArray` level. And yes, every method of `ByteArray` that may modify the length of the data needs to to the check.

Further, I am not sure about calling `Dispose()` from the finalizer. I vaguely recall a possibility that a referenced object from a field may already been garbage-collected, in this case, `_arrayData`. That would lead to an exception, with GC would probably ignore...?

Handling unformatted buffers is not quite clear to me. I find Python documentation on this point incomplete and ambiguous. Maybe I will have to look into CPython's code in the end to get the full picture. But one thing that is clear is that exporters, when given requests without `Formatted` flag, **MUST** provide buffers that report `Format` as `null`. See `BytesView`.

slozier:
I wonder if there's any value in making the `ByteArray._bytes` field readonly. We could do this be adding an `ArrayData<T>.Capacity` property (similar to `List<T>.Capacity`) to allow shrinking the underlying array.

Personally I don't see the issue with calling `Dispose` from the finalizer. It's essentially what the [dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implementing-the-dispose-pattern-for-a-base-class) says to do? Although I'm not an expert in this domain.

BCSharp:
> I wonder if there's any value in making the `ByteArray._bytes` field readonly.

I like the sound of it. This would place all export/resize logic and locking in one class. Feel free to experiment with it and commit your results. My plan for the next steps is to:
1. Provide conversion wrappers for `Memory`/`ReadOnlyMemory` as `IBufferProtocol`.
1. Provide automatic conversion from `Memory` et al. to `IBufferProtocol` by the binder.
1. Change all signatures using `[BytesLike]ReadOnlyMemory<byte>` to `IBufferProtocol` (that would be mostly in codecs).
1. Eliminate `ToMemory()` from `IPythonBuffer`.
1. Let `array.array` use `ArrayDataView`.

After that, I could go back to try to make `BytesArray._bytes` field readonly.

BCSharp:
> Personally I don't see the issue with calling `Dispose` from the finalizer.

Calling `Dispose` from the finalizer is not a problem. Disposing **managed** resources during such call is the problem.

The task of the Garbage Collector is to collect and recycle any resources that are no longer in use. GC does a pretty good job for managed resources because, well it manages them. But it cannot do it for unmanaged resources, by definition. That's where the finalizer comes in handy. Its job is to assist GC in freeing unmanaged resources.

A class can also have managed resources in use. The `IDispose` is meant to allow the programmer to release them as soon as they are not needed, rather than waiting on GC to collect them. Plus any unmanaged resources as well (say, close a file, so that it can be reopened).

So in a general case, we have the following:
1. The finalizer has to release unmanaged resources
2. `IDispose.Dispose()` has to release unmanaged as well as managed resources.

There is an overlap between them so it makes sense to factor it out to a separate method, conventionally `void Dispose(bool disposing)`. When this method is called with `true` argument, it means it is being called from `IDispose.Dispose()` and should release both managed and unmanaged resources. After such call, the finalizer is not needed anymore, because all unmanaged resources are already taken care of.
If it is called with `false` argument, it means the call comes from the finalizer. In such case, it has to dispose the unmanaged resources **only**.
If the class can be subclassed, then `Dispose(bool)` is made protected virtual, so that the subclasses can properly handle their managed and unmanaged resources, but in the same manner, that is, only unmanaged resources are released on calls with `disposing == false`.

This all together is the dispose pattern. Since `ArrayDataView` does not have any unmanaged resources, it does not need any finalizer. If the full dispose pattern were implemented in `ArrayDataView`, the call from the finalizer would be a no-op, and the type would only incur generation penalty for having a finalizer.

But all this is theory, the primary reason why I deleted the finalizer in `ArrayDataView` is because it could turn ordinary bugs into heisenbugs. For a class that does not own any unmanaged resources (hence does not have a finalizer), it may be occasionally OK not to call `Dispose()` on it after use (although IMO it is a poor programming practice). In the case of `ArrayDataView` the consequence is keeping its size permanently frozen. But in most cases this is a coding error. This will eventually pop up as an exception down the road when an attempt is made to resize it. If the finalizer calls `Dispose()` behind the scenes at unpredictable moments, such bug may become difficult to reproduce, depending on anything that influences the demand on GC to collect data.

In fact, I have recently had to deal with exactly this scenario. After finishing work on `ArrayData` buffer exports, the full IronPython test run was passing everywhere except on `urllib`. Because failures were deterministic, I was able to trace it down to a bug in `HTTPResponse.client.readinto` and come up with a proper workaround (see commit a0da749876a0200e9e472b840c0135d88af16e6b). It would have been much more tricky if the bug was non-deterministic.

slozier:
@BCSharp thanks for the write-up on the dispose pattern. While the `ArrayDataView` object is perhaps not the place for it, I would argue that we still need some sort of finalizer to decrement the counter. Perhaps on `MemoryView`? Consider the following example:
```Python
import gc
b = bytearray()
m = memoryview(b)
del m
gc.collect() # would not be needed with CPython
b.append(1)
```
While it's bad code the fact that it works with CPython means it's more than likely used in the wild...

Now that we have a proper `IBufferProtocol`, using `UnsafeByteList` or `UnsafeByteArray` becomes mostly obsolete. (The issue you ran into would have been eliminated by removing uses of this.)

----

I just pushed a commit, perhaps you have an idea to enable the following:
```Python
import System
import io
b = io.BytesIO()
b.writelines([System.Array[System.Byte]([1,2,3]), b"\1\2\3"])
```
Since I can't have a call-site that would do automatic conversion I was thinking maybe a helper method?
```C#
IBufferProtocol ToBufferProtocol(object obj) {
    if(obj is IBufferProtocol bp) return bp;
    if(obj is byte[] b) return MemoryBufferWrapper(b);
    // ...
    throw PythonOps.TypeError("a bytes-like object is required, not '{0}'", PythonTypeOps.GetName(obj));
}
```

BCSharp:
> While it's bad code the fact that it works with CPython means it's more than likely used in the wild...

If it is widely used in existing Python libraries, that is really bad news, because, as you indicate, it won't work with IronPython without `gc.collect()`. I would argue it would be easier to convince the library owners to add `m.release()` rather than `gc.collect()`, the latter being conditional on which implementation of Python is being used.

The `del` case can be seen as a special case of a more general case when variables simply go out of scope. The following works in CPython:

```python
def f(b):
    m = memoryview(b)

b = bytearray()
f(b)
b.append(1)
```

What I labelled a "bug" in `HTTPResponse.client` is exactly such case. Arguably, it is still bad code because it depends on a specific implementation of garbage collection, which, though documented for CPython, is not a part of the language. Unless we run `GC.Collect(1)` after each scope end, I see no way to mimic CPython's behaviour. I don't think that running forced garbage collection so frequently is practical, for performance reasons.

But having a finalizer that reduces the export count may be beneficial to clients of such libraries, assuming the libraries themselves do not suffer from it. If an IronPython programmer uses such a library and encounters this problem, they can then call `gc.collect()` and move on.  Since this is an issue of Python objects left behind without explicit release, I think the finalizer should then go in classes that implement those Python types, in this instance, indeed, `MemoryView` rather than `ArrayDataView`. 

Unfortunately, this still has the ability to create heisenbugs, and would have hindered me with  `HTTPResponse.client`. Maybe we provide the finalizer only in _Release_ builds? Or a commandline switch that can disable this behaviour? Or the _Debug_ builds would print a warning when `~MemoryView()` has to decrement the exports counter? That would give us an idea how often this is occuring in the wild... Ideally it would only do it for cases when the underlying array is still in use, to avoid warning about benign cases like:

```python
def f():
    b = bytearray()
    m = memoryview(b)

f()
```

Though I don't know if it is technically possible to differentiate.

BCSharp:
> Since I can't have a call-site that would do automatic conversion I was thinking maybe a helper method?

Thanks for the commit. It's nice to see things get simpler with the new `IBufferProtocol`. I agree that  `UnsafeByteList` and `UnsafeByteArray` can be phased out now.

The issue with `io.BytesIO.writelines` reminds me of `Bytes(PythonList bytes)`, where the list elements have to be cast to `byte` at runtime. The binder only handles simple (but common) cases where there is just one object that needs to support Buffer Protocol, the method can tupe its parameter as `IByfferProtocol` and there are no ambiguous overloads. For more complicated cases I think we need a helper method, maybe in `PythonOps`, in the spirit of `TryGetEnumerator`/`GetEnumerator`? I would also make a little more general in handling any type that behaves as `Memory`. Something like:

```c#
bool TryGetBufferProtocol(object obj, out IBufferProtocol bufferProtocol) {
    bufferProtocol = null;
    if (obj == null) return false;

    if (obj is IBufferProtocol bp) {
        bufferProtocol = bp;
    } else if (obj is Memory<byte> mem) {
        bufferProtocol = new MemoryBufferProtocolWrapper(mem);
    } else if (obj is ReadOnlyMemory<byte> rom) {
        bufferProtocol = new MemoryBufferProtocolWrapper(rom);
    } else if (Converter.HasImplicitConversion(obj.GetType(), typeof(Memory<byte>))) {
        bufferProtocol = new MemoryBufferProtocolWrapper((Memory<byte>)obj);
    } else if (Converter.HasImplicitConversion(obj.GetType(), typeof(ReadOnlyMemory<byte>))) {
        bufferProtocol = new MemoryBufferProtocolWrapper((ReadOnlyMemory<byte>)obj);
    }
    return bufferProtocol != null;
}
```
---
On another tangent of `BytesIO`, I was thinking that maybe it can use `ArrayData` as the backing structure in place of `bytes[]`? `BytesIO` has the same size-freeze on exports behaviour as `array.array`:

```python
>>> import io
>>> bio = io.BytesIO()
>>> mv = bio.getbuffer()
>>> bio.write(b"x")
BufferError: Existing exports of data: object cannot be re-sized
>>> mv.release()
>>> bio.write(b"x") 
1
```

slozier:
> I would argue it would be easier to convince the library owners to add m.release() rather than gc.collect(), the latter being conditional on which implementation of Python is being used.

Python 2 did not have a `memoryview.release` method which means there is probably lots of code where memoryview just "releases" by going out of scope.

> I don't think that running forced garbage collection so frequently is practical, for performance reasons.

Agreed. Probably not a good idea.

> Or the Debug builds would print a warning when ~MemoryView() has to decrement the exports counter?

Could be a `Debug.Assert(false)`?

slozier:
> I would also make a little more general in handling any type that behaves as Memory.
I tried your proposed method and it threw an `InvalidCastException`? Also, `HasImplicitConversion` doesn't seem particularly efficient. I wonder how it would compare to an `is byte[]` if it has to be called frequently.

BCSharp:
> Python 2 did not have a `memoryview.release` method

Aren't we lucky then that Python 2 is EOL? 😃 

---
> I tried your proposed method and it threw an `InvalidCastException`

Apparently, `obj` has to be first cast to its limit type before cast to `Memory`. This will do the trick:

```c#
bufferProtocol = new MemoryBufferProtocolWrapper(Converter.Convert<Memory<byte>>(obj));
```

> Also, `HasImplicitConversion` doesn't seem particularly efficient. 

Agreed. That was the reason I did not use in in my modification of `PythonOverloadResolver`. The overload resolution is already taxed, comparing every overload pairs, and then doing multiple passes. I am still not sure what to do about it. 

In the case of `GetBufferProtocol`, the choice is more obvious, it is either this or an exeption. A fast track for `byte[]` and maybe for `ArraySegment<byte>` will be in order.

BCSharp:
Actually, if we are using `Converter, it will do the whole job:
```c#
bufferProtocol = Converter.Convert<IBufferProtocol>(obj);
```

Src/IronPython.Modules/_codecs.cs
```diff
@@ -182,41 +184,37 @@ public static PythonTuple escape_decode(CodeContext/*!*/ context, [NotNull]strin
         [DisallowNull]
         private static byte[]? _replacementMarker;
 
-        public static PythonTuple/*!*/ escape_encode([BytesLike,NotNull]ReadOnlyMemory<byte> data, string? errors = null) {
-            var buf = new List<byte>(data.Length);
-            var span = data.Span;
+        public static PythonTuple/*!*/ escape_encode([NotNull]IBufferProtocol data, string? errors = null) {
```
slozier, line 187: While testing, I noticed this method only accepts `Bytes` in CPython.

Src/IronPython.Modules/_codecs.cs
```diff
@@ -182,41 +184,37 @@ public static PythonTuple escape_decode(CodeContext/*!*/ context, [NotNull]strin
...
-        private static PythonTuple latin_1_decode(CodeContext context, ReadOnlyMemory<byte> input, string? errors = null)  // TODO: remove
-            => DoDecode(context, "latin-1", Encoding.GetEncoding("iso-8859-1"), input, errors, input.Length).ToPythonTuple();
-
         public static PythonTuple latin_1_decode(CodeContext context, [NotNull]IBufferProtocol input, string? errors = null) {
-            // pilot interface change
-            using IPythonBuffer buf = input.GetBuffer();
-            ReadOnlyMemory<byte> mem = buf.ToMemory();
-            return DoDecode(context, "latin-1", Encoding.GetEncoding("iso-8859-1"), mem, errors, mem.Length).ToPythonTuple();
+            using IPythonBuffer buffer = input.GetBuffer();
+            return DoDecode(context, "latin-1", Encoding.GetEncoding("iso-8859-1"), buffer, errors, buffer.ItemCount).ToPythonTuple();
```
slozier, line 217: Shouldn't this be the "bytes length" instead of `ItemCount`? Same comment everywhere `ItemCount` is used.
```Python
assert _codecs.latin_1_decode(array.array("I", (1633771873,))) == ('aaaa', 4)
```

Src/IronPython/Runtime/Bytes.cs
```diff
@@ -35,8 +35,9 @@ public class Bytes : IList<byte>, IReadOnlyList<byte>, IEquatable<Bytes>, ICodeF
             _bytes = bytes.ToArray();
         }
 
-        public Bytes([BytesLike, NotNull]ReadOnlyMemory<byte> bytes) {
-            _bytes = bytes.ToArray();
+        public Bytes([BytesLike, NotNull]IBufferProtocol source) {
```
slozier, line 38: If the `ReadOnlyMemory<byte>` overload is gone can we get rid of the `byte[]` overload now?

Src/IronPython.Modules/_codecs.cs
BCSharp: Interesting, it would be good to check all of them by adding more tests.

Src/IronPython.Modules/_codecs.cs
BCSharp: You're right, I cut corners here too much, wanted to save obtaining the span. In general, "bytes length" == `ItemCount * ItemSize`, but I'd rather not use this expression here.

Src/IronPython/Runtime/Bytes.cs
BCSharp: Yes, I intend to do that in the next commit, which will be removal of some obsolete code, like support for `[BytesLike]ReadOnlyMemory` parameters in the binder and the analyzer.

Src/IronPython.Modules/_codecs.cs
slozier: Whatever the solution, we should probably add a test for this anyway.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -342,11 +342,11 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
         public Bytes tobytes() {
             CheckBuffer();
 
-            if (_shape?.Contains(0) ?? false) {
+            if (_shape.Contains(0)) {
                 return Bytes.Empty;
             }
 
-            var buf = _buffer.AsReadOnlySpan();
+            var buf = _buffer!.AsReadOnlySpan();
```
BCSharp, line 349: Is there a smart way to indicate `_buffer` being not null after the `CheckBuffer()` call, withouth changing the signature of `CheckBuffer()`?

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -396,7 +396,7 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
             return cast(format, null);
         }
 
-        public MemoryView cast([NotNull]string format, [NotNull]object shape) {
+        public MemoryView cast([NotNull]string format, [NotNull, AllowNull]object shape) {
```
BCSharp, line 399: `[NotNull, AllowNull]` — I find it not very readable. Not to say the confusion with `NotNull` from `System.Diagnostics.CodeAnalysis`. How about adopting a convention to always `using`-renaming `Microsoft.Scripting.Runtime.NotNullAttribute` to `NotNoneAttribute`?

Src/IronPython/Runtime/MemoryView.cs
slozier: I believe [MemberNotNullAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.membernotnullattribute?view=net-5.0) would allow that, unfortunately it's still considered a preview feature.

Src/IronPython/Runtime/MemoryView.cs
slozier: I would be fine with `NotNone` if there is a conflict.
```C#
using NotNoneAttribute = Microsoft.Scripting.Runtime.NotNullAttribute;
```

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -140,87 +329,91 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
...
+                    ofs += _strides[dim];
                 }
+                return copied;
             }
-
-            return Bytes.Make(stridedBytes);
         }
 
-        public PythonList tolist() {
+        public object tolist() {
```
slozier, line 383: In what case would this not return a `PythonList`?

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -488,66 +585,46 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
...
-                    int dimensionWidth = Converter.ConvertToInt32(shape[i]);
-                    dimensions.Add(dimensionWidth);
-                    firstIndexWidth *= dimensionWidth;
+                if (_numDims == 0) {
+                    throw PythonOps.TypeError("invalid indexing of 0-dim memory");
                 }
 
-                int newStart = _start + start * firstIndexWidth;
-                int newEnd = _start + stop * firstIndexWidth;
-                int newStep = ((long)_step * step).ClampToInt32();
```
slozier, line 519: I guess we can get rid of `PythonOps.ClampToInt32` now? 😏

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -674,41 +767,33 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
             if (_buffer == null) {
                 return value._buffer == null;
             }
+            // TODO: comparing flat bytes is oversimplification; besides, no data copyimg
             return tobytes().Equals(value.tobytes());
```
slozier, line 771: Even if it's not correct, could we change this to something a bit more efficient like:
```C#
 return _buffer.AsReadOnlySpan().SequenceEqual(value._buffer.AsReadOnlySpan());
```

Src/IronPython/Runtime/ArrayData.cs
```diff
@@ -21,9 +22,15 @@ internal interface ArrayData : IList {
...
         int CountItems(object? item);
         IntPtr GetAddress();
+        void AddRange(ArrayData value);
+        void InsertRange(int index, int count, ArrayData value);
         ArrayData Multiply(int count);
         new bool Remove(object? item);
         void Reverse();
+        Span<byte> AsByteSpan();
+        IPythonBuffer GetBuffer(object owner, string format, BufferFlags flags);
+        bool HasBuffer { get; }
```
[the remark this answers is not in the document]
BCSharp: I had the same question.. So I assume it can go now.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -2,131 +2,320 @@
...
+            get {
+                CheckBuffer();
+                return _isFContig;
+            }
+        }
+
+        public bool contiguous {
+            get {
+                CheckBuffer();
+                return _isCContig;
```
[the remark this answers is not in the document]
BCSharp: Good catch.

Finishing Fortran array support is for another PR.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -140,87 +329,91 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
...
-                return _buffer.Strides;
+                return PythonTuple.Make(_strides);
             }
         }
 
         public PythonTuple suboffsets {
             get {
                 CheckBuffer();
-                return _buffer.SubOffsets ?? PythonTuple.EMPTY;
+                return PythonTuple.EMPTY;
```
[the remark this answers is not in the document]
BCSharp: There is already such check in the main constructor so suboffsets cannot slip in, but it doesn't hurt to restate it here.

Src/IronPython/Runtime/MemoryView.cs
BCSharp: when `ndim == 0`.

BCSharp: Maybe I add a test for this case.

Src/IronPython/Runtime/MemoryView.cs
BCSharp: I almost used it in 6a6841d2...
It can go now, but it may come back depending on the approach to #52.

Src/IronPython/Runtime/MemoryView.cs
BCSharp: Well, it would make it even less correct. At least now we have:
```python
assert memoryview(b'axc')[::2] == memoryview(b'ayc')[::2]
```
I intend to continue working on TODOs in `MemoryView` in near future under new PRs.

slozier: Doh, did not think about strides :disappointed:

BCSharp:
>    `BytesIO.getbuffer` should be reworked to not return a copy

Agreed. I think there are more parts in `BytesIO` that could be reworked now that we have `System.Buffers`.

>     Does `CData` work? (very low priority)

To my knowledge, `CData` works as good (or as bad) as before. That is, it does work except for indirect multidimensional arrays (i.e. arrays of pointers to arrays). That last one will require support for suboffsets in `MemoryView` — very low priority for me at the moment. I am planning simple general cleanup in `CData` in the coming weeks. I will also see if there are more loose ends there.
 
>     Should `MemoryView` be an `IPythonBuffer`?

Good point, but debatable. The way I see it `memoryview` _is_ Python Buffer, but elevated to be a first class Python type. In the early CPython 3 versions even more so, for instance it would return `None` for `shape` if `ndim == 0`, just as it is mandated for Python Buffers. `IPythonBuffer` is not exactly CPython  Buffer, but close enough, so conceptually I think that `MemoryView` is an `IPythonBuffer`.
But I don't feel strongly about it. Introducing a `MemoryViewView` of sorts would prevent code like
```C#
IPythonBuffer buf = memoryView.GetBuffer();

// somewhere else, perhaps in a sub-method:
if (buf is MemoryView mv) {
   // have a feast...
}
```
but I also don't feel strongly of adding extra complexity just to prevent it.

On my side, besides the remaining TODOs, I see the following:
* Constructors/initializers of `bytes`/`bytearray` should be able to accept any object implementing Buffer protocol, not just bytes-like objects.
* Slice deletion in `bytearray`/`array.array` can be simplified and made more efficient.
* Perhaps some of the switch on typecode statements in `array.array` can be simplified or eliminated. On the other hand, what is there now works just fine, so it is more like a mind teaser...

PR #814, opened 2020-04-30
Do simple re.ASCII matching

slozier:
by mapping to `RegexOptions.ECMAScript`. Far from perfect but works for basic cases.


slozier:
This causes test failures since the C# regex engine doesn't allow combining the `ECMAScript` and `IgnorePatternWhitespace`.

PR #823, opened 2020-05-05
Get pip working

slozier:


Src/IronPython/Runtime/SimpleNamespace.cs
```diff
@@ -0,0 +1,45 @@
...
+using IronPython.Runtime.Operations;
+using IronPython.Runtime.Types;
+using Microsoft.Scripting;
+using Microsoft.Scripting.Runtime;
+using NotNullAttribute = Microsoft.Scripting.Runtime.NotNullAttribute;
+
+namespace IronPython.Runtime {
+    [PythonHidden, PythonType("types.SimpleNamespace")]
+    public class SimpleNamespace {
+        public SimpleNamespace([ParamDictionary, NotNull]Dictionary<string, object?> kwargs\u00F8) {
```
BCSharp, line 19: Why "kwargsø" and not "kwargs"?

slozier: It's a trick to avoid having the kwargs keyword try to bind to the argument. For example, to allow `types.SimpleNamespace(kwargs=1)`. It's used in a few places in the codebase (for example if you search for uses of `ParamDictionaryAttribute`, of `StringOps.format`). Ideally the binding should be prevented with some sort of attribute instead...

BCSharp: OK, thanks for the explanation!

PR #825, opened 2020-05-07
Misc changes

slozier:
- Export `os.symlink` on Windows (even if it's not currently implemented)
- Change map to an `IEnumerator` (similar to https://github.com/IronLanguages/ironpython3/pull/242)

### DIFF
--- a/Src/IronPython.Modules/nt.cs
+++ b/Src/IronPython.Modules/nt.cs
@@ -432,10 +432,20 @@ namespace IronPython.Modules {
 
 #if FEATURE_NATIVE
 
-        [PythonHidden(PlatformsAttribute.PlatformFamily.Windows)]
-        public static void symlink(string source, string link_name) {
-            if (Mono.Unix.Native.Syscall.symlink(source, link_name) == 0) return;
-            throw GetLastUnixError(source, link_name);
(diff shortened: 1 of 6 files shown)


slozier:
Interesting, this change made the following issue pop up:
```
>>> print(*map(ord, "abc"))
TypeError: print() argument after * must be a sequence, not map
```
It fails for any `IEnumerator` which is not also `IEnumerable` (e.g. map and filter in this PR).

The easy fix would be to implement `IEnumerable` as well, but it would not resolve the following case:
```Python
class test:
  def __iter__(self):
    return iter([1])

print(*test())
```
which fails with:
> TypeError: print() argument after * must be a sequence, not Object_1$1

Note that if the failure were valid, the error message should be:
> TypeError: print() argument after * must be an iterable, not test

BCSharp:
I don't know if it is related, but during debugging I've noticed that occasionally the assert in `Runtime/Enumerate.cs`, line 287, fails.

slozier:
Haven't noticed that one. Looking at usage it seems unrelated. May be worth trying to find a repro though...

BCSharp:
> May be worth trying to find a repro though...

It turns out it was triggered by debug support code I wrote to investigate pickling of byte iterators. Nothing of it ended in the repository.

PR #827, opened 2020-05-09
Add functools tests

slozier:


slozier:
I has not looked into it beyond the fact that the methods being tested (`str._formatter_parser` and `str._formatter_field_name_split`) no longer existed on `str`. But upon taking a closer look it looks like they were moved to `_string` so I will update this.

PR #829, opened 2020-05-10
Simplify testing and increase test coverage

slozier:


BCSharp:
What will be the equivalent of:
```
./make test-id~byte
```
?

slozier:
Just `./make test-byte`.

PR #833, opened 2020-05-15
Comparison cleanup

slozier:
Removes comparisons from `type` and adds rich comparison methods to `object`.


slozier:
Resolved https://github.com/IronLanguages/ironpython3/pull/833

PR #846, opened 2020-05-22
Remove array usages

slozier:
Includes https://github.com/IronLanguages/ironpython2/pull/712 and resolves https://github.com/IronLanguages/ironpython2/issues/713

### DIFF
--- a/Src/IronPython/Modules/_bytesio.cs
+++ b/Src/IronPython/Modules/_bytesio.cs
@@ -146,21 +146,16 @@ namespace IronPython.Modules {
                 + "Returns number of bytes read (0 for EOF)."
                 )]
             public BigInteger readinto([NotNull] IBufferProtocol buffer) {
-                try {
-                    using var pythonBuffer = buffer.GetBuffer(BufferFlags.Writable);
-                    var span = pythonBuffer.AsSpan();
+                using var pythonBuffer = buffer.GetBufferNoThrow(BufferFlags.Writable)
+                    ?? throw PythonOps.TypeError("readinto() argument must be read-write bytes-like object, not {0}", PythonTypeOps.GetName(buffer));
 
-                    _checkClosed();
+                _checkClosed();
 
-                    int len = Math.Min(_length - _pos, span.Length);
-                    for (int i = 0; i < len; i++) {
-                        span[i] = _data[_pos++];
-                    }
-
-                    return len;
-                } catch (BufferException) {
-                    throw PythonOps.TypeError("readinto() argument must be read-write bytes-like object, not {0}", PythonTypeOps.GetName(buffer));
-                }
+                var span = pythonBuffer.AsSpan();
+                int len = Math.Min(_length - _pos, span.Length);
+                _data.AsSpan(_pos, len).CopyTo(span);
+                _pos += len;
+                return len;
             }
 
             public override BigInteger readinto(CodeContext/*!*/ context, object buf) {
--- a/Src/IronPython/Modules/_bytesio.cs
+++ b/Src/IronPython/Modules/_bytesio.cs
@@ -39,7 +39,7 @@ namespace IronPython.Modules {
                 : base(context) {
             }
 
-            public BytesIO(CodeContext/*!*/ context, object initial_bytes=null)
+            public BytesIO(CodeContext/*!*/ context, params object[] args)
                 : base(context) {
             }
 
--- a/Src/IronPython/Modules/_bytesio.cs
+++ b/Src/IronPython/Modules/_bytesio.cs
@@ -4,19 +4,18 @@
 
 using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Dynamic;
-using System.IO;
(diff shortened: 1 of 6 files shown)


Src/IronPython/Modules/_bytesio.cs
```diff
@@ -146,59 +145,39 @@ public BytesIO(CodeContext/*!*/ context, object initial_bytes=null)
...
-                }
-
-                return len;
-            }
-
-            public BigInteger readinto([NotNull]ArrayModule.array buffer) {
-                _checkClosed();
+            public BigInteger readinto([NotNull] IBufferProtocol buffer) {
+                try {
+                    using var pythonBuffer = buffer.GetBuffer(BufferFlags.Writable);
```
slozier, line 150: Any ideas how to make this try/catch nicer? I was thinking a `TryGetBuffer` extension method might do the trick.

Src/IronPython/Modules/_bytesio.cs
```diff
@@ -146,59 +145,39 @@ public BytesIO(CodeContext/*!*/ context, object initial_bytes=null)
...
-                    return readinto(bytes);
-                }
-
-                if (buf is ArrayModule.array array) {
-                    return readinto(array);
-                }
-
-                _checkClosed();
-                throw PythonOps.TypeError("must be read-write buffer, not {0}", PythonTypeOps.GetName(buf));
+                var bufferProtocol = Converter.Convert<IBufferProtocol>(buf);
```
slozier, line 167: Not ideal since it throws the wrong error message.

Src/IronPython/Modules/_io.cs
```diff
@@ -1128,14 +1128,15 @@ public class BufferedWriter : _BufferedIOBase, IDynamicMetaObjectProvider {
...
+                using var buffer = Converter.Convert<IBufferProtocol>(buf).GetBuffer();
+                var bytes = buffer.AsReadOnlySpan();
                 lock (this) {
                     if (_writeBuf.Count > _bufSize) {
                         FlushNoLock(context);
                     }
 
                     int count = _writeBuf.Count;
-                    _writeBuf.AddRange(bytes);
+                    _writeBuf.AddRange(bytes.ToArray());
```
slozier, line 1139: Having to make a copy here is not ideal.

Src/IronPython/Modules/_bytesio.cs
BCSharp: I think the message is close enough. But for what types this overload is intended to work?

Src/IronPython/Modules/_bytesio.cs
BCSharp: I agree, it would make the code nicer. Though it will "try" though a try/catch, so no performance improvement.

Src/IronPython/Modules/_io.cs
BCSharp: I have a bytes enumerator-like extension in the works here for `IPythonBuffer`.

Src/IronPython/Modules/_bytesio.cs
slozier: Honestly I don't know, but it's an override of a virtual method so I don't know if it'll end up getting invoked somewhere via C#?

Src/IronPython/Modules/_io.cs
slozier: Sounds good, will wait for your PR then.

Src/IronPython/Modules/_bytesio.cs
BCSharp: OK, then even more reasons to believe that the message is good enough...

Src/IronPython/Modules/_io.cs
BCSharp: OK, I'm gonna jump it to the front of the line.

Src/IronPython/Modules/_bytesio.cs
slozier: Not sure if it's any better:
```C#
if (!buffer.TryGetBuffer(out IPythonBuffer pythonBuffer, flags: BufferFlags.Writable)) {
    throw PythonOps.TypeError("readinto() argument must be read-write bytes-like object, not {0}", PythonTypeOps.GetName(buffer));
}
using (pythonBuffer) {
    // ...
}
```

BCSharp: Wow, this does look weird to me. Maybe I am just not used to such a pattern.

slozier: Yeah, not quite sure how else I'd use a `using` with a try pattern. Maybe this instead:
```C#
using var pythonBuffer = buffer.GetBufferNoThrow(BufferFlags.Writable)
    ?? throw PythonOps.TypeError("readinto() argument must be read-write bytes-like object, not {0}", PythonTypeOps.GetName(buffer));
```

BCSharp: To me, this feels much better.

PR #852, opened 2020-05-24
Implement enumeration of bytes for IPythonBuffer

BCSharp:
Intended for constructors/initializers of `bytes`/`bytearray`, `memoryview` comparisons, and #846.

### DIFF
--- a/Src/IronPython.Modules/_codecs.cs
+++ b/Src/IronPython.Modules/_codecs.cs
@@ -275,7 +275,7 @@ namespace IronPython.Modules {
             using IPythonBuffer buffer = input.GetBuffer();
             return PythonTuple.MakeTuple(
                 StringOps.DoDecode(context, buffer, errors, "raw-unicode-escape", StringOps.CodecsInfo.RawUnicodeEscapeEncoding),
-                buffer.ItemCount * buffer.ItemSize
+                buffer.NumBytes()
             );
         }
 
@@ -302,7 +302,7 @@ namespace IronPython.Modules {
             using IPythonBuffer buffer = input.GetBuffer();
             return PythonTuple.MakeTuple(
                 StringOps.DoDecode(context, buffer, errors, "unicode-escape", StringOps.CodecsInfo.UnicodeEscapeEncoding),
-                buffer.ItemCount * buffer.ItemSize
+                buffer.NumBytes()
             );
         }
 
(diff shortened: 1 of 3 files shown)


slozier:
Thanks for the PR.

I also don't see why it would be useful for the bytes/bytearray constructors/initializiers? Since these (I think) all require bytes-like objects the `Span` enumerator would be equivalent (and maybe more optimized by the runtime)?

Also doesn't seem to provide benefits for my PR (unless I'm missing something) since `AddRange` needs an `IEnumerable<byte>?

Is there any non-C-contiguous buffer we can use to test this code path?

BCSharp:
> I also don't see why it would be useful for the bytes/bytearray constructors/initializiers? Since these (I think) all require bytes-like objects ...

I, too, had thought that they all require bytes-like objects (in the sense that they must be C-contiguous). But then I discovered that this works:
```
>>> mv = memoryview(b'abcd')[1::2]
>>> bytes(mv)
b'bd'
>>> bytearray(mv)
bytearray(b'bd')
>>> import array
>>> array.array('B', mv)
array('B', [98, 100])
``` 
If you wonder whether there is maybe special support for `memoryview`, then no, the same still works with strided Numpy ND-arrays.

> ... the `Span` enumerator would be equivalent (and maybe more optimized by the runtime)?
> 
> Also doesn't seem to provide benefits for my PR (unless I'm missing something) since `AddRange` needs an `IEnumerable?

Assuming that `write` rejects non-C-contiguous buffers, then the question is the trade-off of memory efficiency v.s CPU efficiency. `AddRange` has optimized fast-paths for cases like arrays, so it is indeed fast. Converting a span to an array is a fast memcpy, so it wins on speed.
To avoid a data copy, the span can indeed be best iterated directly and then use `IList<byte>.Add()`. This will be more efficient than the extension method from this PR.  I proposed this PR under the assumption that `write` should be able to accept non-contiguous buffers as well, but I did not test this assumption.

Perhaps a balanced approach would be the best: do data copy for small buffers, and for big buffers copy + `AddRange` in some fixed size chunks.

> Is there any non-C-contiguous buffer we can use to test this code path?

To check out CPython's behaviour, I have been using strided memoryviews, under the assumption that `memoryview` is not treated specially in CPython. When in doubt, use Numpy and/or PIL. So far, I have not seen a case when a strided `memoryview` would work but a strided ND-array woud not.
For IronPython tests, a strided `memoryview` is fine.

slozier:
Interesting, it looks like the `bytes` constructor indeed uses `PyBUF_FULL_RO` in its buffer request.

> Assuming that write rejects non-C-contiguous buffers

It looks like this is indeed the case:
```
>>> io.BufferedWriter(io.BytesIO()).write(memoryview(b'abcd')[1::2])
BufferError: memoryview: underlying buffer is not C-contiguous
```

I think I will leave the `ToArray` in there. The whole implementation is not very efficient anyway so no point in going out of my way to try and optimize this particular bit. Since the whole thing is essentially just a queue maybe I could try using a `ReadOnlySequence<byte>`.

BCSharp:
`ReadOnlySequence<byte>` looks promising.

PR #853, opened 2020-05-24
Optimize slice deletion in bytearray/array.array

BCSharp:


BCSharp:
Indeed, it evolved into identical code. How about putting it as a method in `ArrayData<T>`? Then we could get rid of `ArrayData.CheckBuffer()` and make `ArrayData<T>.CheckBuffer()` private.

slozier:
That would be fine. It's all internal details so I don't have a strong preference.

PR #856, opened 2020-05-25
Enable test_hmac

slozier:
@BCSharp requested a review since this is adding `StringOps.TryEncodeAscii`

### DIFF
--- a/Src/IronPython/Runtime/Operations/StringOps.cs
+++ b/Src/IronPython/Runtime/Operations/StringOps.cs
@@ -1766,6 +1766,20 @@ namespace IronPython.Runtime.Operations {
             return preamble.Length > 0 && bytes.StartsWith(preamble) ? preamble.Length : 0;
         }
 
+        internal static bool TryEncodeAscii(string s, out Bytes? ascii) {
+            var bytes = new byte[s.Length];
+            for (int i = 0; i < s.Length; i++) {
+                if (s[i] < 0x80) {
+                    bytes[i] = (byte)s[i];
+                } else {
+                    ascii = default;
+                    return false;
+                }
+            }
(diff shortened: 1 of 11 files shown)


Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1766,6 +1766,20 @@ public static int rindex([NotNull]this string self, [NotNull]string sub, object?
...
+                if (s[i] < 0x80) {
+                    bytes[i] = (byte)s[i];
+                } else {
+                    ascii = default;
+                    return false;
+                }
+            }
+            ascii = Bytes.Make(bytes);
+            return true;
+        }
```
BCSharp: Something like the following will probably be faster if there are no errors:
```c#
internal static bool TryEncodeAscii(string s, [NotNulWhen(true)]out Bytes? ascii) {
    try {
        ascii = DoEncode(DefaultContext.Default, s, "strict", "ascii", Encoding.ASCII, false);
        return true;
    } catch (UnicodeEncodeException) {
        ascii = default;
        return false;
    }
}
```
But it is slower (`try`/`catch`) if there are errors. From the usage I see that you want tor throw an exception on errors anyway so it probably doesn't matter.

Another possibility:
```c#
internal static Bytes DoEncodeAscii(CodeContext context, string s)
    => DoEncode(context, s, "strict", "ascii", Encoding.ASCII, includePreamble: false);
```
and do try-catch-rethrow where you need it. This will make it similar to `DoEncodeUtf8` that we already have.

slozier: Thanks for the suggestions. I included both variants (although I don't know if the `DoEncodeAscii` will be useful anywhere).

PR #861, opened 2020-05-26
Support any buffer type for construction of bytes/bytearray

BCSharp:


Src/IronPython/Runtime/Bytes.cs
```diff
@@ -36,8 +36,16 @@ public class Bytes : IList<byte>, IReadOnlyList<byte>, IEquatable<Bytes>, ICodeF
         }
 
         public Bytes([BytesLike, NotNull]IBufferProtocol source) {
-            using IPythonBuffer buffer = source.GetBuffer();
-            _bytes = buffer.AsReadOnlySpan().ToArray();
+            using IPythonBuffer buffer = source.GetBuffer(BufferFlags.FullRO);
```
slozier, line 39: No idea if it would be used elsewhere, but could be an extension method:
```C#
public static byte[] ToArray(this IPythonBuffer buffer) {
    if (buffer.IsCContiguous()) {
        return buffer.AsReadOnlySpan().ToArray();
    } else {
        var bytes = new byte[buffer.NumBytes()];
        int i = 0;
        foreach (byte b in buffer.EnumerateBytes()) {
            bytes[i++] = b;
        }
        return bytes;
    }
}
```

Src/IronPython/Runtime/ArrayData.cs
```diff
@@ -102,18 +102,30 @@ internal class ArrayData<T> : ArrayData, IList<T>, IReadOnlyList<T> where T : st
         }
 
         public void AddRange(IPythonBuffer data) {
-            ReadOnlySpan<byte> dataSpan = data.AsReadOnlySpan();
+            if (data.IsCContiguous()) {
```
slozier, line 105: Same here, seems like you could hide the details inside a `CopyTo(IPythonBuffer, Span<byte>)` extension?

Src/IronPython/Runtime/Bytes.cs
BCSharp: Looks good, and maybe it will be useful during pickling.

Src/IronPython/Runtime/ArrayData.cs
BCSharp: OK.

PR #869, opened 2020-06-01
Update memoryview typecodes

BCSharp:
This PR removes typecode `u` (was for Unicode), and adds `?`, `n`, `N`, and `P`. Also, handling of byteorder codes is corrected and expanded.

For some of these codes, [Python does not prescribe fixed size in bits](https://docs.python.org/3/library/struct.html?highlight=struct#format-characters), so some decisions have to be made.

`P` stands for a void pointer, so it should be 32-bit for 32-bit builds, and 64-bit for 64-bit builds. Python represents it as unsigned integer that corresponds to `void*` in C. In .NET, we have `IntPtr`/`UIntPtr` for the purpose of representing a pointer. I thought it is more in spirit of this typeformat to choose one of them as `P`, although it is debatable, since it introduces incompatibility with CPython. As for the choice between those two types, I used `UIntPtr` because Python uses unsigned values as well. During the assignment, both signed and unsigned values (in range) are accepted (like in CPython), as well as `IntPtr`/`UIntPtr`.

Typecodes `n`/`N` stand for `ssize_t`/`size_t` in C and integers in Python, without a standard size specified. So only native size is available (`@`). This means the size is implementation dependent. In C, `size_t` stands for an integer type capable of holding a size of any object in general, and an array in particular (arrays tend to be the largest structures). Consequently, this type also stands for a type holding a largest possible array index value. In CPython, it is 64-bit for 64-bit builds and 32-bit for 32-bit builds. It is also the size of `sys.maxsize` value. In .NET, even for 64-bit builds, a maximum size of an array is (a bit less than) 2GB, (or 4GB if specifically configured, but even then, the number of elements is limited to be < 2^31). This is reflected by IronPython's `sys.maxsize` being `Int32.MaxValue`. So I believe it is appropriate to set the size of `N` to 32 bits. `n` is simply the signed version of `N` so it follows the suit.

In general, I think it would be good to integrate `TypecodeOps` with code in `PythonStruct`. Python's documentation around `memoryview` often delegates specifics to the documentation on `struct`, and `memoryview` C code utilizes functions from `struct`. So the two should behave the same way. Having the same code doing the same job could also eliminate potentially subtle bugs.

On a last note, `test_memoryview.py` now passes under CPython 3.4.

### DIFF
--- a/Src/IronPython/Runtime/TypecodeOps.cs
+++ b/Src/IronPython/Runtime/TypecodeOps.cs
@@ -9,59 +9,81 @@ using System.Diagnostics.CodeAnalysis;
 using System.Numerics;
 using System.Runtime.InteropServices;
 
+using IronPython.Runtime.Operations;
+using Microsoft.Scripting.Utils;
+
 namespace IronPython.Runtime {
     internal static class TypecodeOps {
+        // TODO: integrate with PythonStruct
 
-        public static bool IsByteFormat([NotNullWhen(true)]string? format) {
-            int flen = format?.Length ?? 0;
-            if (flen == 0 || flen > 2) return false;
-            char fchar = format![flen - 1];
-            return fchar == 'B' || fchar == 'b' || fchar == 'c';
-        }
+        public const string ValidByteorder = "@=<>!";
+        public const string ValidScalarCodes = "cbB?hHiIlLqQnNfdP";
+        public const string ValidCodes = ValidScalarCodes + "xsp";
 
-        public static bool TryGetTypecodeWidth([NotNullWhen(true)]string? typecode, out int width) {
-            if (string.IsNullOrEmpty(typecode) || typecode!.Length > 2 || (typecode.Length == 2 && typecode[0] != '@')) {
-                width = 0;
+        public static bool TryDecomposeTypecode(ReadOnlySpan<char> format, out char byteorder, out char code) {
+            if (format.Length == 0 || format.Length > 2) {
+                byteorder = code = default;
                 return false;
             }
-            switch (typecode[typecode.Length - 1]) {
+
+            if (format.Length == 1) {
+                byteorder = '@';
+                code = format[0];
+            } else {
+                byteorder = format[0];
+                code = format[1];
+            }
+            // TODO: add validation of combinations
+            return ValidByteorder.IndexOf(byteorder) >= 0 && ValidCodes.IndexOf(code) >= 0;
+        }
+
+        public static void DecomposeTypecode(string format, out char byteorder, out char code) {
+            if (!TryDecomposeTypecode(format.AsSpan(), out byteorder, out code)) {
+                throw PythonOps.ValueError("invalid typecode");
+            }
+        }
+
+        public static bool IsByteCode(char typecode) {
+            return typecode == 'B' || typecode == 'b' || typecode == 'c';
+        }
+
+        public static bool IsScalarCode(char typecode) {
+            return ValidScalarCodes.IndexOf(typecode) >= 0;
+        }
+
+        public static int GetTypecodeWidth(char typecode) {
+            switch (typecode) {
                 case 'c': // char
                 case 'b': // signed byte
                 case 'B': // unsigned byte
-                    width = 1;
-                    return true;
-                case 'u': // unicode char
+                case '?': // bool
+                    return 1;
                 case 'h': // signed short
                 case 'H': // unsigned short
-                    width = 2;
-                    return true;
+                    return 2;
                 case 'i': // signed int
                 case 'I': // unsigned int
                 case 'l': // signed long
                 case 'L': // unsigned long
                 case 'f': // float
-                    width = 4;
-                    return true;
-                case 'P': // pointer
-                    width = IntPtr.Size;
-                    return true;
+                case 'n': // signed index
+                case 'N': // unsigned index
+                    return 4;
                 case 'q': // signed long long
                 case 'Q': // unsigned long long
                 case 'd': // double
-                    width = 8;
-                    return true;
+                    return 8;
+                case 's': // char pointer
+                case 'p': // char pointer
+                case 'P': // void pointer
+                    return UIntPtr.Size;
                 default:
-                    width = 0;
-                    return false;
+                    return 0;
             }
         }
 
-        public static bool TryGetFromBytes([NotNullWhen(true)]string? typecode, ReadOnlySpan<byte> bytes, [NotNullWhen(true)]out object? result) {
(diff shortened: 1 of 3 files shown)


slozier:
Looks like it should throw `NotImplementedError` for invalid formats:
```Python
import array
a = array.array("u", "abcd")
m = memoryview(a)
m[0] = 1
```
> NotImplementedError: memoryview: format u not supported

Sounds good for the choice of `P` being `UIntPtr` (looks like `_struct` used `IntPtr` and should be corrected). As for `n / N` matching `Int32` it makes sense given the standard index in .NET (`_struct` appears to change with bitness so we should change to be consistent).

As for integrating `TypeCodeOps` with `PythonStruct`, the latter lives in `IronPython.Modules` so things would have to move up to `TypeCodeOps`? But indeed, there's a lot of duplication which should be consolidated (probably in `array` as well).

slozier:
I guess I had misunderstood about using `UIntPtr` objects for `P`. I'm not sure the incompatibility is really worth it. In my experience incompatibilities just end up leading to things not working (although in this case it seems unlikely to be much of an issue). I guess I'd need to see "real" code that uses the `P` format to see if it would cause an issue. It may be preferable to just return the int representation and then letting the user do the cast to a pointer type.

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -9,70 +9,87 @@
...
-                    return true;
+                    return 8;
+                case 's': // char pointer
+                case 'p': // char pointer
+                case 'P': // void pointer
+                    return UIntPtr.Size;
                 default:
-                    width = 0;
-                    return false;
+                    return 0;
```
slozier, line 76: Probably better to throw instead of returning 0?

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -100,29 +119,29 @@ internal static class TypecodeOps {
...
-            if (string.IsNullOrEmpty(typecode) || typecode!.Length > 2 || (typecode.Length == 2 && typecode[0] != '@')) {
-                return false;
-            }
-            switch (typecode[typecode.Length - 1]) {
+        public static bool TryGetBytes(char typecode, object obj, Span<byte> dest) {
+            switch (typecode) {
                 case 'c':
-                    var cbyteVal = (byte)Convert.ToChar(obj);
-                    return MemoryMarshal.TryWrite(dest, ref cbyteVal);
+                    var bytecharVal = (byte)((Bytes)obj)[0];
```
slozier, line 134: I know there's already a check in `MemoryView` to make sure it's a `Bytes` object, but the cast seems odd? Maybe it's fine, just caught my eye...

BCSharp:
> about using `UIntPtr` objects for `P`. I'm not sure the incompatibility is really worth it

I had my doubts about this as well. I implemented it this way because, in my opinion, it better matches the intention, or "spirit" of the operation. Whenever CPython can, it will represent the element as an equivalently typed value, not just a signed or unsigned number. For instance, typecode `?` does not return a value 0 or 1 but `False` or `True`. In such case, the mapping is exact.
```python
assert type(memoryview(b"a").cast('?')[0]) == bool
```
Similarly for typecode `c` (single byte character in C). It is not a number. Python does have byte-strings, but does not have a type for individual byte characters (or Unicode characters for that matter). So a single byte character is approximated by a bytestring of length 1.
```python
assert type(memoryview(b"a").cast('c')[0]) == bytes
```
Following this line of thought, if Python had a dedicated type for a pointer, I believe that would have been used for `P`. Lacking that, it uses an integer.

CPython doesn't have a pointer type, but .NET does. I agree that incompatibility is highly undesirable. But I also regret giving up convenient interop support. So how about this: `N` will be a typecode for pointers represented as unsigned `int`, just as it is in CPython, for full compatibility. Additionally we introduce codes `r`/`R` (currently unused/invalid), to represent `IntPtr`/`UIntPtr`? This scheme has a nice property that it allows an explicit choice between `IntPtr` and `UIntPtr`, rather than always forcing `UIntPtr`, just because CPython's number representation is always unsigned. `IntPtr` is often preferable in .NET (it is CLS-compliant, while `UIntPtr` is not). 

The only "incompatibility" this scheme would carry is that CPython would report `R` typecode as an error, while in IronPython it would have a legitimate meaning. I don't see it as a problem, and there are many cases like this already, since IronPython "extends" Python in many ways.

Src/IronPython/Runtime/TypecodeOps.cs
```diff
@@ -9,70 +9,87 @@
...
-            char fchar = format![flen - 1];
-            return fchar == 'B' || fchar == 'b' || fchar == 'c';
-        }
+        public const string ValidByteorder = "@=<>!";
+        public const string ValidCodes = "xcbB?uhHiIlLqQnNfdspP";
 
-        public static bool TryGetTypecodeWidth([NotNullWhen(true)]string? typecode, out int width) {
-            if (string.IsNullOrEmpty(typecode) || typecode!.Length > 2 || (typecode.Length == 2 && typecode[0] != '@')) {
-                width = 0;
+        public static bool TryDecomposeTypecode(ReadOnlySpan<char> format, out char byteorder, out char code) {
```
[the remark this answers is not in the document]
BCSharp: This is a remnant from when I thought of consolidating `TypecodeOps` with `PythonStruct` code and was making a few provisions to support that. But after reading `_strucs.cs` I think any consolidated solution, though placed in `TypecodeOps` will be largely based on the code from `_strucs.cs` rather than `TypecodeOps.cs`. 

I will remove it in the next cleanup commit, since I see this consolidation work as low priority at this stage.

Src/IronPython/Runtime/TypecodeOps.cs
BCSharp: Sure.

Src/IronPython/Runtime/TypecodeOps.cs
BCSharp: It does seem odd. It is the consequence of the division of responsibilities between `memoryview` and `TypecodeOps` in general, and this method signature in particular. But it works. I saw little benefit rethinking/redesigning this in isolation from `struct`, so I've left it the way it is.

slozier:
> we introduce codes r/R (currently unused/invalid), to represent IntPtr/UIntPtr?

This is fine with the caveat that if Python ever introduces an `r/R` typecode it would trump our use of it.

BCSharp:
If Python ever expands typecodes, we will have a chance to participate in the discussion over the PEP and maybe influence the choice so that no typecode clash occurs. In my experience, CPython folks are quite reasonable and cooperative.

PR #879, opened 2020-06-05
Enable test_doc and test_help

slozier:


slozier:
Disabled failing tests on Mono due to https://github.com/mono/mono/issues/17192

Issue #868, opened 2020-06-01
Overload resolver incorrectly resolves when both 32 and 64-bit parameter type present


BCSharp:
If a method group has two methods that only differ in the bit-size of an integer parameter type (e.g. `uint` and `ulong`), the wider type overload is not selected, even if the value would require it.

Examples:

```
>>> import System
>>> System.UIntPtr(0xFFFFFFFFFFFFFFFF)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OverflowError: Value was either too large or too small for a UInt32.
```

Also, it seems to prefer `double` over `uint` or `ulong`:
```
>>> System.Math.Max(0x7FFFFFFF,1)
2147483647
>>> System.Math.Max(0x80000000,1)
2.14748e+09
```

slozier:
For the `System.UIntPtr` case, sure, we could do `BigInteger`->`ulong` (since there are no signed overloads to pick from). As for `Math.Max`, without doing a runtime check (which I would strongly oppose) we could maybe do `BigInteger`->`long`?

BCSharp:
Since every integer is represented by either `int` or `BigInteger`, for the `IntPtr`/`UIntPtr` case it think it is sufficient to prefer `BigInteger` -> `long`/`ulong` over -> `int`/`uint` overloads.

For the cases like `Math.Max`, when there are both `long`/`ulong` (or `int`/`uint`), if I had to pick, I'd prefer the signed overload. But I do not understand the remark 

> without doing a runtime check

The overload resolver does overload resolution at runtime, so there is already a runtime check. After that, the site cache is updated with the binding rule to be used for the subsequent calls, but the rule itself has constraints (tests) attached to it, which will be checked at runtime as well. Most (all?) constraints check for type equality, but DLR reportedly also supports value-based constraints. We could break the tie by sending non-negative values to the unsigned overloads, and negative to signed. Such constraint would only be generated for those overloads that have `long`/`ulong` or `int`/`uint` ties. Am I overlooking something?

slozier:
I was not aware the DLR could do value-based constraints. I would be curious about the performance of such checks. Looking at `BigInteger` the sign check is cheap enough so assuming it does not have a significant performance impact it would be an option.

BCSharp:
I got the value-based constraints from the DLR documentation. This may be outdated, I will have to experiment with the code to confirm it.

BCSharp:
[...]

PR #891, opened 2020-06-14
Prefer BigInteger to UInt64 conversion over to Int32/UInt32

BCSharp:
This resolves the problem with `System.UIntPtr` from #868.  `BigInteger` to `UInt64` is now performed at narroving level _One_, just like for `Int64`. This puts it at a higher preference level than the conversion from `BigInteger` to `Int32`/`UInt32`, which are done at narrowing level _All_ (lowest).

It does not address the fact that `double`/`float` will be preferred over `ulong`. This is because conversion from `BigInteger` to `Double` and `Single` is already done at narrowing level _None_ (highest priority), which normally means no data loss. This is not entirely accurate:

```
>>> import System 
>>> int(float(System.UInt64.MaxValue)) - System.UInt64.MaxValue
1
>>> int(float(0xFFFFFFFFFFFFFFFF)) - 0xFFFFFFFFFFFFFFFF
-2047
>>> System.UInt64.MaxValue - 0xFFFFFFFFFFFFFFFF        
0
```

However the alternative, that is, to prefer `ulong` over `double` seems worse, because `BigInteger` can really be out of range. But I think that to prefer the conversion to `Single` is really a mistake. And this is what is happening with `Math.Max`:

```
>>> hex(int(System.Math.Max(0xFFFFFFF000000000,1)))    
'0x10000000000000000'
>>> type(System.Math.Max(0xFFFFFFF000000000,1))    
<class 'Single'>
```

Therefore at least I would like to remove the preference of `Single`.

### DIFF
--- a/Src/IronPython/Runtime/Binding/PythonOverloadResolver.cs
+++ b/Src/IronPython/Runtime/Binding/PythonOverloadResolver.cs
@@ -71,6 +71,26 @@ namespace IronPython.Runtime.Binding {
                 return basePreferred;
             }
 
+            if (Converter.IsNumeric(arg.LimitType)) {
+                if (Converter.IsFloatingPoint(arg.LimitType)) {
+                    if (Converter.IsFloatingPoint(candidateOne.Type)) {
+                        if (!Converter.IsFloatingPoint(candidateTwo.Type)) {
+                            return Candidate.One;
+                        }
+                    } else if (Converter.IsFloatingPoint(candidateTwo.Type)) {
+                        return Candidate.Two;
+                    }
+                } else { // arg is integer
+                    if (Converter.IsInteger(candidateOne.Type)) {
+                        if (!Converter.IsInteger(candidateTwo.Type)) {
+                            return Candidate.One;
+                        }
+                    } else if (Converter.IsInteger(candidateTwo.Type)) {
+                        return Candidate.Two;
+                    }
+                }
+            }
+
             bool isBytesLikeOne = IsBytesLikeParameter(candidateOne);
             bool isBytesLikeTwo = IsBytesLikeParameter(candidateTwo);
 
--- a/Src/IronPython/Runtime/Converter.cs
+++ b/Src/IronPython/Runtime/Converter.cs
@@ -429,6 +429,7 @@ namespace IronPython.Runtime {
         private static readonly Type DoubleType = typeof(System.Double);
         private static readonly Type DecimalType = typeof(System.Decimal);
         private static readonly Type Int64Type = typeof(System.Int64);
+        private static readonly Type UInt64Type = typeof(System.UInt64);
         private static readonly Type CharType = typeof(System.Char);
         private static readonly Type SingleType = typeof(System.Single);
         private static readonly Type BooleanType = typeof(System.Boolean);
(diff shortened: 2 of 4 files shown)


slozier:
Apologies for the delay, I wanted to try this against some of the binder tests (which were not in working order). I've pushed the tests to https://github.com/IronLanguages/ironpython3/pull/893. I have not yet merged the tests since it would cause this PR to fail.

This appears to break `str(System.Int64())` (from `test_methoddispatch`). Also getting some failures in `test_methodbinder2` although those may be expected?

BCSharp:
> Apologies for the delay

No problem, I have been preoccupied with other projects myself.

> This appears to break `str(System.Int64())` (from `test_methoddispatch`).

Yes, it does, due to the removal of line 564 in Src/IronPython/Runtime/Converter.cs in commit 1f36bc4. This change deprioritized conversion from `BigInteger` to `Single`, but a side effect is that overloads with `BigInteger` and `Single` are now ambiguous. For this particular case (`str(...)` constructor) it could have been mitigated by either removing the `Single` overload (it seems it does not add anything that the `Double` overload couldn't handle?) or adding a proper `Int64` overload to the collection (in C#, calling `StringOps.__new__` with `Int64` already gets rejected as ambiguous). So the changed behaviour is consistent with C#, but in IronPython, we can do better. 

I think the root cause of the whole issue here is the hardcoded rules in `Converter.HasImplicitNumericConversion` that pretend that there is an implicit conversion from `BigInteger` to `Single`, `Double`, and `Complex`, while in reality no such implicit conversions exist. On top of that, the code assumes that if there is an implicit conversion, it will be lossless so it qualifies when the narrowing level is _None_. This creates all kinds of problems with `BigInteger`, which in a general case cannot be losslessly converted to any of those types. I think this should be eventually rectified, but in this PR I tried to limit the changes to minimum, leaving a bigger overhaul for another PR (or maybe several PRs). It would also help if we had better test coverage of the resolver, so I appreciate your work in #893.

One of the rules the resolver uses is that in case of several candidates, the most narrow type gets the highest preference. This makes sense for reference types, because it favours a derived type over a base type. It is more questionable for numeric types but it can still work fine assuming that the "narrowness" is correctly defined (in the sense that casting from a narrower type to a wider type never causes information loss). Pretending that there is a lossless conversion from `BigInteger` to `Single` solves the ambiguity in the `str` constructor, but creates unwanted side effects. I think this is not a proper way or place of resolving this ambiguity. It seems to me that the intended place for it is the `SelectBestConversionFor` override. So my proposed solution is a new rule that breaks the ambiguity by preferring conversions that maintain "integerness". 

> Also getting some failures in test_methodbinder2

I haven't had time to look into it yet.

BCSharp:
>> Also getting some failures in `test_methodbinder2`

> I haven't had time to look into it yet.

Looking into all currently occurring failures and errors in `test_methodbinder2`, I'd say they are expected. That is, the way how IronPython resolves the overloads is what I expect, and what the tests expect is incorrect.

slozier:
> > > Also getting some failures in `test_methodbinder2`
> 
> > I haven't had time to look into it yet.
> 
> Looking into all currently occurring failures and errors in `test_methodbinder2`, I'd say they are expected. That is, the way how IronPython resolves the overloads is what I expect, and what the tests expect is incorrect.

Sounds good. I'll merge in the other PR then patch this one to get `test_methodbinder2` working. Will make it easier to see what changed as a consequence.

slozier:
Alright, I finally got around to patching `test_methodbinder2`. I just made it pass but did not do any checks to make sure the overload resolution changes make sense. Hopefully I can find a bit of time to do that this weekend.

PR #905, opened 2020-07-02
Calculate fraction of BigInteger division

### LINKED ISSUE #844: long true division lacks precision for very large values
Causes a failure in test_fractions
```Python
assert int('2'*400+'7') / int('3'*400+'1') == 2 / 3
```

jameslan:
Fix #844


slozier:
Thanks for the PR!

While this is definitely an improvement in certain cases, it causes a regression in the following case:
```Python
x = int('2'*400+'7')
assert x / (x+1) == 1.0
```

jameslan:
Got it. Add the logic to round to the nearest FPU.

Issue #906, opened 2020-07-02
EngineTest.ScenarioStreamRedirect fails with a LookupException


slozier:
It looks like this is due to the the standard library not being available and so none of the codecs are available via `_codecs.lookup`. The failure occurs even though the particular encoding causing the failure (cp437) is available to use via `bytes.decode`/`str.encode`. This is because `io.TextIOWrapper` does not make use of the `StringOps.TryGetEncoding` mechanism but instead relies on `PythonOps.LookupEncoding`.

Related to https://github.com/IronLanguages/ironpython2/issues/659 (probably the same issue)

@BCSharp not sure if this something you encountered before?

BCSharp:
Yes, I was aware of that. This is one of the main reasons why several tests at the top of `test_codecs_stdlib` are not passing yet. I am not quite sure what is the best course of action though. Here are some of my musings on this topic.

1. Python documentation of `codecs` reads:
   > While the builtin `open()` and the associated `io` module are the recommended approach for working with encoded text files, this module provides additional utility functions and classes that allow the use of a wider range of codecs when working with binary files.

   This tells me that there is not much need to use text encodings in `codecs` directly, as long as `open()` and `io` are optimized to use .NET encodings, which should be the case anyway for performance reasons. `codecs` seems to me a legacy from Python 2.

2. But from the compatibility point of view, the `codecs` module should be available and complete. The quickest way to ensure this is to provide a proxy class to any .NET `Encoding` that can impersonate a Python codec by providing a `codecs`-like interface (`codecs.CodecInfo`, `codecs.IncrementalEncoder`, `codecs.IncrementalDecoder`, `codecs.StreamReader`, `codecs.StreamWriter`). And if not "the quickest" then at least the most comprehensive. There are some challenges with this approach (e.g. preserving the   state of incremental encoding/decoding is done differently in .NET than Python), but once done, all .NET codecs become available through `codecs`. Remaining codecs written in pure Python top it up and there is only a small number of codecs left out that would have to be manually addressed.

[...]

PR #914, opened 2020-07-30
Support custom encoding error handler

BCSharp:
`PythonEncoderFallback` in `StringOps` had the ability to support custom encoding error handlers, but only those that provide fallback data as string. On the other hand, `PythonEncoderFallback` in `PythonEncoding` has the ability to process fallback data in a form of raw bytes, but not string and it did not support custom error handlers. In Python, a custom error handler may produce fallback data as either string or bytes.

This PR merges both solutions into one: `PythonErrorHandlerEncoding.PythonHandlerEncoderFallback` so that custom error handlers during encoding may be correctly supported.

Moving the encoding cursor from a custom error handler is still not supported.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -3605,7 +3605,7 @@ namespace IronPython.Runtime.Operations {
             return (EncoderFallbackException)ctor.Invoke(new object[] { message, charUnknownHigh, charUnknownLow, index });
         }
 
-        public static Exception UnicodeEncodeError(string message, int runeUnknown, int index) {
+        internal static Exception UnicodeEncodeError(string message, int runeUnknown, int index) {
             if (runeUnknown <= char.MaxValue) {
                 return PythonOps.UnicodeEncodeError(message, (char)runeUnknown, index);
             } else {
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -3605,6 +3605,15 @@ namespace IronPython.Runtime.Operations {
             return (EncoderFallbackException)ctor.Invoke(new object[] { message, charUnknownHigh, charUnknownLow, index });
         }
 
+        public static Exception UnicodeEncodeError(string message, int runeUnknown, int index) {
+            if (runeUnknown <= char.MaxValue) {
+                return PythonOps.UnicodeEncodeError(message, (char)runeUnknown, index);
+            } else {
+                string s = char.ConvertFromUtf32(runeUnknown);
+                return PythonOps.UnicodeEncodeError(message, s[0], s[1], index);
+            }
+        }
+
         public static Exception IOError(Exception inner) {
             return OSError(inner.Message, inner);
         }
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -137,24 +138,20 @@ namespace IronPython.Runtime.Operations {
 
         [LightThrowing]
         internal static object LookupEncodingError(CodeContext/*!*/ context, string name) {
-            Dictionary<string, object> errorHandlers = context.LanguageContext.ErrorHandlers;
-            lock (errorHandlers) {
-                if (errorHandlers.ContainsKey(name))
-                    return errorHandlers[name];
-                else
-                    return LightExceptions.Throw(PythonOps.LookupError("unknown error handler name '{0}'", name));
-            }
+            ConcurrentDictionary<string, object> errorHandlers = context.LanguageContext.ErrorHandlers;
+            if (errorHandlers.TryGetValue(name, out object? handler))
+                return handler;
+            else
+                return LightExceptions.Throw(PythonOps.LookupError("unknown error handler name '{0}'", name));
         }
 
         internal static void RegisterEncodingError(CodeContext/*!*/ context, string name, object? handler) {
-            Dictionary<string, object> errorHandlers = context.LanguageContext.ErrorHandlers;
+            ConcurrentDictionary<string, object> errorHandlers = context.LanguageContext.ErrorHandlers;
 
-            lock (errorHandlers) {
-                if (!PythonOps.IsCallable(context, handler))
-                    throw PythonOps.TypeError("handler must be callable");
+            if (!PythonOps.IsCallable(context, handler))
+                throw PythonOps.TypeError("handler must be callable");
 
-                errorHandlers[name] = handler;
-            }
+            errorHandlers[name] = handler;
         }
 
         internal static PythonTuple LookupEncoding(CodeContext/*!*/ context, string encoding) {
--- a/Src/IronPython/Runtime/PythonEncoding.cs
+++ b/Src/IronPython/Runtime/PythonEncoding.cs
@@ -250,9 +250,9 @@ namespace IronPython.Runtime {
             public override int GetByteCount(char[] chars, int index, int count, bool flush) {
                 var fbuf1 = GetPythonEncoderFallbackBuffer(_pass1encoder);
 
-                fbuf1?.Prepare(chars, forEncoding: false);
+                fbuf1?.PrepareIncrement(chars, forEncoding: false);
                 int numBytes = _pass1encoder.GetByteCount(chars, index, count, flush);
-                fbuf1?.ThrowIfNotEmpty(count, flush);
+                fbuf1?.FinalizeIncrement(count, flush);
 
                 return numBytes;
             }
@@ -260,7 +260,7 @@ namespace IronPython.Runtime {
             public override int GetBytes(char[] chars, int charIndex, int charCount, byte[] bytes, int byteIndex, bool flush) {
                 var fbuf1 = GetPythonEncoderFallbackBuffer(_pass1encoder);
                 var fbuf2 = GetPythonEncoderFallbackBuffer(_pass2encoder);
-                fbuf1?.Prepare(chars, forEncoding: true);
+                fbuf1?.PrepareIncrement(chars, forEncoding: true);
 
                 int written = _pass1encoder.GetBytes(chars, charIndex, charCount, bytes, byteIndex, flush);
 
@@ -281,7 +281,7 @@ namespace IronPython.Runtime {
                         fbuf2 = (PythonEncoderFallbackBuffer)_pass2encoder.FallbackBuffer;
                     }
                 }
-                fbuf2.Prepare(chars, forEncoding: true);
+                fbuf2.PrepareIncrement(chars, forEncoding: true);
 
                 // Restore original fallback bytes
                 var bytes2 = new byte[written];
@@ -302,8 +302,8 @@ namespace IronPython.Runtime {
                 }
 
                 // Check if all fallback bytes are restored properly
-                fbuf1.ThrowIfNotEmpty(charCount, flush);
-                fbuf2.ThrowIfNotEmpty(charCount, flush);
+                fbuf1.FinalizeIncrement(charCount, flush);
+                fbuf2.FinalizeIncrement(charCount, flush);
 
                 return written;
             }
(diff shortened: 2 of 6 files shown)


Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -235,123 +322,179 @@ public abstract class PythonEncoderFallback : EncoderFallback, ICloneable {
         }
 
         protected abstract class PythonEncoderFallbackBuffer : EncoderFallbackBuffer {
-            private readonly char _marker;
-            private readonly Queue<byte> _fallbackBytes;
-
-            private struct ByteCounter {
-                private int _bcmCounter; // used when only counting bytes
-                private int _encCounter; // used during actual encoding
+            private struct RestorableInt {
```
slozier, line 325: Tempted to say that this could be a readonly struct. Maybe not worth the effort if it works...

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -235,123 +322,179 @@ public abstract class PythonEncoderFallback : EncoderFallback, ICloneable {
...
+            public abstract Tuple<ReadOnlyMemory<char>, ReadOnlyMemory<byte>> GetFallbackCharsOrBytes(int runeUnknown, int index);
 
-            public abstract byte[] GetFallbackBytes(char charUnknown, int index);
+            public override bool Fallback(char charUnknown, int index)
+                => FallbackImpl(charUnknown, index);
+
+            public override bool Fallback(char charUnknownHigh, char charUnknownLow, int index)
+                => FallbackImpl(char.ConvertToUtf32(charUnknownHigh, charUnknownLow), index);
+
+            private bool FallbackImpl(int runeUnknown, int index) {
```
slozier, line 395: Too bad we can't use `System.Text.Rune`... Maybe we could fake it on .NET Framework?

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -856,4 +998,105 @@ protected void Throw(byte[] bytesUnknown, int index)
                 => throw new DecoderFallbackException($"'surrogatepass' error handler: not a surrogate character", bytesUnknown, index);
         }
     }
+
+    internal class PythonErrorHandlerEncoding : PythonEncoding {
```
slozier, line 1002: Maybe move this to `PythonErrorHandlerEncoding.cs`? I would suggest the same for the other classes in here but it might make reviewing a bit difficult if it were done in this PR. We could probably add an `IronPython.Runtime.Text` namespace and move all the encoding stuff in there... We could also move some of the classes in StringOps to make the file a bit lighter?

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -856,4 +998,105 @@ protected void Throw(byte[] bytesUnknown, int index)
...
+                base.PrepareIncrement(data, forEncoding);
+            }
+
+            public override Tuple<ReadOnlyMemory<char>, ReadOnlyMemory<byte>> GetFallbackCharsOrBytes(int runeUnknown, int index) {
+                if (_handler == null) {
+                    _handler = LightExceptions.CheckAndThrow(PythonOps.LookupEncodingError(_encoding._context, _encoding._errors));
+                }
+
+                // create the exception object to hand to the user-function...
+                int runeLen = runeUnknown > char.MaxValue ? 2 : 1;
```
slozier, line 1041: Maybe overkill, but this comes up in a few places so maybe a helper? I assume it would just get inlined by the JIT.
```C#
GetUtf16SequenceLength(int rune) => rune > char.MaxValue ? 2 : 1;
```
The name is based on the `Rune.Utf16SequenceLength` property.

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -3608,6 +3605,15 @@ public static partial class PythonOps {
             return (EncoderFallbackException)ctor.Invoke(new object[] { message, charUnknownHigh, charUnknownLow, index });
         }
 
+        public static Exception UnicodeEncodeError(string message, int runeUnknown, int index) {
```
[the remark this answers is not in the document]
BCSharp: Maybe it is a good time to ask: what are the rules to decide when a method in this assembly is made public or internal? 

I made this new overload public, because the other members in this method group are public, so for consistency. But technically, none of them need to be public. This is also true for many other exception factories, and yet the vast majority of them are public.

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -153,60 +169,131 @@ public override Decoder GetDecoder()
...
+
+                if (!(_pass1encoder.Fallback is PythonEncoderFallback) && _parentEncoding.Pass1Encoding.EncoderFallback is PythonEncoderFallback) {
+                    // Non-conformant Encoder implementation, the challenge is to get to the fallback buffer used by such encoder.
+
+                    // Possibility 1: _pass1encoder is EncoderNLS .
+                    // This weirdo (.NET Core only) doe not use Falback and FallbackBuffer properties from its Encoder base class;
+                    // it redefines them as new properties and uses them instead.
+                    // Although the new FallbackBuffer is public, it is not easilly accessible because the EncoderNLS class is internal.
+                    // One way of accessing it is by reflection. This will be handled by GetPythonEncoderFallbackBuffer()
+                    if (_pass1encoder.GetType().FullName == "System.Text.EncoderNLS") return;
```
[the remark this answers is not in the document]
BCSharp: If they do this, than yes, it will be a whole new game. But if it happens, tests in `EncodingTests.cs` will start failing.

Frankly I really detest this particular workaround, probably mostly because the breaking change that .NET Core introduced seems so senseless and unnecessary. To be more specific, I don't know the reason why they did it, but for any possible benefit I can think of that this change provides, there is another way to achieve it without regressing on functionality. I am considering opening an Issue in the corefx repo.

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -153,60 +169,131 @@ public override Decoder GetDecoder()
...
+                    fbuf2?.Reset();
                     return written;
                 }
 
                 // Lazy creation of _pass2encoder
                 if (_pass2encoder == null) {
                     _pass2encoder = _parentEncoding.Pass2Encoding.GetEncoder();
-                    fbuf2 = (PythonEncoderFallbackBuffer)_pass2encoder.FallbackBuffer;
+                    fbuf2 = GetPythonEncoderFallbackBuffer(_pass2encoder);
+                    if (fbuf2 == null && _parentEncoding.Pass2Encoding.EncoderFallback is PythonEncoderFallback && _pass2encoder.GetType().FullName != "System.Text.EncoderNLS") {
```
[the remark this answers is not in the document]
BCSharp: There is still more to come for `PythonEncoding` and things are a little bit in a flux; I was planning to do a pass at the end to get at things like this. But I will refactor this one now, at least I won't overlook it.

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: It could be an interesting exercise. I am curious how it would compare.

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: You mean creating this struct but implementing only the used subset of its public members? Or copying over the whole `Rune.cs` from .NET Core and its dependants? Is there a NuGet package available with it?

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: I like the idea of `IronPython.Runtime.Text` and cleaning up `StringOps` from encoding stuff, I am just not sure the time is ripe. You may have noticed that `PythonEncoding` and friends are much more complex than strictly necessary from the current usage point of view. This is because they support incremental encoding/decoding, but IronPython runtime never uses it in this capacity. 

I developed it this way so that it could possibly evolve to [one of the solutions described earlier](https://github.com/IronLanguages/ironpython3/issues/906#issuecomment-656807221). If so, there is more to come. If not, the whole `PythonEncoding` machinery could be substantially pruned, and perhaps even merged in to one class.

In any case, I'd prefer to do namespace refactoring is a separate PR.

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: This idea had crossed my mind. But I couldn't decide on a name or where to place it. `GetUtf16SequenceLength` is kind of long but since it is already used...

BCSharp:
> Will probably need to go over it another time

No rush, I have more commits of `PythonEncoder.cs` coming up, though slowly, it is summer time.

Briefly, on the codecs front, I intend to commit:
 1. Performance improvements for encoding (chiefly avoiding data copying).
 2. A similar update like in this PR but for decoding.
 3. Consolidating handling of build-in error handles and the exported error handling Python functions. This will probably happen over several PRs, and maybe encoding and decoding separately, as it is easier to think and CPython is not symmetric in handling them either.
 4. Abolishing `PythonAsciiEncoding`. This feels like a reward PR.
 5. Finishing PEP 3120. There are just a few little things left out. Then a proper checkbox in #33 can be checked off. Another reward.

Maybe it will be good to do a standalone comprehensive code review when the dust settles.

Src/IronPython/Runtime/Operations/PythonOps.cs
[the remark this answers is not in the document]
slozier: Well, my preference is to make everything internal unless it needs to be public. The reason being the smaller the public API surface is the easier it is to evolve and maintain the codebase without having to worry about breaking changes. With IronPython 3 being a "new" project it's probably a good time to try and minimize the public API surface.

Examples of things that need(?) to be public:
- Methods that are used in expression trees and may end up in compiled code. Typically these (at least the methods) appear in `IronPython.Compiler.Ast.AstMethods` but there are probably some that are used directly?
- Python types (obviously)
- Things that would need to be exposed so someone could properly implement their own Python types and modules

Another thing that has been nagging at me is `PythonOps` being a sort of grab bag of things as opposed to a class that:
> Contains functions that are called directly from generated code to perform low-level runtime functionality.

I guess creating .NET exceptions could be considered "low-level runtime functionality" so maybe this is fine. Although I could also imagine having a separate class that would be the goto place for creating exceptions. I don't really have a good solution to this.

Src/IronPython/Runtime/PythonEncoding.cs
slozier: I had come up with something along these lines:
```C#
private readonly struct RestorableInt {
    private readonly int _value;
    private readonly int _saved;

    private RestorableInt(int value, int saved) {
        _value = value;
        _saved = saved;
    }

    public int Value => _value;
    public int Saved => _saved;

    public static implicit operator int(RestorableInt dc) => dc._value;
    public static implicit operator RestorableInt(int dc) => new RestorableInt(dc, dc);

    /// <summary>Assignment preserving the save</summary>
    public static RestorableInt operator <<(RestorableInt dc, int value) => new RestorableInt(value, dc._saved);

    public static RestorableInt operator + (RestorableInt dc, int value) => new RestorableInt(dc._value + value, dc._saved);
}
```

Src/IronPython/Runtime/PythonEncoding.cs
slozier: Yes, this is what I meant:
> You mean creating this struct but implementing only the used subset of its public members?

I am not aware of any NuGet package making this available to older runtimes.

Src/IronPython/Runtime/PythonEncoding.cs
slozier: If we had a fake Rune class this would just be `runeUnknown.Utf16SequenceLength`.

Src/IronPython/Runtime/PythonEncoding.cs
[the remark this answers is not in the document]
slozier: Presumably there are technical reasons for why they did it the way they did (for example wanting to have the System.Text.Encoding.CodePages package work on .NET Framework). In any case, you should definitely open an issue in the dotnet/runtime repo. At the very least they might explain the reasoning...

Src/IronPython/Runtime/Operations/PythonOps.cs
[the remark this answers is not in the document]
BCSharp: OK, by these rules the new overload should be internal, since it is not really needed for anybody to " properly implement their own Python types and modules". Unlike the other two overloads, which provide access to a non-public constructor. Another reason to have it internal is that if the signature changes to use `Rune`, I'd rather not make our home-made Rune knock-off public.

I leave other exception factories as they are as it is outside the scope of this PR.

Src/IronPython/Runtime/PythonEncoding.cs
BCSharp: I think it would be best done in a separate PR: `Rune` is usable in many more places than just `PythonEncoderFallbackBuffer`.

In what namespace would you place this "fake Rune"? `System.Text`? `IronPython.Runtime.Text`? Somewhere else?

slozier: Probably `IronPython.Runtime.Text` assuming we create such a namespace.

PR #917, opened 2020-08-04
Fix test__socket failure on macOS

BCSharp:
Fixes one of the remaining problems from the change in #916.

It appears that without process isolation, the server port is owned by the kernel, which then refuses connections from a regular user process. With process isolation, the server process is `ipy` and runs under the same user ID as the client.

I am not sure whether this patch is not simply masking a deeper issue with the current `_socket` implementation.

### DIFF
--- a/Src/IronPythonTest/Cases/IronPythonCasesManifest.ini
+++ b/Src/IronPythonTest/Cases/IronPythonCasesManifest.ini
@@ -198,6 +198,9 @@ RetryCount=2
 [IronPython.modules.misc.test_system_namespaces]
 Timeout=120000 # 2 minute timeout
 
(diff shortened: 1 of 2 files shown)


BCSharp:
It turned out that the Linux failures were a red herring: the `IRONPYTHONPATH` was pointing to a non-existent location and the server process was terminating abnormally.

But the macOS problem was real. As if `SO_REUSEADDR` was not honored... The failures sometimes do not happen, depending on a run, a timing issue perhaps. Deleting `find_free_port()` and letting the server to choose a free port made it work.

I also changed the "wait for the server to start" logic. My mac is old and slow, it takes more than 5 seconds to start `ipy` but I don't want other systems to wait unnecessarily.

PR #918, opened 2020-08-04
Add mutex to prevent running same test in parallel

slozier:


slozier:
I don't like how it will "bloat" the reported test time when it's actually waiting for the test to start. Will see if I can find a way to avoid that.

slozier:
I tried moving the mutex to setup/teardown in the hope that those would not be included in the test duration, but without success. I'm open to suggestions, otherwise this is ready for review.

BCSharp:
I have no better ideas. I was only wondering whether the name of the mutex should be in the "Global\" namespace but I don't know the CI configuration and how it manages test runs.

slozier:
> I was only wondering whether the name of the mutex should be in the "Global" namespace

I also have no idea. :smile: I guess the worst case scenario we get some test failures due to resource contention.

Issue #875, opened 2020-06-04
sys.builtin_module_names should be sorted


slozier:
```Python
assert sys.builtin_module_names == tuple(sorted(sys.builtin_module_names))
```

syn2083:
I notice there isn't a test in the StdLib test_sys.py for this currently, would you want one added to diverge from CPython standard testing, or have this added elsewhere, or as a known background addition that is self validated?

Apologies if this is a lame question..

slozier:
It's a great question. Tests that are not in the standard CPython library go in `Tests` folder. In this case a good place for this would be https://github.com/IronLanguages/ironpython3/blob/master/Tests/modules/system_related/test_sys.py

syn2083:
Aha, thank you for that, I was browsing around, and had seen some of the different test locations, but was unsure of where. Ok, thank you!

PR #920, opened 2020-08-06
Fix the sort on the builtin module names array (#875)

syn2083:
Add Array.Sort to the keys array prior to publication in sys.builtin_module_names so that it is properly sorted as expected.

### DIFF
--- a/Tests/modules/system_related/test_sys.py
+++ b/Tests/modules/system_related/test_sys.py
@@ -231,7 +231,8 @@ f()
         self.assertTrue('dummy result' in str(w[0].message))
     
     def test_builtin_module_names(self):
-        ''' Validate properly sorted module names for issue #875 '''
+        ''' Validate properly sorted module names for issue #875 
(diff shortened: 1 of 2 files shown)


Tests/modules/system_related/test_sys.py
```diff
@@ -229,5 +229,9 @@ def test_getrefcount(self):
         self.assertNotEqual(0, count)
         self.assertTrue(w)
         self.assertTrue('dummy result' in str(w[0].message))
+    
+    def test_builtin_module_names(self):
+        ''' Validate properly sorted module names for issue #875 '''
```
slozier, line 234: Maybe put the full URL for the issue instead of just the number? The repo has moved around a few times before so it saves time when trying to track down a particular issue.

syn2083: no problem! added those in

slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/875

PR #923, opened 2020-08-08
Eliminate data copy when encoding string with PythonEncoding

BCSharp:
Encoding strings is what IronPython currently does with `PythonEncoding`.  If another scenario emerges, (say, encoding a `char` array while reading a file), then some additional tweaks may be needed (like lazy string creation to be passed to a custom error handler).

### DIFF
--- a/Src/IronPython/Runtime/PythonEncoding.cs
+++ b/Src/IronPython/Runtime/PythonEncoding.cs
@@ -10,6 +10,7 @@ using Microsoft.Scripting.Runtime;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Text;
 
 namespace IronPython.Runtime {
@@ -46,8 +47,8 @@ namespace IronPython.Runtime {
         private Encoding Pass2Encoding { get; }
 
         private readonly string _name;
+        private PythonEncoder _residentEncoder;
         private Decoder _residentDecoder;
-        private Encoder _residentEncoder;
 
         public PythonEncoding(Encoding encoding, PythonEncoderFallback encoderFallback, PythonDecoderFallback decoderFallback, string name)
             : base(0, encoderFallback, decoderFallback) {
@@ -58,11 +59,16 @@ namespace IronPython.Runtime {
 
             _name = name ?? "<unknown>";
 
-            // TODO: make lazy
             try {
-                byte[] markerBytes = encoding.GetBytes(new[] { Pass1Marker });
-                CharacterWidth = markerBytes.Length;
-                IsBigEndian = markerBytes[0] == 0;
+                unsafe {
+                    char* markerSpan = stackalloc char[] { Pass1Marker };
+                    CharacterWidth = encoding.GetByteCount(markerSpan, 1);
+                    if (1 <= CharacterWidth && CharacterWidth <= 4) {
+                        byte* markerBytes = stackalloc byte[CharacterWidth];
+                        encoding.GetBytes(markerSpan, 1, markerBytes, CharacterWidth);
+                        IsBigEndian = markerBytes[0] == 0;
+                    }
+                }
             } catch (EncoderFallbackException) {
                 // Q: What encoding cannot encode '?' A: Incomplete charmap.
                 CharacterWidth = 1;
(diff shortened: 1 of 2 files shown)


BCSharp:
When I look at the failed pipeline run #20200808.1, one of the reasons it failed is because the socket server had not started (on time, or at all). What is strange is the timing: 
```
2020-08-08T20:38:04.2173600Z [netcoreapp2.1 ironpython]  test_misc (__main__.SocketMakefileTest) ... ok
2020-08-08T20:38:04.2180514Z [netcoreapp2.1 ironpython]  test_cp5814 (__main__.SocketTest) ... FAIL
```
So there is merely 0.7 ms spent on `test_cp5814`, while the sleep loop there should be waiting at least 10 s before reporting "Server not found". Is `time.sleep(1)` not always working? Or are those timestamps not reliable?

Another thing, I think for diagnostics purposes it would be good to report `EXIT_CODE` for such fail. Then we could tell whether the problem is on the client or the server side. I will submit a patch for it.

Src/IronPython/Runtime/PythonEncoding.cs
```diff
@@ -254,22 +286,104 @@ public class PythonEncoder : Encoder {
...
                 var fbuf1 = GetPythonEncoderFallbackBuffer(_pass1encoder);
                 var fbuf2 = GetPythonEncoderFallbackBuffer(_pass2encoder);
-                fbuf1?.PrepareIncrement(chars, forEncoding: true);
+                fbuf1?.PrepareIncrement(data, forEncoding: true);
+
+                var chars = data.AsSpan();
 
-                int written = _pass1encoder.GetBytes(chars, charIndex, charCount, bytes, byteIndex, flush);
+                int written;
+                unsafe {
```
slozier, line 381: .NET Core provides `Span` overloads for all(?) of these pointer based calls. I wonder if we could define extensions methods which would replicate the API surface for .NET Framework and then use the nicer `Span` syntax. This would just become:
```C#
int written = _pass1encoder.GetBytes(chars.Slice(0, chars.Length), bytes.Slice(0, bytes.Length), flush);
```
and we could add the following extension method:
```C#
#if !NETCOREAPP
internal static int GetBytes(this Encoder encoder, ReadOnlySpan<char> chars, Span<byte> bytes, bool flush) {
    unsafe {
        fixed (char* pChars = &MemoryMarshal.GetReference(chars))
        fixed (byte* pBytes = &MemoryMarshal.GetReference(bytes)) {
            return encoder.GetBytes(pChars, chars.Length, pBytes, bytes.Length, flush);
        }
    }
}
#endif
```

I think we could eliminate most (if not all) of the unsafe blocks this way.

BCSharp: Great idea! Will get on it.

I think .NET 5 also has these. Will `NETCOREAPP` be defined for .NET 5?

slozier:
I don't think the time stamps are good`. The output is probably getting buffered by nunit? For example:
```
2020-08-08T20:38:04.2081684Z [netcoreapp2.1 ironpython]  IronPython.modules.misc.test_zlib
2020-08-08T20:38:04.2085772Z [netcoreapp2.1 ironpython]  test_deflate (__main__.ZlibTest)
2020-08-08T20:38:04.2091181Z [netcoreapp2.1 ironpython]  raw data, no header ... ok
2020-08-08T20:38:04.2095133Z [netcoreapp2.1 ironpython]  test_gzip (__main__.ZlibTest)
2020-08-08T20:38:04.2099452Z [netcoreapp2.1 ironpython]  decompression with gzip header ... ok
2020-08-08T20:38:04.2103353Z [netcoreapp2.1 ironpython]  test_gzip_stream (__main__.ZlibTest)
2020-08-08T20:38:04.2106770Z [netcoreapp2.1 ironpython]  gzip header, uncomplete header ... ok
2020-08-08T20:38:04.2110918Z [netcoreapp2.1 ironpython]  test_gzip_stream_with_extra (__main__.ZlibTest) ... ok
2020-08-08T20:38:04.2115073Z [netcoreapp2.1 ironpython]  test_gzip_with_extra (__main__.ZlibTest)
2020-08-08T20:38:04.2120073Z [netcoreapp2.1 ironpython]  gzip header with extra field ... ok
2020-08-08T20:38:04.2124750Z [netcoreapp2.1 ironpython]  test_header_auto_detect (__main__.ZlibTest)
2020-08-08T20:38:04.2129855Z [netcoreapp2.1 ironpython]  autodetect zlib and gzip header ... ok
2020-08-08T20:38:04.2133548Z [netcoreapp2.1 ironpython]  
2020-08-08T20:38:04.2137566Z [netcoreapp2.1 ironpython]  ----------------------------------------------------------------------
2020-08-08T20:38:04.2141634Z [netcoreapp2.1 ironpython]  Ran 6 tests in 0.774s
2020-08-08T20:38:04.2145172Z [netcoreapp2.1 ironpython]  
```
is clearly not 0.774 seconds...

Src/IronPython/Runtime/PythonEncoding.cs
slozier: Yes, `NETCOREAPP` is also defined with .NET 5. I'm not sure if I'll be adding a .NET 5 target (unless there's some API surface that we want to use), however I'll probably try and get testing working on .NET 5 to make sure there are no issues with it.

PR #933, opened 2020-08-17
Use System.Diagnostics.Process to implement subprocess.Popen, which could be used on posix and windows

### LINKED ISSUE #541: subprocess.Popen does not work on Linux
```
  File "Tests/modules/io_related/test_codecs.py", line 526, in <module>
  File "/mnt/r/Documents/Programming-github/ironpython3/bin/Debug/net45/Lib/iptest/ipunittest.py", line 356, in run_test
  File "/mnt/r/Documents/Programming-github/ironpython3/Src/StdLib/Lib/test/support/__init__.py", line 1817, in run_unittest
  File "/mnt/r/Documents/Programming-github/ironpython3/Src/StdLib/Lib/test/support/__init__.py", line 1792, in _run_suite
test.support.TestFailed: Traceback (most recent call last):
  File "Tests/modules/io_related/test_codecs.py", line 382, in test_cp11334
    p = subprocess.Popen([sys.executable, os.path.join(self.test_dir, "encoded_files", "cp11334_warn.py")], shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "/mnt/r/Documents/Programming-github/ironpython3/Src/StdLib/Lib/subprocess.py", line 850, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/mnt/r/Documents/Programming-github/ironpython3/Src/StdLib/Lib/subprocess.py", line 1396, in _execute_child
    self.pid = _posixsubprocess.fork_exec(
NotImplementedError: fork_exec is currently not implemented
```

jameslan:
Fix #541

### DIFF
--- a/Src/StdLib/Lib/subprocess.py
+++ b/Src/StdLib/Lib/subprocess.py
@@ -1344,7 +1344,7 @@ class Popen(object):
             if stderr == PIPE:
                 self.stderr = io.BufferedReader(open(p.StandardError.BaseStream))
 
-            # dotnet an't redirect stdio to file/stream, thus has to feed from parent
+            # dotnet can't redirect stdio to file/stream, thus has to feed from parent
             if stdin not in (None,  DEVNULL, PIPE):
                 # assume file-like object
                 input = stdin.read()
@@ -1371,7 +1371,7 @@ class Popen(object):
             if timeout is None:
                 self._handle.WaitForExit()
             else:
-                self._handle.WaitForExit(timeout)
+                self._handle.WaitForExit(int(timeout * 1000))
             self.returncode = self._handle.ExitCode
             return self.returncode
 
--- a/Src/StdLib/Lib/subprocess.py
+++ b/Src/StdLib/Lib/subprocess.py
@@ -345,6 +345,7 @@ Popen(["/bin/mycmd", "myarg"], env={"PATH": "/usr/bin"})
 """
 
 import sys
+ironpython = (sys.implementation.name == 'ironpython')
 mswindows = (sys.platform == "win32")
 
 import io
@@ -401,6 +402,11 @@ if mswindows:
         hStdOutput = None
         hStdError = None
         wShowWindow = 0
+elif ironpython:
+    import threading
+    import clr
+    clr.AddReference("System")
+    from System.Diagnostics import Process
 else:
     import _posixsubprocess
     import select
@@ -457,6 +463,8 @@ if mswindows:
 
         __del__ = Close
         __str__ = __repr__
+elif ironpython:
+    __all__.extend(["Process"])
 
 try:
     MAXFD = os.sysconf("SC_OPEN_MAX")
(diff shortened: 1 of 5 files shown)


slozier:
Thanks. Nice way to get things working on posix. Did you look at the ipy2 mono implementation (https://github.com/IronLanguages/ironpython2/blob/master/Src/StdLib/Lib/subprocess.py)? There might be something useful in there.

Is there any advantage to using this on windows instead of the standard implementation? If not maybe give priority to the `if mswindows` conditional.

jameslan:
> Is there any advantage to using this on windows instead of the standard implementation?

I was trying to keep consistency across OSes on IronPython implementation, i.e. IronPython provides identical capability on all OSes, which is limited only by the API design of dotnet. 

The reason for this choice is simple for me: since with this approach, we can't make it exactly the same as CPython on POSIX, why not just make it align with dotnet on all OSes.

Basically I am not very sure how important, from the users' perspective, to keep the behavior identical to CPython.

jameslan:
`CPython.test_subprocess` is still ignored, because it has many cases running `ipy -c "import sys; sys.stdout.write('text')"`, but there's nothing output.

Src/StdLib/Lib/subprocess.py
```diff
@@ -1272,6 +1286,117 @@ def terminate(self):
...
+            self._handle = p
+
+            if stdin == PIPE:
+                self.stdin = open(p.StandardInput.BaseStream)
+            if stdout == PIPE:
+                self.stdout = io.BufferedReader(open(p.StandardOutput.BaseStream))
+            if stderr == PIPE:
+                self.stderr = io.BufferedReader(open(p.StandardError.BaseStream))
+
+            # dotnet an't redirect stdio to file/stream, thus has to feed from parent
```
slozier, line 1347: Typo:
```suggestion
            # dotnet can't redirect stdio to file/stream, thus has to feed from parent
```

Src/StdLib/Lib/subprocess.py
```diff
@@ -1272,6 +1286,117 @@ def terminate(self):
...
+
+        def wait(self, timeout=None, endtime=None):
+            """Wait for child process to terminate.  Returns returncode
+            attribute."""
+            if endtime is not None:
+                timeout = self._remaining_time(endtime)
+            if timeout is None:
+                self._handle.WaitForExit()
+            else:
+                self._handle.WaitForExit(timeout)
```
slozier, line 1374: I think you need to convert from float seconds to integer milliseconds?
```suggestion
                self._handle.WaitForExit(int(timeout * 1000))
```

slozier:
>CPython.test_subprocess is still ignored, because it has many cases running ipy -c "import sys; sys.stdout.write('text')", but there's nothing output.

Interesting, I think it's not flushing before the process terminates.

Src/StdLib/Lib/subprocess.py
jameslan: A lot of cases left untested, because of the unflushed `FileIO` and `BufferredWriter`. There will be more work for the `Popen` implementation once `test_subprocess` could run.

PR #940, opened 2020-09-18
Fix some .NET 5.0 compile issues

slozier:
This is prep work to enable testing with .NET 5.0.

There are still some failures with `test_re_stdlib` which are due to some bugs in .NET 5.0 which should be fixed before its release (https://github.com/dotnet/runtime/issues/42390, https://github.com/dotnet/runtime/issues/42392, https://github.com/dotnet/runtime/issues/42448).

**SupportedOSPlatform**
Should see if we can make use of this to annotate and guard our code (e.g. wherever we use `PythonHidden`).

**SYSLIB0001**
UTF-7 has been obsolete in .NET 5. In particular the `string.GetEncoding` methods will no longer return UTF-7. @BCSharp I believe we are not using this to obtain a UTF-7 encoding anywhere? See https://docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0#utf-7-code-paths-are-obsolete for details.

**SYSLIB0004**
The Constrained Execution Region (CER) feature is not supported. I don't think this is a change specific to .NET 5, just a new warning. This comes up in `ctypes`. Need to file an issue to investigate what the consequences are.

**SYSLIB0006**
Thread.Abort is not supported and throws PlatformNotSupportedException. I don't think this is a change specific to .NET 5, just a new warning. Should investigate to see if there's anything we can use in place of this.

**SYSLIB0011**
BinaryFormatter serialization is obsolete and should not be used. Should be investigated, maybe we can remove the `clr.Serialize/clr.Deserialize` methods.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,6 +27,10 @@ jobs:
       uses: actions/setup-dotnet@v1
       with:
         dotnet-version: '3.1.x'
+    - name: Setup .NET 5.0
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: '5.0.100-rc.1.20452.10'
     - name: Set Mono Version
       if: matrix.os == 'macos-latest'
       run: echo ::add-path::/Library/Frameworks/Mono.framework/Versions/6.4.0/bin
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Setup .NET 5.0
       uses: actions/setup-dotnet@v1
       with:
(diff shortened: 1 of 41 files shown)


BCSharp:
**SYSLIB0001**
Indeed, `System.Text.Encoding.GetEncoding()` is not used to get UTF-7. I think that after factoring out of `IronPython.Runtime.Text` we can limit the warning silencer to just one place of using the `UTF7Encoding` constructor.

slozier:
@BCSharp I'll merge this in if you have no objections, now that .NET 5.0 and VS 16.8 are out.

BCSharp:
Fine with me.

PR #944, opened 2020-09-22
Disable parallel execution for test_gettext

BCSharp:
This works fine on Windows, but not so on macOS/Linux. Apparently, `Mutex` implementation in Mono doesn't work as expected. As a workaround, I run tests with `-frameworks notcoreapp2.1,netcoreapp3.1` and then Mono separately.

I don't know how to make it work under Mono. I am making this pull request anyway since it is better than nothing. @slozier, @jameslan, any ideas?

### DIFF
--- a/Src/IronPythonTest/Cases/CPythonCasesManifest.ini
+++ b/Src/IronPythonTest/Cases/CPythonCasesManifest.ini
@@ -494,6 +494,10 @@ Ignore=true
 Ignore=true
 Reason=ImportError: No module named '_testcapi'
 
+[CPython.test_gettext]
+NotParallelSafe=true
+Reason=Creates/deletes a directory with static name 'xx'
+
 [CPython.test_glob]
 Ignore=true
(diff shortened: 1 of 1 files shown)


slozier:
`Mutex` doesn't work at all, or in combination with the .NET core mutex? If I had to guess, Mono and .NET Core are using incompatible `Mutex` implementations. More work for us, but we could probably accomplish the same thing via lock files.

slozier:
Ah, mutex isn't cross-process on Mono non-Windows. See https://aakinshin.net/posts/namedmutex-on-mono/. It looks like there's a link to an alternative solution using file streams.

BCSharp:
Thanks @slozier for investigating it. The blog post you found really explains a lot. So lock files it is. I suppose this can be implemented in a separate PR.

slozier:
@BCSharp Created a separate issue for the `Mutex` thing: https://github.com/IronLanguages/ironpython3/issues/946

PR #948, opened 2020-09-24
Enable test_bytes from stdlib

BCSharp:
I have disabled `test_bytes.test_from_format` when run under IronPython as it depends in `ctypes.pythonapi`, which is a CPython-specific DLL. This makes the test suite `test_bytes` complete and `test_bytes_stdlib` is redundant hence deleted.

I also have checked off a few bytes-related points in `WhatsNewInPython30.md` as a sign that those aspects should now functionally behave like in Python 3, and if not, an Issue on GitHub is warranted. I have left the checkbox on line 29 unchecked because I am not sure if all IronPython API has been migrated to new `bytes`/`str`. Perhaps the best moment to check that point off is once we get rid of `BytesLikeAttribute`.


slozier (approved):
Nice!

PR #950, opened 2020-09-25
Support __index__ on additional code paths in Bytes/ByteArray construction

BCSharp:
I had to tweak `bytes` construction a little again. Apparently, it is not only  "always call `__bytes__`", but also  "always call `__index__`" — except when constructing from `str` or `int` derived classes…

A similar story in `bytearray`.


slozier (approved):
Looks good to me.

A though that came to be while looking is that `int.from_bytes(2, 'big')` probably shouldn't work (and indeed it doesn't with CPython). Similarly, there should be no `__index__` calls when going through `int.from_bytes`.

BCSharp:
Any idea why `test_smtpnet.SmtpTest('test_connect_starttls')` fails on Azure Pipelines?

slozier:
No idea for the smtp stuff. Looking at the log is seems like `test_smtplib` runs to completion in about 5 seconds, but them becomes stuck and times out after 10 minutes.

Issue #957, opened 2020-10-03
Subclasses of str missing __iter__ attribute


BCSharp:
```python
class A(str): pass
getattr(A("abc"), '__iter__')
```

CPython produces `<method-wrapper '__iter__' of A object at 0x00000000031584F8>` (a factory for an iterator of length 3).
IronPython (release) throws `AttributeError: 'A' object has no attribute '__iter__'`
IronPython (debug) hits `Debug.Assert(false)` at PythonTypes.cs:523

Originally discovered during analysis of #955.

slozier:
This got resolved by https://github.com/IronLanguages/ironpython3/pull/1088

PR #958, opened 2020-10-03
Prevent using str type arg in int.from_bytes

### LINKED ISSUE #955: Assertion error when running test_bytes in debug mode
`test_bytes.test_dunder_index` raises a .NET assertion error when run in `Debug` mode. Haven't looked at the details.

@BCSharp

BCSharp:
Resolves #955, that is, the debug mode tests pass and the error message for `int.from_bytes("abc", 'big')` is aligned with CPython.

It does not resolve the underlying issue, for which I have created a separate issue report (#957).


slozier (approved):
Thanks!

PR #962, opened 2020-10-06
Use UTF-8 as default with ScriptEngine

BCSharp:
This is the last bit of the work to use UTF-8 as default source encoding that I am aware of.

The test for encoding used by `ScriptEngine` was already in place (disabled) but it wasn't testing the right thing. I went to the [CodePlex archive](https://archive.codeplex.com/?p=ironpython) to look up the original report 20472 (_Issue with non-unicode script file encoding_) and fixed the test to test for that.

One caveat: if the user uses `CreateScriptSourceFromFile(file, System.Text.Encoding.Default)`, UTF-8 will be used rather than `Encoding.Default` which on .NET Framework is the active Windows code page. Perhaps this is OK, since the IronPython 3 default **is** UTF-8. If this is not right, the solution might be to modify `Microsoft.Scripting.Utils.StringUtils.DefaultEncoding` to use a clone of ` System.Text.Encoding.Default`.

For now, if somebody really needs to use the active Windows code page, they may either specify it explicitly or clone the default themselves, e.g.:
```C#
CreateScriptSourceFromFile(file, Encoding.GetEncoding(1252)); // or whatever codepage is active
CreateScriptSourceFromFile(file, Encoding.Default.Clone());
```

### DIFF
--- a/Src/IronPython/Runtime/PythonContext.cs
+++ b/Src/IronPython/Runtime/PythonContext.cs
@@ -892,7 +892,13 @@ namespace IronPython.Runtime {
 
             if (sourceEncoding == null) {
                 // not autodetected and not declared, hence use default
-                sourceEncoding = defaultEncoding;
+                if (ReferenceEquals(defaultEncoding, StringUtils.DefaultEncoding)) {
+                    // Override DLR default encoding with Python default
+                    sourceEncoding = DefaultEncoding;
+                } else {
+                    // user-provided encoding
+                    sourceEncoding = defaultEncoding;
+                }
             } else {
                 sourceEncoding = (Encoding)sourceEncoding.Clone();
                 sourceEncoding.DecoderFallback = DecoderFallback.ExceptionFallback;
--- a/Src/IronPython/Runtime/PythonContext.cs
+++ b/Src/IronPython/Runtime/PythonContext.cs
@@ -892,13 +892,7 @@ namespace IronPython.Runtime {
 
             if (sourceEncoding == null) {
                 // not autodetected and not declared, hence use default
-                if (ReferenceEquals(defaultEncoding, StringUtils.DefaultEncoding)) {
-                    // Override DLR default encoding with Python default
-                    sourceEncoding = DefaultEncoding;
-                } else {
-                    // user-provided encoding
-                    sourceEncoding = defaultEncoding;
-                }
+                sourceEncoding = defaultEncoding;
             } else {
                 sourceEncoding = (Encoding)sourceEncoding.Clone();
                 sourceEncoding.DecoderFallback = DecoderFallback.ExceptionFallback;
--- a/Src/IronPythonTest/EngineTest.cs
+++ b/Src/IronPythonTest/EngineTest.cs
@@ -423,17 +423,22 @@ x = 42", scope);
 #if NETCOREAPP
             Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
 #endif
-
+            // This test file is encoded in Windows codepage 1251 (Cyrilic) but lacks a magic comment (PEP-263)
+            string fileName = Path.Combine(Path.Combine(Common.ScriptTestDirectory, "encoded_files"), "cp20472.py");
             try {
-                string fileName = Path.Combine(Path.Combine(Common.ScriptTestDirectory, "encoded_files"), "cp20472.py");
-                _pe.CreateScriptSourceFromFile(fileName, Encoding.GetEncoding(1251)).Compile();
+                _pe.CreateScriptSourceFromFile(fileName).Compile();
 
-                //Disabled. The line above should have thrown a syntax exception or an import error,
-                //but does not.
+                // The line above should have thrown a syntax exception or an import error
+                // because the default source file encoding is UTF-8 and there are decoding errors
 
-                //throw new Exception("ScenarioCodePlex20472");
+                throw new Exception("ScenarioCodePlex20472");
             }
-            catch (IronPython.Runtime.Exceptions.ImportException) { }
+            catch (SyntaxErrorException) { }
+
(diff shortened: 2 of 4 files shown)


Src/IronPython/Runtime/PythonContext.cs
```diff
@@ -892,7 +892,13 @@ internal static Version GetPythonVersion()
 
             if (sourceEncoding == null) {
                 // not autodetected and not declared, hence use default
-                sourceEncoding = defaultEncoding;
+                if (ReferenceEquals(defaultEncoding, StringUtils.DefaultEncoding)) {
```
slozier, line 895: Probably better to use `Encoding.Default` here. I'm tempted to deprecate `StringUtils.DefaultEncoding` since realistically it no longer serves a purpose (which was supporting runtimes which don't define `Encoding.Default`).

That being said, I'm not sure that we should be overriding the default. Is this only to accommodate the case where `CreateScriptSourceFromFile` is called without an encoding?

BCSharp: > Probably better to use `Encoding.Default` here.

I did not want to use `Encoding.Default` here because it would effectively eliminate the possibility to make `CreateScriptSourceFromFile(file, System.Text.Encoding.Default)` behave correctly, which now can be achieved with a fairly modest and backward-compatible modification to the DLR, e.g:

```C#
#if FEATURE_ENCODING
        public static Encoding DefaultEncoding { get; } = Encoding.GetEncoding(Encoding.Default.CodePage);
#else
        public static Encoding DefaultEncoding => Encoding.UTF8;
#endif
```
But maybe this is not needed and the workarounds I proposed originally are good enough.

That being said, this whole thing is a workaround around a deeper issue:

> I'm not sure that we should be overriding the default.

Perhaps the comment there is a bit misleading. Conceptually, this path is setting the correct default, which was not recognized/known by the DLR. DLR default is `StringUtils.DefaultEncoding` but IronPython default is `PythonContext.DefaultEncoding`.

>  Is this only to accommodate the case where CreateScriptSourceFromFile is called without an encoding?

Yes. As a Python developer, I would be surprised to discover that while
```shell
PS> python3 my_prog.py
PS> ipy3 my_prog.py
```
worked fine,
```C#
var src = CreateScriptSourceFromFile("my_prog.py");
```
silently mangled all UTF-8 encoded string literals, and would consider it a bug.

I was trying to avoid (big) changes to the DLR, but if you are planing on working on that part of the DLR, by far my preferred solution would be to add a virtual property `LanguageContext.DefaultEncoding`. There should be only one place where a default encoding for source files is defined, rather than several "defaults" competing with other (BTW, in DLR, there is one in `Microsoft.Dynamic` and one in `Microsoft.Scripting`. I'm unsure which one `PythonContext.SourceCodeReader` is actually getting).

slozier: > it would effectively eliminate the possibility to make `CreateScriptSourceFromFile(file, System.Text.Encoding.Default)`

I would argue that anything using `Encoding.Default` is effectively broken since it has a different meaning on .NET Framework and .NET Core (it's UTF-8 on .NET Core). I guess we should probably eliminate any uses of it from the codebase. Although using it as a placeholder for an undefined encoding is probably fine.

> a fairly modest and backward-compatible modification to the DLR

Unfortunately it is not as simple as that since as you noticed it is defined in two places `Microsoft.Scripting` is internal and used in the `CreateScriptSourceFromFile` definition, while `Microsoft.Dynamic` is the public one (which you used in this PR). Since `Microsoft.Dynamic` references `Microsoft.Scripting` and not the other way around (and internals are not visible) we'd probably have to move the `StringUtils` class down to `Microsoft.Scripting` and then do a type-forward in `Microsoft.Dynamic` to avoid a breaking change. In other words probably not worth the effort.

>  As a Python developer, I would be surprised to discover that...

Well, the `CreateScriptSourceFromFile` overload without an encoding does specify that it uses `Encoding.Default` as its encoding, but you really have to be in the weeds to know that this can vary depending on the runtime so it's probably not a good idea to skip out on specifying the encoding.

> this whole thing is a workaround around a deeper issue

Indeed, the thought had crossed my mind. Assuming everything works out (at a glance it seems like it would), I would be fine with adding the virtual property `LanguageContext.DefaultEncoding` and then reworking the overloads to use that instead of `Encoding.Default`/`StringUtils.DefaultEncoding`. In fact now if probably a good time for the change since I just bumped the minor version of the DLR. I can try to take a look at it tomorrow (assuming you don't beat me to it).

BCSharp: > Well, the `CreateScriptSourceFromFile` overload without an encoding does specify that it uses `Encoding.Default` as its encoding,

Yes, it is specified, but, as you pointed out, has a different meaning on different platforms, and is still not practical, as in all cases the default is a non-throwing encoding. So for most practical cases the user has to instantiate their own throwing UTF-8 encoding and explicitly use it. Sure it can be done, it is easy to forget or be tempted to simply use `Encoding.UTF8`. The whole point of having defaults is to make usage easier, not more tricky. In such case it would be better if the overload of `CreateScriptSourceFromFile` that assumes some default encoding were removed altogether.

> I would be fine with adding the virtual property `LanguageContext.DefaultEncoding`

Great. I leave it up to you. Otherwise I would be fine to remove the reference to `StringUtils.DefaultEncoding` and replace it with `Encoding.Default` and `Encoding.UTF8` to set the encoding according to the Python standard. Note that the encoding provided to `CreateScriptSourceFromFile` (default or explicit) may already be overridden by BOM or a magic comment (as per Python standard).

slozier: Alright, changes are merged in so I guess this particular workaround can be reverted (assuming everything worked out).

PR #963, opened 2020-10-07
Eliminate data copy when calling str() on a bytes-like object

BCSharp:
This also aligns error types and warnings with CPython's.

### DIFF
--- a/Src/IronPython/Runtime/ByteArray.cs
+++ b/Src/IronPython/Runtime/ByteArray.cs
@@ -15,6 +15,7 @@ using System.Text;
 using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
+using IronPython.Runtime.Exceptions;
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
 
@@ -1076,6 +1077,11 @@ namespace IronPython.Runtime {
             lock (this) {
                 return "bytearray(" + _bytes.BytesRepr() + ")";
             }
+        }
+
+        public virtual string __str__(CodeContext context) {
+            PythonOps.Warn(context, PythonExceptions.BytesWarning, "str() on a bytearray instance");
+            return Repr();
         }
 
         public virtual string __repr__(CodeContext context) => Repr();
--- a/Src/IronPython/Runtime/ByteArray.cs
+++ b/Src/IronPython/Runtime/ByteArray.cs
@@ -1080,7 +1080,9 @@ namespace IronPython.Runtime {
         }
 
         public virtual string __str__(CodeContext context) {
-            PythonOps.Warn(context, PythonExceptions.BytesWarning, "str() on a bytearray instance");
+            if (context.LanguageContext.PythonOptions.BytesWarning) {
+                PythonOps.Warn(context, PythonExceptions.BytesWarning, "str() on a bytearray instance");
+            }
             return Repr();
         }
 
--- a/Src/IronPython/Runtime/Bytes.cs
+++ b/Src/IronPython/Runtime/Bytes.cs
@@ -856,7 +856,9 @@ namespace IronPython.Runtime {
         }
 
         public virtual string __str__(CodeContext context) {
-            PythonOps.Warn(context, PythonExceptions.BytesWarning, "str() on a bytes instance");
+            if (context.LanguageContext.PythonOptions.BytesWarning) {
+                PythonOps.Warn(context, PythonExceptions.BytesWarning, "str() on a bytes instance");
+            }
             return _bytes.BytesRepr();
         }
 
--- a/Src/IronPython/Runtime/Bytes.cs
+++ b/Src/IronPython/Runtime/Bytes.cs
@@ -16,6 +16,7 @@ using System.Text;
 using Microsoft.Scripting.Runtime;
 using Microsoft.Scripting.Utils;
 
+using IronPython.Runtime.Exceptions;
 using IronPython.Runtime.Operations;
 using IronPython.Runtime.Types;
 using NotNullWhenAttribute = System.Diagnostics.CodeAnalysis.NotNullWhenAttribute;
@@ -852,6 +853,11 @@ namespace IronPython.Runtime {
                 ),
                 GetType() == typeof(Bytes) ? null : ObjectOps.ReduceProtocol0(context, this)[2]
             );
+        }
+
+        public virtual string __str__(CodeContext context) {
+            PythonOps.Warn(context, PythonExceptions.BytesWarning, "str() on a bytes instance");
+            return _bytes.BytesRepr();
         }
 
         public virtual string __repr__(CodeContext context) {
--- a/Src/IronPython/Runtime/Operations/StringOps.cs
+++ b/Src/IronPython/Runtime/Operations/StringOps.cs
@@ -162,7 +162,6 @@ namespace IronPython.Runtime.Operations {
                 return "None";
             }
             if (x is string) {
-                // check ascii
                 return x;
             }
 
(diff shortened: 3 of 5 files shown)


Src/IronPython/Runtime/ByteArray.cs
```diff
@@ -1078,6 +1079,11 @@ public int rindex(BigInteger @byte, object? start, object? end)
             }
         }
 
+        public virtual string __str__(CodeContext context) {
+            PythonOps.Warn(context, PythonExceptions.BytesWarning, "str() on a bytearray instance");
```
slozier, line 1083: I'm fine with adding these warnings, however it should be noted that they should only occur when you run with the `-b` argument (and raise an exception with `-bb`).

slozier (approved):

Src/IronPython/Runtime/ByteArray.cs
slozier: I took a quick look and it seems that to match CPython we just need to add an `if (context.LanguageContext.PythonOptions.BytesWarning)`. I guess it could also be moved down to `PythonOps.Warn`.

BCSharp: The warnings seem to work just fine at my end. The only thing is that by default, all warnings seem to be sent to `ignore` filter. So to get the same behaviour as CPython, I run `ipy -W default` to reset the warning filter to default. Setting the environment variable `IRONPYTHONWARNINGS` to `default` also works. So I think that IronPython's default warning filter should be `default` and not `ignore`. Any reason why IronPython starts up with `ignore`?

I didn't know about option `-b`, it is not listed in help printed by `ipy -h`; what does it do? Option `-bb` doesn't work at all, says `File -bb does not exist.`

slozier: > The warnings seem to work just fine at my end.

That's the odd bit since I'm not getting this warning unless I explicitly ask for it via the command line. When I run `test_str` from the PR with CPython they fail with:
```
AssertionError: BytesWarning not triggered
```

> Any reason why IronPython starts up with ignore?

It seems to be set up fine on my end. With IronPython:
```
>>> warnings.filters
[('ignore', None, <class 'DeprecationWarning'>, None, 0), ('ignore', None, <class 'PendingDeprecationWarning'>, None, 0), ('ignore', None, <class 'ImportWarning'>, None, 0), ('ignore', None, <class 'BytesWarning'>, None, 0)]
```
and with CPython 3.4:
```
>>> warnings.filters
[('ignore', None, <class 'DeprecationWarning'>, None, 0), ('ignore', None, <class 'PendingDeprecationWarning'>, None, 0), ('ignore', None, <class 'ImportWarning'>, None, 0), ('ignore', None, <class 'BytesWarning'>, None, 0), ('ignore', None, <class 'ResourceWarning'>, None, 0)]
```

> I didn't know about option -b, it is not listed in help printed by ipy -h; what does it do?

From the CPython `-h`:
> -b     : issue warnings about str(bytes_instance), str(bytearray_instance) and comparing bytes/bytearray with str. (-bb: issue errors)

All this option actually does is set `sys.flags.bytes_warning` to the number of `b` (so `-b` gives `sys.flags.bytes_warning == 1`, `-bb` gives `sys.flags.bytes_warning == 2` and so on). It looks like IronPython only supports the `-b` option right now. However as you mention it's not listed in the help.

Anyway, doesn't seem to explain why you're getting the warning and I'm not. Can you check the value of your `sys.flags.bytes_warning`? Maybe there's another way to enable this flag.

BCSharp: Okay, I think I understand the source of the confusion here.

> Anyway, doesn't seem to explain why you're getting the warning and I'm not. Can you check the value of your `sys.flags.bytes_warning`? Maybe there's another way to enable this flag.

It turns out that my `python3.4` command was an a shell function that run `python -b @args`. So all my CPython tests are being run with `BytesWarning` enabled. I must have set it up when I first installed Python 3.4 and now have completely forgotten it. Mystery solved. 

Now about handling of bytes warnings by CPython. It wasn't clear to me that I **have** to enable `BytesWarning` on commandline with `-b` to see them at all. The 3.4 documentation [says](https://docs.python.org/3.4/using/cmdline.html#cmdoption-b):

> Issue a warning when comparing str and bytes. Issue an error when the option is given twice (-bb).

At first I thought it is simply a shortcut for `-W default::BytesWarning` and `-W error::BytesWarning`. After all, `warnings.filters` by default contains `('ignore', None, <class 'BytesWarning'>, None, 0)`. But its more than that. From [What’s New In Python 3.7](https://docs.python.org/3.7/whatsnew/3.7.html#id13):

> Due to changes in the way the default warnings filters are configured, setting `Py_BytesWarningFlag` to a value greater than one is no longer sufficient to both emit `BytesWarning` messages and have them converted to exceptions. Instead, the flag must be set (**to cause the warnings to be emitted in the first place**), and an explicit `error::BytesWarning` warnings filter added to convert them to exceptions.

(Emphasis mine). Perhaps to avoid this confusion, default `('ignore', None, <class 'BytesWarning'>, None, 0)` disappears from the default `warnings.filters` [from 3.7 onwards](https://docs.python.org/3/library/warnings.html?highlight=warnings#default-warning-filter):

> _Changed in version 3.7_: `BytesWarning` no longer appears in the default filter list and is instead configured via `sys.warnoptions` when `-b` is specified twice.

The above is not exactly accurate. The precise behaviour is that `-bb` (or `-b -b`) will add `'error::BytesWarning` to `sys.warnoptions` and insert `('error', None, <class 'BytesWarning'>, None, 0)` at the front of `warnings.filters`, but also a single `-b` will add `'default::BytesWarning'` to `sys.warnoptions` and insert `('default', None, <class 'BytesWarning'>, None, 0)` to `warnings.filters`. In earlier CPython version, of course, `warning.filters` already had a filter for `BytesWarnings`, which then is simply modified from `ignore` to `error` or `default`, respectively. And `sys.warnoptions` is not modified.

Now on to differences between IronPython 3.4 and CPython 3.4.

1. The commandline option `-b` does set `sys.flags.bytes_warning` to 1, but does not update `warnings.filters`. 
2. There is no `-bb` option. Passing `-b -b` works, but `sys.flags.bytes_warning` is still 1, rather than 2 (how else, since `PythonOptions.BytesWarning` is a `bool`). And of course `warnings.filters` is not updated either.
3. As you pointed out, `BytesWarning` should not be emitted at all unless `PythonOptions.BytesWarning` is true (or rather, non-zero, if we change it to `int`). This is best done in `PythonOps.Warn`.
4. When IronPython does print out a warning, the line number reported seems to be always 1, though when the warning is converted to an error (with a filter), the error itself displays the correct line number.

I propose to resolve all this in a separate PR. I am not sure that `PythonOptions.BytesWarning` has to change to `int`, as the role of value 2 seem to be completed once the warning filter list is updated. On the other hand, it would be good for a better CPython compatibility. E.g. `-bbbb` will set `sys.flags.bytes_warning` to 4.

For this PR, I propose I simply make that part of the test that checks for `BytesWarning` run conditionally on `sys.flags.bytes_warning`.

BCSharp: > `BytesWarning` should not be emitted at all unless `PythonOptions.BytesWarning` is true (...). This is best done in `PythonOps.Warn`.

Second thoughts on that. I looked into CPython source code and see that there is only a handful of places where `BytesWarning` is being issued. Basically only for `__str__`, `__eq__`, and `__ne__` on `bytes` or `bytearray`. So the check against `PythonOptions.BytesWarning` can be at the source of the problem rather than in `PythoOps.Warn`, without adding too much complexity (and perhaps keeping `PythonOps.Warn` slightly more efficient).

Also, a `BytesWarning` issued by the user code with `warnings.warn("My bytes warning", BytesWarning)` should be displayed (or not displayed, if filtered out) regardless of the `-b` switch.

I will amend the code in this PR along these lines.

PR #964, opened 2020-10-07
Override DefaultLanguage

slozier:
See discussion in https://github.com/IronLanguages/ironpython3/pull/962

@BCSharp


BCSharp:
Looks great! Thanks!

PR #970, opened 2020-10-10
Make int.from_bytes recognize __bytes__ in all cases

BCSharp:
Fixes:
```python
class BytesBytesSubclass(bytes):
    def __bytes__(self):
        return BytesBytesSubclass(b"xyz")

assert int.from_bytes(BytesBytesSubclass(b"abc"), 'big') == int.from_bytes(b"xyz", 'big')
```
Also:
```python
assert int.from_bytes(memoryview(b"abc")[::2], 'big') == int.from_bytes(b"ac", 'big')
```

### DIFF
--- a/Src/IronPython/Runtime/Bytes.cs
+++ b/Src/IronPython/Runtime/Bytes.cs
@@ -165,6 +165,8 @@ namespace IronPython.Runtime {
                 return (Bytes)o;
             } else if (TryInvokeBytesOperator(context, o, out Bytes? res)) {
                 return res;
+            } else if (o is IBufferProtocol bp) {
+                return new Bytes(bp);
             } else if (o is string || o is ExtensibleString) {
                 throw PythonOps.TypeError("cannot convert unicode object to bytes");
             } else {
--- a/Src/IronPython/Runtime/Operations/IntOps.cs
+++ b/Src/IronPython/Runtime/Operations/IntOps.cs
@@ -528,27 +528,15 @@ namespace IronPython.Runtime.Operations {
             return Bytes.Make(res.ToArray());
         }
 
-        public static BigInteger from_bytes([NotNull] byte[] bytes, [NotNull] string byteorder, bool signed=false) {
-            // TODO: signed should be a keyword only argument
-            // TODO: return int when possible?
-
-            bool isLittle = byteorder == "little";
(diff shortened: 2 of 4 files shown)


slozier (approved):
Not sure if the `from_bytes(byte[], ...)` overload provides much value. Otherwise looks good to me.

Any idea if `from_bytes` is special or are we we going to have to jump through hoops to handle all the cases whenever we have a bytes argument?

BCSharp:
> Not sure if the `from_bytes(byte[], ...)` overload provides much value. 

I wasn't sure either. It is there to avoid one unnecessary data copy on .NET Framework. I did not want to make the reworked version any less efficient than the original for all usage cases. But since you are not sure is is worth it as well, I'd rather simplify it and get rid of the overload. After all, the input size should be small anyway.

> Any idea if `from_bytes` is special or are we we going to have to jump through hoops to handle all the cases whenever we have a bytes argument?

I haven't seen anything like this anywhere else, but then again, I don't have many ideas for where to look. I am currently looking at the constructors `float(bytes)` (implemented) and `int(bytes)` (not implemented), and neither of them does what `int.from_bytes` does. The moment I had discovered that it accepts a strided `memoryview` I knew it had to go though the full-blown `bytes` constructor; there are few places in the CPython codebase that use `FULL_RO` to obtain a buffer from a buffer protocol object. I think that `int.from_bytes` may be a special case. There seem to be no other type that has a `from_bytes` defined, and for `int.from_bytes` itself [CPython had to jump though hoops](https://bugs.python.org/issue23640) to arrive at the current behaviour.

Note that our `int.from_bytes` is not fully compliant yet. Besides the existing points mentioned in the TODO comments, this doesn't work:
```python
class MyInt(int): pass

assert type(MyInt.from_bytes(b"abc", 'big')) == MyInt
```
but I don't know how this can be done. For `str` there is a dedicated `class ExtensibleString` to support subclassing, but I don't see `ExtensibleBigInteger` anywhere.

slozier:
> but I don't know how this can be done. For str there is a dedicated class ExtensibleString to support subclassing, but I don't see ExtensibleBigInteger anywhere.

```C#
[ClassMethod, StaticExtensionMethod]
public static BigInteger from_bytes(CodeContext context, PythonType/*!*/ cls, object bytes, string byteorder, bool signed = false);
```

PR #971, opened 2020-10-10
Enable some float construction tests

BCSharp:


Src/IronPython/Runtime/LiteralParser.cs
```diff
@@ -635,12 +635,17 @@ private static string DoParseString<T>(ReadOnlySpan<T> data, bool isRaw, bool is
...
 
-        private static string ReplaceUnicodeDigits(string text) {
+        private static string ReplaceUnicodeCharacters(string text) {
             StringBuilder replacement = null;
             for (int i = 0; i < text.Length; i++) {
                 if (text[i] >= '\x660' && text[i] <= '\x669') {
+                    // replace unicode digits
                     if (replacement == null) replacement = new StringBuilder(text);
                     replacement[i] = (char)(text[i] - '\x660' + '0');
+                } else if (text[i] >= '\x80' && char.IsWhiteSpace(text[i])) {
```
slozier, line 645: Probably an insignificant optimization, but we could do `var t = text[i];`?

slozier (approved):
Looks good to me.

Src/IronPython/Runtime/LiteralParser.cs
BCSharp: Sure.

PR #973, opened 2020-10-10
Misc changes

slozier:
Resolves the following:
```Python
class MyInt(int): pass

assert type(MyInt.from_bytes(b"abc", 'big')) == MyInt
```
```Python
from _collections import deque
class test:
    def __lt__(self, other):
        return True

assert not (deque([test()]) < deque())
```
This one is a regression in ipy3:
```Python
assert hex(0x800000000) == '0x800000000'
```

### DIFF
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -625,10 +625,16 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
         public static string hex(object? o) {
             object res = PythonOps.Index(o);
             if (res is BigInteger b) {
-                if (b < 0) {
-                    return "-0x" + (-b).ToString("x");
+                if (b == 0) {
+                    return "0x0";
+                } else if (b < 0) {
+                    var val = (-b).ToString("x");
+                    if (val[0] == '0') val = val.TrimStart('0');
+                    return "-0x" + val;
                 } else {
-                    return "0x" + b.ToString("x");
+                    var val = b.ToString("x");
+                    if (val[0] == '0') val = val.TrimStart('0');
+                    return "0x" + val;
                 }
             }
 
--- a/Src/IronPython/Modules/Builtin.cs
+++ b/Src/IronPython/Modules/Builtin.cs
@@ -628,13 +628,9 @@ Noteworthy: None is the `nil' object; Ellipsis represents `...' in slices.";
                 if (b == 0) {
                     return "0x0";
                 } else if (b < 0) {
-                    var val = (-b).ToString("x");
-                    if (val[0] == '0') val = val.TrimStart('0');
-                    return "-0x" + val;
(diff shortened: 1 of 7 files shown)


Src/IronPython/Modules/Builtin.cs
```diff
@@ -625,10 +625,16 @@ public static object eval(CodeContext/*!*/ context, [NotNull]FunctionCode code,
         public static string hex(object? o) {
             object res = PythonOps.Index(o);
             if (res is BigInteger b) {
-                if (b < 0) {
-                    return "-0x" + (-b).ToString("x");
+                if (b == 0) {
+                    return "0x0";
+                } else if (b < 0) {
+                    var val = (-b).ToString("x");
+                    if (val[0] == '0') val = val.TrimStart('0');
```
slozier, line 632: Seems like a bug in .NET. Should probably file an issue and add a comment here.

Src/IronPython/Runtime/Operations/IntOps.cs
```diff
@@ -540,19 +540,16 @@ public static partial class Int32Ops {
...
-            } else {
-                bool msbSet = (bytesArr[0] & 0x80) == 0x80;
-                if (!msbSet) return new BigInteger(bytesArr.Reverse());
-                return new BigInteger(bytesArr.Reverse().Concat(Enumerable.Repeat<byte>(signed ? (byte)0xff : (byte)0, 1)).ToArray());
-            }
+            if (!isLittle) bytesArr = bytesArr.Reverse();
+            bool msbSet = (bytesArr[bytesArr.Length - 1] & 0x80) == 0x80;
+            var val = msbSet
+                ? new BigInteger(bytesArr.Concat(Enumerable.Repeat<byte>(signed ? (byte)0xff : (byte)0, 1)).ToArray())
+                : new BigInteger(bytesArr);
```
BCSharp, line 550: I think if `signed` then there is no need for padding with `0xff` because `msbSet` so the value will already be recognized as negative. Something like:
```C#
var val = msbSet && !signed
    ? new BigInteger(bytesArr.Concat(Enumerable.Repeat<byte>((byte)0, 1).ToArray())
    : new BigInteger(bytesArr);
```
?
Not tested.

Src/IronPython/Runtime/Operations/IntOps.cs
```diff
@@ -540,19 +540,16 @@ public static partial class Int32Ops {
...
 #endif
+            return __new__(context, type, val);
```
BCSharp, line 552: This seems to address also `// TODO: return int when possible?` from line 534

Src/IronPython/Modules/Builtin.cs
BCSharp: It does indeed look like a bug in .NET, both Framework and Core.

BCSharp (approved):

PR #976, opened 2020-10-11
Enable some test_long tests

slozier:


Src/IronPython/Runtime/Operations/IntOps.cs
```diff
@@ -153,6 +153,22 @@ public static partial class Int32Ops {
             return LiteralParser.ParseIntegerSign(s, @base, start);
         }
 
+        [StaticExtensionMethod]
+        public static object __new__(PythonType cls, string s, object @base) {
```
slozier, line 157: @BCSharp since you might have some thoughts about this:

It seems to me like some sort of `[PythonIndex]` attribute would be interesting for this. Signature could be something like:
```C#
public static object __new__(PythonType cls, string s, [PythonIndex] int @base);
```
which could do the proper conversions and throw if necessary. The `PythonIndexAttribute` could define different behavior for the overflow (e.g. throw or clamp).

Also, I used `PythonOps.Index`, but I just noticed there's also a similar `Converter.ConvertToIndex` method. I wonder which one is "better".

BCSharp: While I was working on the public `Bytes` API, I was entertaining the same idea, but never developed it far enough to make a proposal. Having an attribute like `[PythonIndex]` would allow to simplify `Bytes` API considerably, and probably in a number of other places as well. Since .NET objects are limited to 2GB (or 4GB but with caveats), `int` is perfectly fine for most methods. Also, It would create consistency in how index-like arguments are handled. By default, it is probably best if the attribute triggers an exception for values not fitting in `int`, but in some cases, clamping should be done, so an optional parameter for the `PythonIndexAttribute` constructor?

At a time I was also thinking the attribute may be applicable not only to `int` parameters abut to any integer types, but so far never encountered a situation where that was self-evident. The biggest obstacle was that of aligning the exception type and/or message with CPython (`OverflowError` vs. `TypeError` vs. `ValueError`). Maybe an attribute construvtor paramenter would do the trick. Or clever defaults.

I always used `Converter.ConvertToIndex` because it returns an `int` and seems more efficient than `PythonOps.Index`. The latter does all kind of stuff that appears unnecessary, but perhaps I am overlooking something. If indeed `PythonOps.Index` has some added value, maybe it can be incorporated into `Converter.ConvertToIndex`.

slozier: Going to leave this as-is for now, can revisit the idea later.

PR #979, opened 2020-10-13
Use IBufferProtocol in zlib

BCSharp:
Here is an example where `PythonIndexAttribute` would be quite useful, and on a different type than `int`. The `crc` parameter for `binascii.crc32`, (and `value` parameter for `zlib.crc32`) honours `__index__` and accepts any integer value, but for any value not fitting in `UInt32`, the value is masked, i.e. the 32 least significant bits are taken and interpreted as unsigned. I suspect something similar for `zlib.adler32`, though I haven't tesed it. In general, I suspect honouring `__index__` is quite widespread and the error handling differences can be clamping, masking, `OverflowError`, `ValueError`.

----
`zlib.crc32` seems to do the same as `binascii.crc32` so I assume it is OK for it to return `object` as well?

----
It can be risky to convert `IBufferProtocol` to `IList` in a general case by calling the `Bytes` constructor. This constructor is one of the very few places that accept any possible buffer protocol, but in most cases Python API requires a bytes-like object (i.e. C-contiguous). I have noticed that the (old) `binascii.crc32` was doing it, but I haven't checked the whole `binascii`. But since it is a negative noncompliance, I am not too worried about it.

### DIFF
--- a/Src/IronPython.Modules/binascii.cs
+++ b/Src/IronPython.Modules/binascii.cs
@@ -374,7 +374,7 @@ both encoded.  When quotetabs is set, space and tabs are encoded.")]
 
         #region crc32
 
-        [Documentation("crc32(data[, crc]) -> string\n\nComputes a CRC (Cyclic Redundancy Check) checksum of string.")]
+        [Documentation("crc32(data[, crc]) -> string\n\nComputes a CRC (Cyclic Redundancy Check) checksum of data.")]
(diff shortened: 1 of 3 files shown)


slozier (approved):
Looks good to me

> but in most cases Python API requires a bytes-like object

I expect all the `binascii` methods should require bytes-like objects and I used the `Bytes` constructor so I wouldn't have to rewrite all the worker methods.

PR #983, opened 2020-10-14
Use separate test directories per framework

BCSharp:
This is instead marking `test_dynamic_regressions` `NotParallelSafe`. Probably helps in other places too.

### DIFF
--- a/Src/IronPython/Lib/iptest/ipunittest.py
+++ b/Src/IronPython/Lib/iptest/ipunittest.py
@@ -111,7 +111,10 @@ def load_ironpython_test(*args):
 class IronPythonTestCase(unittest.TestCase, FileUtil, ProcessUtil):
 
     def setUp(self):
-        self._temporary_dir = os.path.join(self.temp_dir, "IronPythonTest", clr.TargetFramework.translate(str.maketrans(" ,=", "_--")))
+        if is_cli:
+            self._temporary_dir = os.path.join(self.temp_dir, "IronPythonTest", clr.TargetFramework.translate(str.maketrans(" =,", "__-")))
+        else:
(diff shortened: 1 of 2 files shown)


Src/IronPython/Lib/iptest/ipunittest.py
```diff
@@ -111,16 +111,16 @@ def load_ironpython_test(*args):
 class IronPythonTestCase(unittest.TestCase, FileUtil, ProcessUtil):
 
     def setUp(self):
-        self._temporary_dir = os.path.join(self.temp_dir, "IronPython")
+        self._temporary_dir = os.path.join(self.temp_dir, "IronPythonTest", clr.TargetFramework.translate(str.maketrans(" ,=", "_--")))
```
slozier, line 114: `clr.TargetFramework` is probably fine. My only thought is if we ever produce an `IronPython.dll` assembly which is not in sync with the `exe` (e.g. if we made a `net5.0` exe for testing that references the `netcoreapp3.1` assemblies) then the whole thing breaks down... Although I wonder if we could just run the `netcoreapp3.1` exe using the .NET 5 runtime without producing a different assembly?

BCSharp: My attitude is that we'll fix it when we get there. The whole framework detection logic is pretty complex and hard to guess in advance for future frameworks. For the scenarios you describe `test_env` will most likely have to be updated. We will need something like `is_net50`. Maybe then we can define a dedicated, platform-unique variable that can replace `clr.TargetFramework`.

The good news is that in the worst case two of the frameworks will share the same test directory, which is easily noticeable and still better than what we have today.

slozier: This will break running the tests with CPython (`clr` not defined).

PR #989, opened 2020-10-18
Use IBufferProtocol in compressors and decompressors

BCSharp:
`BZ2Compress.cs` and `BZ2Decompress.cs` are small could be merged into `BZ2Module.cs` (which is practically empty), but I suppose are kept separate to keep the same structure as of `zlib`.


slozier (approved):
LGTM!

PR #990, opened 2020-10-18
Use IBufferProtocol in pyexpat and _multiprocessing

BCSharp:
`_multiprocessing.send` => no data copy

`pyexpat.parse(str)` => eliminated one data copy on .NET Framework
`pyexpat.parse(byteslike)` => eliminated one data copy on .NET Core for input types other than `Bytes` or `byte[]`


slozier (approved):
Looks good to me.

There was a failure in `net46.CPython.test_bytes` on Github actions. Didn't see anything useful in the logs and I don't see how it would be related so not too worried about it. Not sure what's going up with the Github actions macOS test. Rerunning the tests.

Issue #882, opened 2020-06-07
Support `IReadOnlyDictionary` with `ParamDictionaryAttribute`


slozier:
I think this might require changes in the DLR? For example:
```C#
public static PythonList sorted(CodeContext/*!*/ context,
            object? iterable,
            [ParamDictionary, NotNull]IReadOnlyDictionary<string, object> kwArgs);
```

PR #992, opened 2020-10-21
Complete support for bytes paths in nt

BCSharp:


Src/IronPython.Modules/nt.cs
```diff
@@ -817,11 +952,11 @@ public class uname_result : PythonTuple {
         /// <summary>
         /// Convert a sequence of args to a string suitable for using to spawn a process.
         /// </summary>
-        private static string ArgumentsToString(CodeContext/*!*/ context, object args) {
+        private static string ArgumentsToString(CodeContext/*!*/ context, object args, string methodname) {
             IEnumerator argsEnumerator;
             System.Text.StringBuilder sb = null;
             if (!PythonOps.TryGetEnumerator(context, args, out argsEnumerator)) {
-                throw PythonOps.TypeError("args parameter must be sequence, not {0}", DynamicHelpers.GetPythonType(args));
+                throw PythonOps.TypeError("args parameter must be sequence, not {1}", DynamicHelpers.GetPythonType(args));
```
slozier, line 959: Not sure why this changed from `{0}` to `{1}`?

Src/IronPython.Modules/nt.cs
```diff
@@ -1317,36 +1475,45 @@ public sealed class terminal_size : PythonTuple {
...
-                        default:
-                            throw PythonOps.TypeError("'{0}' is an invalid keyword argument for this function", pair.Key);
-                    }
+            foreach (var kvp in kwargs) {
+                switch (kvp.Key) {
+                    case nameof(times):
+                        if (numArgs > 0) throw PythonOps.TypeError("argument for {0}() given by name ('{1}') and position ({2})", nameof(utime), nameof(times), 2);
+                        times = kvp.Value == null ? null
+                                : args[0] is PythonTuple pt2 ? pt2
+                                : throw PythonOps.TypeError("utime: 'times' must be either a 2-value tuple (atime, mtime) or None");
```
slozier, line 1494: Should `args[0]` here be `kvp.Value`? Maybe wrap the logic in a local function:
```C#
static PythonTuple GetTimes(object val) {
    return val == null ? null
        : val is PythonTuple pt ? pt
        : throw PythonOps.TypeError("utime: 'times' must be either a 2-value tuple (atime, mtime) or None");
}
```
I guess we could also do the length check on the tuple in there?

Src/IronPython.Modules/nt.cs
```diff
@@ -1805,6 +1979,32 @@ public static void utime(string path, [ParamDictionary]IDictionary<string, objec
         }
 
 #endif
+        private static string ToFsString(this Bytes b) => _filesystemEncoding.GetString(b.AsMemory().Span);
+
+        private static Bytes ToFsBytes(this string s) => Bytes.Make(_filesystemEncoding.GetBytes(s));
+
+        private static Bytes ToFsBytes(this IBufferProtocol bp) {
+            // TODO: DeprecationWarning starting from Python 3.7
```
slozier, line 1987: I would be fine with adding the deprecation warning right away (if it doesn't cause testing failures).

slozier (approved):
I know it's still marked as draft, but here are my notes.

Src/IronPython.Modules/nt.cs
```diff
@@ -459,30 +506,45 @@ public static PythonList listdir(CodeContext/*!*/ context, [BytesLike][NotNull]I
             }
         }
 
+        public static PythonList listdir(CodeContext context, [NotNull] Bytes path) {
+            PythonList ret = new PythonList();
+            foreach (object name in listdir(context, _filesystemEncoding.GetString(path.AsMemory().Span))) {
+                ret.AddNoLock(Bytes.Make(_filesystemEncoding.GetBytes((string)name)));
```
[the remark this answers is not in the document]
BCSharp: Yeah, a remnant from pre-helpers time.

Src/IronPython.Modules/nt.cs
BCSharp: Wow, no idea why...

Src/IronPython.Modules/nt.cs
BCSharp: OK. In general, I am still working on the keyword/default parameter handling and am musing about some generic, multipurpose mechanism that will cut though so much ceremony that is now required. The `\u00F8` seems not to work anymore since Python 3 allows Unicode letters in identifiers.  But perhaps I will start with something simpler, like #882.

BCSharp:
Thanks for early feedback!

slozier:
I think this may solve some of the failures in `test_glob`.

BCSharp:
I have kept it on hold because some things can get simplified once we have support for keyword-only arguments. But since I am preoccupied with other stuff right now, I can't say when this will be. So if it helps, this PR can be merged as is. I can always revisit it later on.

slozier:
Thanks. I'll merge it in as-is. Better to have incremental progress than wait indefinitely. 😄

PR #999, opened 2020-10-27
Support subclasses of str as kwargs keys

### LINKED ISSUE #998: Keyword dictionaries don't work with keywords that are subclasses of string
Not that we necessarily need to support this, but it works in CPython 😦 ...
```Python
def test(**kwargs): pass

class mystr(str): pass

test(**{mystr("bla"): 1})
```
fails with
> TypeError: test() keywords must be strings

BCSharp:
Resolves #998.


Tests/test_function.py
```diff
@@ -287,6 +287,65 @@ def __init__(self, group=None, target=None):
...
+            self.assertTrue(SplatTest1.FuncWithIDictKwargs(**{"bla": "bla"}))
+            self.assertTrue(SplatTest1.FuncWithIDictGenOOKwargs(**{"bla": "bla"}))
+            self.assertTrue(SplatTest1.FuncWithIDictGenSOKwargs(**{"bla": "bla"}))
+            self.assertTrue(SplatTest1.FuncWithIDictKwargs(**mymapping("bla", "bla")))
+            self.assertTrue(SplatTest1.FuncWithIDictGenOOKwargs(**mymapping("bla", "bla")))
+            self.assertTrue(SplatTest1.FuncWithIDictGenSOKwargs(**mymapping("bla", "bla")))
+
+            # now create the sites in the reverse order of key types
+
+            self.assertTrue(SplatTest2.FuncWithIDictKwargs(**{"bla": "bla"}))
```
slozier, line 335: Hmm, how does creating the sites in the reverse order matter? I guess my knowledge of bindings is lacking...

Tests/test_function.py
```diff
@@ -287,6 +287,65 @@ def __init__(self, group=None, target=None):
 
         foo.__init__(a, target='baz')
 
+    def test_kwargs3(self):
+        """verify keyword args with keys of subtypes of str"""
+
+        class mystr(str): pass
+
+        class mymapping:
```
slozier, line 295: Could wrap a dict instead:
```Python
class mymapping:
    def __init__(self, d):
        self.d = d
    def keys(self):
        return self.d.keys()
    def __len__(self):
        return self.d.__len__()
    def __getitem__(self, x):
        return self.d.__getitem__(x)
```
although maybe not worth it...

Src/IronPython/Runtime/StringDictionaryStorage.cs
```diff
@@ -149,12 +150,17 @@ internal class StringDictionaryStorage : DictionaryStorage {
         public override bool HasNonStringAttributes() {
             if (_data != null) {
                 lock (this) {
-                    if (TryGetObjectDictionary() != null) {
-                        return true;
+                    Dictionary<object, object> dataDict = TryGetObjectDictionary();
```
slozier, line 153: My only concern here is that different `DictionaryStorage` return different answers? For example `CommonDictionaryStorage` would return `true` if it contains `Extensible<string>` elements?

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -2280,7 +2280,9 @@ public static partial class PythonOps {
         public static PythonDictionary CopyAndVerifyDictionary(PythonFunction function, IDictionary dict) {
             foreach (object? o in dict.Keys) {
                 if (!(o is string)) {
-                    throw TypeError("{0}() keywords must be strings", function.__name__);
+                    if (!(o is Extensible<string> es)) {
```
slozier, line 2283: Maybe combine the ifs?

Tests/test_function.py
BCSharp: It does, that's why I threw in [76453cc](https://github.com/IronLanguages/dlr/pull/242/commits/76453ccb4c3934da38c8bb04d8c2a1a1e25b8741) in the DLR. I got tripped on that.

On the first call, when the site is created, the expression to access _kwargs_ by a specific key is cached and reused on subsequent calls with the same key. But the DLR does not really fully support `Extensible<string>` as _kwarg_ keys, now simply treating them as equivalent with `string`. So without 76453cc, if the first call was with `mystr("bla")`, every subsequent call with an equivalent _kwarg_ list, say, plain `"bla"`, would still use `mystr("bla")`, which is weird and unexpected. So I let the expression cache plain `string`, which is the common denominator type for all _kwarg_ keys.

It just shows that fully supporting subclasses of `string` in the DLR is far from trivial. I am going to let `IReadOnlyDictionary<object, something>` be supported in IronLanguages/dlr#240, since this seems to be the intention, but all keys are going to arrive upcast to `string` regardless whether they were originally `string` or `Extensible<string>`.

This particular test case catches a bug that was fixed in IronLanguages/dlr#242. I wanted to have it there in case I (or somebody else) will be tempted in the future to mess around with _kwarg_ support further. And there is still more to finish.

Tests/test_function.py
BCSharp: Yes, good idea, it's cleaner and maybe copy/paste-reusable somewhere else.

Src/IronPython/Runtime/StringDictionaryStorage.cs
BCSharp: Good point. I actually went though all concrete classes that override this method to make sure they are consistent (so no, `CommonDictionaryStorage` would return `false` if it contains only `string` or `Extensible<string>` elements), but forgot the abstract `DictionaryStorage` itself. Will fix it.

Src/IronPython/Runtime/Operations/PythonOps.cs
BCSharp: Sure, the same in `UserMappingToPythonDictionary` I presume.

Tests/test_function.py
slozier: Might be worth including the link to the DLR PR in the comment.

slozier (approved):
Looks good to me.

Issue #791, opened 2020-04-18
sys.stdout.buffer does not get flushed on app termination


slozier:
Running the following file from console does not output anything. 

test.py:
```Python
import sys
sys.stdout.buffer.write(b"Hello world!")
```

This caused a failure in `test_base64`.

slozier:
Same as part of https://github.com/IronLanguages/ironpython3/issues/938

PR #1007, opened 2020-10-31
Flush std files on shutdown

slozier:
Calling flush on the standard file stream is the same approach CPython uses. Resolves https://github.com/IronLanguages/ironpython3/issues/791. Also related to https://github.com/IronLanguages/ironpython3/issues/938.

@jameslan


jameslan:
Yes I think this is the solution!

PR #1011, opened 2020-11-01
Allow arbitrary indexable table with str.translate

slozier:


Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1169,51 +1169,68 @@ public static int rindex([NotNull]this string self, [NotNull]string sub, object?
...
+            if (!PythonTypeOps.TryGetOperator(context, table, "__getitem__", out object getitem)) {
+                throw PythonOps.TypeError($"'{PythonTypeOps.GetName(table)}' object is not subscriptable");
+            }
+
+            StringBuilder ret = new StringBuilder();
+            for (int i = 0; i < self.Length; i++) {
+                var idx = self[i];
+                try {
+                    ret.AppendValueForTranslate(PythonCalls.Call(context, getitem, (int)idx));
+                } catch (Exception e) when (e is LookupException || e is IndexOutOfRangeException || e is KeyNotFoundException) {
```
slozier, line 1229: I can't seem to think of a good way to check for `LookupError` from C# without checking every other derived type. @BCSharp have you run into anything like this before?

BCSharp: I haven't, or better said, if I have, I have missed the complexity altogether.

Tn the code above, if the call throws a user-defined Python error that is a subclass of `LookupError`, `KeyError`, or `IndexError`, would the clause still catch it?

slozier: Hmm, looks like user defined subclasses are not caught with the above. I guess I'd have to add an extra `|| PythonOps.IsInstance(e.GetPythonException(), PythonExceptions.LookupError)` or something like that.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -1226,7 +1226,7 @@ public static int rindex([NotNull]this string self, [NotNull]string sub, object?
                 var idx = self[i];
                 try {
                     ret.AppendValueForTranslate(PythonCalls.Call(context, getitem, (int)idx));
-                } catch (Exception e) when (e is LookupException || e is IndexOutOfRangeException || e is KeyNotFoundException) {
+                } catch (Exception e) when (e is LookupException || e is IndexOutOfRangeException || e is KeyNotFoundException || PythonOps.IsInstance(e.GetPythonException(), PythonExceptions.LookupError)) {
```
BCSharp, line 1229: It appears that `IsInstance` covers the other three tests as well, if you are looking for simplifying the condition.

slozier: Unfortunately it doesn't cover the other three cases if the exception is thrown from C#. The underlying Python exception isn't created until it's needed (so `e.GetPythonException()` can be `null`). It looks like I could use `PythonExceptions.ToPython(e)` which would instantiate the Python exception, but that would make an already super slow code path even slower.

Hopefully code that takes this overload won't have too many misses. In my case it came up in the stdlib where it uses a list that covers every character so I guess it won't be an issue.

BCSharp: I see. I've checked the code parts known to me and I do not see any other places where such catch would be necessary, so perhaps this is an outlier. (Otherwise I was thinking about a static test method somewhere, maybe in `LookupException`).

slozier: Here's one:
```Python
class Test(SystemExit): pass

raise Test
```

Anyway, will merge this as-is for now and file a separate issue.

PR #1014, opened 2020-11-06
Update test_arguments to match DLR

BCSharp:
Matches IronLanguages/dlr#247.

Tests marked with `# msg` may need a review. I think the messages are correct, but a different message (from a different angle) would also be acceptable.

Labels `!!!` or `FIXME` indicate unexpected or erroneous behaviour.

`TODO` means more tests needed.

### DIFF
--- a/Tests/interop/net/method/test_arguments.py
+++ b/Tests/interop/net/method/test_arguments.py
@@ -24,7 +24,7 @@ class ArgumentsTest(IronPythonTestCase):
 
     def test_0_1_args(self):
 
-        # public void M100() { Flag.Reset(); Flag.Set(10); }
+        # public void M100() { Flag.Set(10); }
         f = self.o.M100
         f()
         self.assertRaisesMessage(TypeError, 'M100() takes no arguments (1 given)', lambda: f(1))
@@ -36,7 +36,7 @@ class ArgumentsTest(IronPythonTestCase):
         self.assertRaisesMessage(TypeError, 'M100() takes no arguments (1 given)', lambda: f(**{'x':10}))
         f(*(), **{})
 
-        # public void M200(int arg) { Flag.Reset(); Flag.Set(arg); }
+        # public void M200(int arg) { Flag.Set(arg); }
         f = self.o.M200
         self.assertRaisesMessage(TypeError, "M200() takes exactly 1 argument (0 given)", lambda: f())
         f(1)
@@ -57,7 +57,7 @@ class ArgumentsTest(IronPythonTestCase):
         self.assertRaisesMessage(TypeError, "M200() takes exactly 1 argument (2 given)", lambda: f(arg = 1, other = 2))
         self.assertRaisesMessage(TypeError, "M200() takes exactly 1 argument (2 given)", lambda: f(arg = 1, **{'arg' : 2})) # msg
 
-        # public void M201([DefaultParameterValue(20)] int arg) { Flag.Reset(); Flag.Set(arg); }
+        # public void M201([DefaultParameterValue(20)] int arg) { Flag.Set(arg); }
         f = self.o.M201
         f()
         f(1)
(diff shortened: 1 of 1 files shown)


Tests/interop/net/method/test_arguments.py
```diff
@@ -268,7 +269,7 @@ def test_default_values_2(self):
 
         f(1, **{'y':2}); Flag.Check(3)
 
-        # public void M320([DefaultParameterValue(40)] int y, int x) { Flag.Reset(); Flag.Set(x + y); }
+        # public void M320([DefaultParameterValue(40)] int y, int x) { Flag.Set(x + y); }
```
slozier, line 272: Might be nice to change the definition in the DLR to:
```
public void M320([DefaultParameterValue(40)] int x, int y) { Flag.Set(x * 10 + y); }
```

slozier (commented):
Some of these are tricky to reason about. Seems generally fine.

Tests/interop/net/method/test_arguments.py
BCSharp: Totally agree. Will do it in one of the upcoming PRs for the DLR.

Tests/interop/net/method/test_arguments.py
```diff
@@ -373,7 +381,8 @@ def test_many_args(self):
         f(1, 2, 3, arg9 = 9, arg10 = 10, *(4, 5, 6, 7, 8)); # Flag[str].Check(expect)  # bug 311195
         f(1, 2, 3, arg10 = 10, *(4, 5, 6, 7, 8), ** {'arg9': 9}); # Flag[str].Check(expect) # bug 311195
 
-        self.assertRaisesMessage(TypeError, "M650() got multiple values for keyword argument 'arg5'", lambda: f(2, 3, arg5 = 5, arg10 = 10, *(4, 6, 7, 9), **{'arg8': 8, 'arg1': 1})) # msg (should be 6 given)
+        # msg below, should it be pos 4? (positon of the splatee) but 5 matches CPython
```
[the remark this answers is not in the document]
BCSharp: OK, this makes things simpler too.

BCSharp:
Maybye `test_numbers` could benefit from a longer timeout. It takes about 100 seconds on my Mac and I get the feeling that Azure server is not the fastest either. I can easily imagine that if the system is busy, it takes the test over 120 s. Is Azure macOS on a shared or dedicated host?

BCSharp:
> Some of these are tricky to reason about.

I could write an op-ed on each of the questionable points, but I feel it is a little too early for that. I plan more DLR commits soon. but I wanted this one to be out the door in case someone wants to do development on the latest DLR commit and use IronPyton for testing. So maybe we could review each point later.

One thing that is salient already now and influences many tests is this: given a signature `void F(object x, object y, object z)`, if a call site has `F(1, x=1, z=2)`, does it mean the call has three arguments (as visible) or two arguments (since the first `1` and `x=1` is basically the same argument). This is a fundamental question because the method selection algorithm first filters method candidates based on arity, and only then does other checks, like duplicate/unassigned arguments or type conversion failure. Currently the DLR uses the first interpretation, which most of the time matched CPython, but CPython is not always consistent, and it also changes depending of a release, so I am not putting too much weight into this. Anyway, If we can agree that the first interpretation is preferable, then many questionable cases will get resolved.

slozier:
> Maybe test_numbers could benefit from a longer timeout.

Fine by me, although I wonder why Mono+mac is so slow compared to the other platforms...

> If we can agree that the first interpretation is preferable, then many questionable cases will get resolved.

While IronPython is the main driver for this, the goal of the DLR is to be language agnostic. Checking arity first seems like a reasonable approach so I'm fine with it.

slozier:
@BCSharp Just a heads up. I'm likely to do a DLR release (alongside ipy2) soon after the .NET 5 release (which is on the 10th I assume). So depending on the complexity of the changes I might hold off on merging some things in.

Issue #937, opened 2020-08-29
sys.executable is not executable with Mono


jameslan:
With Mono, `sys.executable` returns `/absolute/path/to/ipy.exe`, which can't be directly executed, but needs to be run by mono: `mono ipy.exe`. This makes `subprocess.Popen([sys.executable])` fail.

On the contrary, with .NET Core, `sys.executable` returns the shell wrapper, which could be launched directly. Thus `subprocess.Popen([sys.executable])` succeeds.

slozier:
We could add some Mono shell wrappers and do a runtime check when setting `sys.executable`.

PR #1020, opened 2020-11-08
Enable some tests on posix

slozier:
They were disabled due to https://github.com/IronLanguages/ironpython3/issues/541 which has since been resolved. Resolves https://github.com/IronLanguages/ironpython3/issues/937


slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/937

PR #1042, opened 2020-11-18
Update tests to expect no binding by keyword to *args/**kwargs

BCSharp:
Matches IronLanguages/dlr#250.


slozier (approved):
Looks good to me. This means that we can remove the unicode char from all the arguments that use the `ParamDictionaryAttribute`?

BCSharp:
That's right. I've seen it also being used sometimes on parameters with `ParamArrayAttribute` (e.g. `BaseException.__init__`), that could go now too.

PR #1060, opened 2020-11-29
Enable test_stdconsole

slozier:


Tests/test_stdconsole.py
```diff
@@ -48,7 +48,7 @@ def setUp(self):
         # This is relative to working directory so the path related tests (e.g.'print __name__')
         # return predictable results.
 
-        self.tmpdir = "tmp" #os.path.join(self.temp_dir, "tmp")
+        self.tmpdir = os.path.join(self.temp_dir, "tmp")
```
BCSharp, line 51: A specific reason for using `self.temp_dir` here?
Using it keeps this test not-parallel safe (unlike `self.temporary_dir`, which is segregated per framework).

Tests/test_stdconsole.py
```diff
@@ -48,7 +48,7 @@ def setUp(self):
         # This is relative to working directory so the path related tests (e.g.'print __name__')
```
BCSharp, line 48: This is no longer true? (since `self.temp_dir` is used)

Tests/test_stdconsole.py
slozier: No specific reason beyond that's what was commented. I'll see if I can get rid of `self.temp_dir`...

slozier:
I feel a bit bad enabling test_stdconsole case since it's so slow to run...

BCSharp:
1.5 minutes over here, not too bad.

Well, maybe it's time to look into the startup performance. We are sometimes discussing variations with a marginal performance impact, while there is this massive cascade of CLR exceptions right at the beginning of every ipy run.

Or maybe a better time for this would be after the first alpha release. 😄

slozier:
Almost 10 minutes during CI.

Indeed, startup performance is something that's been bugging me. Maybe having this slow test running will be an incentive to try and speed things up... We could try and mitigate some of the exception overhead by making some of them light exceptions, but I'm guessing what really needs to happen is fixing up the import system so we don't need to rely on `importlib` anymore.

But yeah, right now I'm just aiming for something that works. 😏

PR #1066, opened 2020-12-01
Update name binding code and comments

BCSharp:
These are some updates that did not make it into #1054, mainly because they were not needed for the implementation of `nonlocal`. Hopefully this will make the name binding code a little more readable and resilient.


slozier (approved):
Thanks!

Issue #508, opened 2018-12-03
test_memory fails intermittently


slozier:
Investigate why `test_memory` fails intermittently (frequently on CI).

PR #1077, opened 2020-12-13
Enable test_memory

slozier:
The issue is that the `PythonType._subtypes` list (which is used for the `__subclasses__` method) is growing every time a new subclass is created and not getting cleaned up as types get GCd. This is not actually a regression with ipy2, if you swap the class definitions with new-style classes then it also fails there.

I'm not super happy with the perf impact of this solution but I'm not sure what else to do (other than adding a finalizer to PythonType). @BCSharp any other ideas?

Resolves https://github.com/IronLanguages/ironpython3/issues/508

### DIFF
--- a/Src/IronPython/Runtime/Types/PythonType.cs
+++ b/Src/IronPython/Runtime/Types/PythonType.cs
@@ -40,7 +40,7 @@ type(name, bases, dict) -> creates a new type instance with the given name, base
         private PythonTypeAttributes _attrs;                // attributes of the type
         private int _flags;                                 // CPython-like flags on the type
         private int _version = GetNextVersion();            // version of the type
-        private List<WeakReference> _subtypes;              // all of the subtypes of the PythonType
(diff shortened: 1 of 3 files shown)


BCSharp:
Yeah, I am not a big fan of using finalizers for stuff like this. They come with their own performance cost. But in this case, I assume creating new types happens far more frequently than getting them garbage collected, so I guess that using finalizers could be justified.

Another idea: keep the current `_subtypes` list cleanup in `AddSubType`, but instead of doing it every new type creation, do it only every 16 or so. Or when the list is about to expand its capacity. Or a combination of both, since both have some unique advantages. Just doing this would reduce the performance impact by an order of magnitude, without any serious memory penalty. I suppose most types will have a short list of user-defined subtypes anyway, with the exception of `object` of course.

BCSharp:
Another additional trigger for the cleanup could be `gc.collect` (unless it is already happening somehow).

PR #1079, opened 2020-12-14
Use qualified module name from import statement during importing

### LINKED ISSUE #936: pip install fails with SystemError
For example, pip installing `pyyaml` fails with:
> SystemError: Parent module 'pkg_resources._vendor.packaging' not loaded, cannot perform relative import

I've attached a "simple" repro of the issue:
[ipy3_import.zip](https://github.com/IronLanguages/ironpython3/files/5111237/ipy3_import.zip)

BCSharp:
Resolves #936.

I still want to add tests.


slozier (approved):
Looks good to me! (although ArraySegment indexing sure is ugly)

PR #1083, opened 2020-12-20
Support rewriting of comprehension scopes

### LINKED ISSUE #817: Name fails to resolve in generator's if when inside comprehension
The following fails:
```Python
{i: tuple(j for j in t if i != j) for t in ((1,2),) for i in t}
```
with
> NameError: global name 'i' is not defined

Causes `import sre_compile` to fail.

Workaround is to use a comprehension instead of a generator:
```Python
{i: tuple([j for j in t if i != j]) for t in ((1,2),) for i in t}
```

BCSharp:
Resolves #817.

Redesigning the inheritance hierarchy would have been a more elegant solution, but that perhaps for another time.

Credit for the tests to the work/discussion done under #1073.

@slozier @isaiah

### DIFF
--- a/Src/IronPython/Compiler/Ast/PythonAst.cs
+++ b/Src/IronPython/Compiler/Ast/PythonAst.cs
@@ -699,7 +699,7 @@ namespace IronPython.Compiler.Ast {
 
         /// <summary>
         /// Rewrites the tree for performing lookups against globals instead of being bound
-        /// against the optimized scope.  This is used if the user compiles optimied code and then
+        /// against the optimized scope.  This is used if the user compiles optimized code and then
         /// runs it against a different scope.
         /// </summary>
         internal PythonAst MakeLookupCode() {
@@ -739,7 +739,7 @@ namespace IronPython.Compiler.Ast {
                     return PythonAst._globalContext;
                 }
 
-                // we need to re-write nested scoeps
+                // we need to re-write nested scopes
                 if (node is ScopeStatement scope) {
                     return base.VisitExtension(VisitScope(scope));
                 }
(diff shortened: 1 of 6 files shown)


Src/IronPython/Compiler/Ast/PythonAst.cs
```diff
@@ -752,6 +752,10 @@ internal class LookupVisitor : MSAst.ExpressionVisitor {
                     return base.VisitExtension(new GeneratorExpression((FunctionDefinition)VisitScope(generator.Function), generator.Iterable));
                 }
 
+                if (node is Comprehension comprehension) {
+                    return VisitComprehension(comprehension);
```
slozier, line 756: Any reason for putting the `base.VisitExtension` inside `VisitComprehension` instead of here like is done with `VisitScopes`? I'm not particularly concerned about it, the difference just caught my eye.

Src/IronPython/Compiler/Ast/PythonNameBinder.cs
```diff
@@ -250,6 +250,7 @@ internal class PythonNameBinder : PythonWalker {
 
         public override bool Walk(ListComprehension node) {
             node.Parent = _currentScope;
+            node.Scope.Parent = _currentScope;
```
slozier, line 253: Why is `ListComprehension` different from `SetComprehension` and `DictionaryComprehension`?

slozier (commented):
Thanks for fixing up the scope rewriting, much appreciated.

Src/IronPython/Compiler/Ast/PythonAst.cs
BCSharp: Because a scope comprehension does not "actually show up in the AST hierarchy and instead hangs off the comprehension expression", to quote the comment from the source code.

It has to be this way so that the comprehension (and all its expressions, including e.g. a generator item) will be rewritten within the comprehension scope, rather than the parent scope.

Oddities like this would have been ironed out with a redesign of the inheritance hierarchy.

Src/IronPython/Compiler/Ast/PythonNameBinder.cs
BCSharp: Because I forgot to clean it up. Thanks for the catch.

slozier (approved):

PR #1086, opened 2021-01-12
Class variable

BCSharp:
This is another take on #723 (an alternative to #734). As such it partially undoes #721 (parser changes). 

It is still a hack, because `__class__` variables are now manages as global variables named `$__class__NN`, not an ideal situation, but it is meant as a proof-of-concept for how a `__class__` variable should behave. In particular, it seems that a `__class__` variable behaves as a regular class scope variable that just happen to be initialized to the class type. For instance, it is possible to set `__class__` to any value, even if it is not a type (e.g. it is possible to set it to a number), while setting the **attribute** `__class__` is guarded against invalid values.

Alternative implementations to global variables:
1. use an ordinary local variable as `__class__`, placed in the class scope and lifted to be a closure cell; this is my favourite, but I was unable to get it work (2ac9d8ce)
2. use a closure from the parent scope to place `__class__`; it works OK but only for non-top-level classes and still pollutes the namespace, this time the local one
3. use a separate closure from the parent closure, which is managed by the class scope itself; since 2. works, this should be possible as well, but I was unable to get it to work (with the same error as with 1., so maybe I am overlooking something) (2d954769)
4. use a separate dedicated storage that is out of sight, perhaps as part of the context

@slozier @isaiah @in-code-i-trust

### DIFF
--- a/Src/IronPython/Compiler/Ast/ClassDefinition.cs
+++ b/Src/IronPython/Compiler/Ast/ClassDefinition.cs
@@ -32,6 +32,7 @@ namespace IronPython.Compiler.Ast {
 
         private static readonly MSAst.ParameterExpression _parentContextParam = Ast.Parameter(typeof(CodeContext), "$parentContext");
         private static readonly MSAst.Expression _tupleExpression = MSAst.Expression.Call(AstMethods.GetClosureTupleFromContext, _parentContextParam);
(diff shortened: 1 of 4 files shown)


slozier:
Closing this in favor of https://github.com/IronLanguages/ironpython3/pull/1087.

Issue #722, opened 2020-01-14
__class__ does not work when the value of variable holding the class is replaced


slozier:
Another busted scenario I encountered while looking into this:
```Python
class two(object):
    def test(self):
        return __class__

a = two
two = "z"
assert a().test() == a().__class__
```

_Originally posted by @slozier in https://github.com/IronLanguages/ironpython3/issues/451#issuecomment-437641168_

Issue #1085, opened 2021-01-06
Assign to nonlocal __class__ variable in instance method causes stack overflow


isaiah:
```python
class A():
    def f(self):
        nonlocal __class__
        __class__ = A

if __name__ == '__main__':
    A().f()
```
Execute the script will result in
```
Stack overflow.
Repeat 174476 times:
--------------------------------
   at IronPython.Compiler.Ast.PythonReferenceVariable.get_LimitVariable()
--------------------------------
   at IronPython.Compiler.Ast.ScopeStatement.GetVariableExpression(IronPython.Compiler.Ast.PythonVariable)
   at IronPython.Compiler.Ast.FunctionDefinition.CreateFunctionLambda()
   at IronPython.Compiler.Ast.FunctionDefinition.EnsureFunctionLambda()
   at IronPython.Compiler.Ast.FunctionDefinition.GetLambda()
...
```

This also causes the `CPython.test_super` test to crash.

BCSharp:
The current implementation of `__class__` is a hack (see #721, #723) that does fail in some not so straightforward cases. There has been one attempt on a proper implementation (#734), so far unsuccessful. I looked into it too, several weeks ago, and I believe I was on the right track but I run out of time before I could finish it. Perhaps I should have commited the unfinished works anyway, if not to just contribute to the discussion about a proper solution.

isaiah:
Thanks for pointing the direction. AFAIK, cpython generates a new scope to encapsulate the `__class__` cell when either `super` or `__class__` is found in the class body. The solution proposed by #734 doesn't really scale, as it assumes the class scope is the grand parent. If you don't mind, maybe you can share your work and I can try to pick it up?

BCSharp:
The idea I had been pursuing was based on an observation that in CPython, `__class__` used within the class scope and all nested scopes (like member methods) behaves like an ordinary variable that just happen to be initialized with the value of the class type. This is remarkably different than the `__class__` **attribute** on the same type. For instance, you can assign anything to the former, while with the latter only values of a workable type are accepted. Whether this is intentional or not, I don't know, but this is how it works in CPython.

So the solution was to introduce an extra variable `ClassVariable` in `ClassDefinition`, initialized with name `__class__` and value the type of the class; almost identical to the currently existing `PythonVariable` (which also exists in `FunctionDefinition`), ecxept that `ClassVariable` lives in the class scope while `PythonVariable` in the parent scope and represents the actual class name.

[...]

PR #1087, opened 2021-01-27
PEP-3135 __class__ cell

isaiah:
This is another attempt to fix the `__class__` implicit free variable in class body.

It follows the CPython implementation. When usage of either `__class__` or `super` is found in the class body, a cell variable named `__class__` is defined. When class body is executed at class definition time, a `__classcell__` local variable is injected into the class namespace, this variable is not visible to python world, as it point to the raw cell object of `__class__`. Then the namespace is passed to `type.__new__`, where it will check the existence of `__classcell__` and inject the newly created type instance into the cell.

Resolves https://github.com/IronLanguages/ironpython3/issues/722, https://github.com/IronLanguages/ironpython3/issues/723 and https://github.com/IronLanguages/ironpython3/issues/1085.


Src/IronPython/Compiler/Ast/ClassDefinition.cs
```diff
@@ -194,11 +208,17 @@ public class ClassDefinition : ScopeStatement {
             var createLocal = CreateLocalContext(_parentContextParam);
 
             init.Add(Ast.Assign(LocalCodeContextVariable, createLocal));
+            // __classcell__ == ClosureCell(__class__)
+            if (needClassCell) {
+                MSAst.Expression assignClassCell = AssignValue(GetVariableExpression(ClassCellVariable), GetVariableExpression(ClassVariable)) ;
```
isaiah, line 213: @BCSharp here I need your help, the plan is to do:
```python
__classcell__ = ClosureCell(__class__)
```

but it turns into:

```python
__classcell__ = __class__
```

is there anyway to get the `ClosureCell` object from the context?

isaiah:
@BCSharp thanks for showing me #1086 , as you said it's still a hack, I'm wondering if we can use the same implementation as in cpython to fix it properly, but I got stuck. Please let me know if you have questions regarding the mechanism.

Tests/test_dict.py
```diff
@@ -55,10 +55,12 @@ def __setitem__(self, *args):
                 super(MyDict, self).__setitem__(*args)
 
         a = MyDict()
-        a[0] = 'abc'
-        self.assertEqual(a[0], 'abc')
-        a[None] = 3
-        self.assertEqual(a[None], 3)
+        with self.assertRaises(SystemError): # TODO: remove assertRaises when https://github.com/IronLanguages/ironpython3/issues/456 is fixed
```
slozier, line 58: We should not be reverting this change since the PR is supposed to fix it.

Tests/test_regressions.py
```diff
@@ -999,7 +999,16 @@ def __init__(self):
             super(two, self).__init__()
             self.cnt += 1
 
-        self.assertEqual(two().cnt, 3)
+        if is_cli:
```
slozier, line 1002: We should not be reverting this change since the PR is supposed to fix it.

Tests/test_regressions.py
```diff
@@ -1443,69 +1452,4 @@ def run(self):
             self.assertEqual(thread.thread_executed, 1)
         test_thread()
 
-    def test_ipy3_gh451(self):
```
slozier, line 1446: I think we should keep all these tests?

Src/IronPython/Runtime/Types/PythonType.cs
```diff
@@ -264,22 +264,41 @@ internal PythonType(PythonContext context, PythonType[] baseTypes, Type underlyi
...
                 // creation.                   
-                return meta.CreateInstance(context, name, bases, classdict);
-            }
+                type = meta.CreateInstance(context, name, bases, classdict);
+            } else {
+                // no custom user type for __new__
+                type = new PythonType(context, name, bases, dict, selfNames);
+            }
+            object cell = dict.get("__classcell__");
+            Console.WriteLine("--- class cell ---");
```
slozier, line 285: Looks like a bunch of testing code here?

Src/IronPython/Compiler/Ast/PythonNameBinder.cs
```diff
@@ -813,28 +813,10 @@ internal class PythonNameBinder : PythonWalker {
             node.Parent = _currentScope;
 
             if (node.Target is NameExpression nameExpr && nameExpr.Name == "super" && _currentScope is FunctionDefinition func) {
+                _currentScope.Reference("__class__");
```
slozier, line 816: Should this go inside the if?

slozier (commented):
Thanks! Just some quick notes about the PR.

Src/IronPython/Compiler/Ast/ClassDefinition.cs
BCSharp: Initializing `__class__` during `__new__` is an interesting approach, I haven't tried it. 

I don't think you can simply do `__classcell__ = ClosureCell(__class__)` at this point, but you can set up a closure for holding `__class__` upfront, this is what I did in 2d95476.

BCSharp:
I agree with @slozier that no existing tests should be disabled. We already have a workaround, so this time the aim is to have a proper implementation. It is OK if it takes several PRs to implement this functionality completely, but regressions are not OK.

isaiah:
@slozier @BCSharp Thanks for the review, the changes in the tests are not intentional, I'll add them back once I got the proof of concept working.

Regarding debugging, besides setting breakpoints, is there a way to check the expression tree after `reduce`? Or do you use decompiler such as `Reflector` and `ILSpy`?

BCSharp:
>  is there a way to check the expression tree after `reduce`

I don't know any, but would gladly hear about it if you find something

Src/IronPython/Compiler/Ast/PythonNameBinder.cs
isaiah: Yes, the free variable should only be created when either `super` or `__class__` is referenced in the class body.

slozier: I guess I should have been more specific. I meant the if below. 😄

isaiah: I see. I think you are right, though this is from previous code, it was like this before the "hack", not sure if we should touch it :)

isaiah:
@BCSharp @slozier Please take another look, I think it's working. Also the example in https://github.com/IronLanguages/ironpython3/issues/1085 is no longer crashing.

isaiah:
> I don't know any, but would gladly hear about it if you find something
I used ILSpy to check the DLL generated by `ipyc`, it works very well, and also it's cross platform.

Src/IronPython/Compiler/Ast/PythonNameBinder.cs
slozier: I checked this against CPython and the way you have it is correct. Calling `super` with arguments still adds `__class__` to the scope:
```Python
class test:
    def test(self):
        assert "__class__" in dir()
        super(test, self).__init__() # calling super with arguments still adds __class__ to the local scope

test().test()
```

Src/IronPython/Compiler/Parser.cs
```diff
@@ -1939,29 +1939,9 @@ public class Parser : IDisposable { // TODO: remove IDisposable
...
                     string name = (string)t.Value;
-                    ParserSink?.StartName(GetSourceSpan(), name);
-                    var fixedName = FixName(name);
-                    if (ShouldReplaceClassKeyword(fixedName)) {
-                        // if `__class__` variable is used in a class method, replace with `name_of_the_class`.
-                        ret = new NameExpression(CurrentClass.Name);
-                    } else {
-                        ret = new NameExpression(fixedName);
-                    }
+                    _sink?.StartName(GetSourceSpan(), name);
```
slozier, line 1942: This should be `ParserSink`
```suggestion
                    ParserSink?.StartName(GetSourceSpan(), name);
```

Src/IronPython/Compiler/Ast/Comprehension.cs
```diff
@@ -262,7 +262,7 @@ internal class ComprehensionScope : ScopeStatement {
             return false;
         }
 
-        internal override PythonVariable BindReference(PythonNameBinder binder, PythonReference reference) {
+       internal override PythonVariable BindReference(PythonNameBinder binder, PythonReference reference) {
```
slozier, line 265: Fix the change in indent?
```suggestion
        internal override PythonVariable BindReference(PythonNameBinder binder, PythonReference reference) {
```

Src/IronPython/Compiler/Ast/ClassDefinition.cs
```diff
@@ -194,11 +208,18 @@ public class ClassDefinition : ScopeStatement {
...
+                var exp = (ClosureExpression) GetVariableExpression(ClassVariable);
+                MSAst.Expression assignClassCell = AssignValue(GetVariableExpression(ClassCellVariable), exp.ClosureCell);
+                init.Add(assignClassCell);
+            }
 
             List<MSAst.Expression> statements = new List<MSAst.Expression>();
             // Create the body
             MSAst.Expression bodyStmt = Body;
 
+
```
slozier, line 222: Remove the extra line that got added?

Src/IronPython/Compiler/Ast/FunctionDefinition.cs
```diff
@@ -626,18 +626,6 @@ private class FunctionDefinitionInstruction : Instruction {
 
             CreateFunctionVariables(locals, init);
 
-            // If the __class__ variable is used in a class method then we need to initialize it.
-            // This must be done before parameter initialization (in case one of the parameters is called __class__).
-            ClassDefinition parent = FindParentOfType<ClassDefinition>();
```
slozier, line 631: I think this was the last call to `FindParentOfType` so maybe we can remove the method?

Src/IronPython/Runtime/Types/PythonType.cs
```diff
@@ -263,22 +263,31 @@ internal PythonType(PythonContext context, PythonType[] baseTypes, Type underlyi
...
-
-                // we have the right user __new__, call our ctor method which will do the actual
-                // creation.                   
-                return meta.CreateInstance(context, name, bases, classdict);
+            } else {
+                // no custom user type for __new__
+                type = new PythonType(context, name, bases, dict, selfNames);
+            }
+            object cell = dict.get("__classcell__");
+            if (cell != null && cell is ClosureCell pycell)
```
slozier, line 284: Null check is unnecessary:
```suggestion
            if (cell is ClosureCell pycell)
```

Src/IronPython/Compiler/Ast/ClassDefinition.cs
```diff
@@ -80,6 +80,9 @@ public class ClassDefinition : ScopeStatement {
         /// </summary>
         internal PythonVariable ModuleNameVariable { get; set; }
 
+        internal PythonVariable ClassCellVariable { get;set; }
+        internal PythonVariable ClassVariable { get;set; }
```
slozier, line 84: Make these private unless there's a reason not to:
```suggestion
        private PythonVariable ClassCellVariable { get; set; }
        private PythonVariable ClassVariable { get; set; }
```

Src/IronPython/Compiler/Ast/ClassDefinition.cs
```diff
@@ -91,9 +94,20 @@ public class ClassDefinition : ScopeStatement {
 
         internal override bool ExposesLocalVariable(PythonVariable variable) => true;
 
+        internal bool needClassCell {get; set;}
```
slozier, line 97:
```suggestion
        private bool needClassCell { get; set; }
```

slozier (commented):
This is more of a high level review since I'm not super familiar with this part of the codebase.

Since this resolves https://github.com/IronLanguages/ironpython3/issues/722 and https://github.com/IronLanguages/ironpython3/issues/723 it would be nice to add some tests.

slozier:
I guess a test for your issue https://github.com/IronLanguages/ironpython3/issues/1085 would be good as well. 😄

Src/IronPython/Compiler/Ast/FunctionDefinition.cs
isaiah: Good catch

Tests/test_regressions.py
```diff
@@ -1211,6 +1211,33 @@ def __eq__(self,other):
...
+        a = two
+        two = "z"
+        self.assertTrue(a().test() == a().__class__, "class name doesn't match")
+
+    def test_gh723(self):
+        import ast
+        x = """
+        class test(object):
+            def test(self):
+                return __class__
```
slozier, line 1228:
```suggestion
class test(object):
    def test(self):
        return __class__
```

isaiah:
@slozier Do you think we can merge this PR?

slozier:
@isaiah Yeah, I want to take another look at it before I do (later today hopefully), but it looks good to me.

slozier:
Thanks for the PR!

BCSharp:
@isaiah I'm a little late to the party, but this is nice progress; I had no additional comments. Good job!

PR #1088, opened 2021-02-03
Add rich comparison operators to number types

slozier:
also does some cleanup of `__cmp__`.

### DIFF
--- a/Src/IronPython.Modules/_datetime.cs
+++ b/Src/IronPython.Modules/_datetime.cs
@@ -1417,7 +1417,7 @@ namespace IronPython.Modules {
             #region Rich Comparison Members
 
             /// <summary>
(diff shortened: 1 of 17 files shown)


slozier:
`IronPython.test_stdmodules` fails because of the following:
```Python
import locale
locale.strxfrm("a") < locale.strxfrm("b")
```
The reason for the failure is that `locale.strxfrm` currently returns a `System.Globalization.SortKey` object and the comparison operators no longer work since `__cmp__` is gone.

Possible solutions:
1. Change `locale.strxfrm` to return a string like CPython.
2. Wrap `SortKey` in something that's comparable.
3. Somehow map rich comparison operators to `Compare`. I'm not sure how many other types would actually benefit from this.

I'm currently leaning toward the 2nd solution.

PR #1091, opened 2021-02-08
Remove fallback comparison using __cmp__

slozier:


slozier:
`CharOps.__cmp__` which did `char`/`string` comparisons should probably be replaced with rich comparisons.

PR #1094, opened 2021-02-11
Fix _collections.deque comparisons

slozier:


slozier:
Note to self: can't just compare the underlying arrays directly.

PR #1123, opened 2021-03-07
Use unique context param for ComprehensionScope

### LINKED ISSUE #1119: NullReferenceException in comprehension expression
The following code raises a `NullReferenceException`:
```Python
x = [[[1]]]
[[tuple(str(d) for d in z) for z in y] for y in x]
```
> SystemError: Object reference not set to an instance of an object.

The failure above is extracted from `test_calendar` (which we don't run during CI). I ran a git bisect and looks like the failure started occurring with https://github.com/IronLanguages/ironpython3/pull/1073.

BCSharp:
Fixes #1119

### DIFF
--- a/Src/IronPython/Compiler/Ast/Comprehension.cs
+++ b/Src/IronPython/Compiler/Ast/Comprehension.cs
@@ -219,8 +219,9 @@ namespace IronPython.Compiler.Ast {
     /// this doesn't actually show up in the AST hierarchy and instead hangs off the comprehension expression.
     /// </summary>
     internal class ComprehensionScope : ScopeStatement {
+        private static int _comprehensionId;
+        private readonly MSAst.ParameterExpression _compContext = Ast.Parameter(typeof(CodeContext), "$compContext$" + System.Threading.Interlocked.Increment(ref _comprehensionId));
         private readonly Comprehension _comprehension;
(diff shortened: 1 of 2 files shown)


slozier:
I wonder what makes `ComprehensionScope` different from all the other `ScopeStatement` which use a static.

BCSharp:
Good wonder.

As understand it: both `ClassDefinition` and `FunctionDefinition` track their parent context, which is then used to create the local context (if needed). In case of the function, it goes though `internal class GetParentContextFromFunctionExpression` and eventually `PythonOps`. But `ComprehensionScope` doesn't do that and just uses `Parent.LocalContext` so in case of nested comprehensions there is a parameter name ambiguity. I guess it's because comprehensions were supposed to be light-weight?

I went here for a simple solution that works (I believe), but more elaborate solutions are possible too, of course. One would need to track the expression that represents the parent context of a comprehension.

Ideally, the whole `ComprehensionScope` could be redesigned and properly hooked up in the scope hierarchy.

slozier:
Thanks the the explanation and the PR!

isaiah:
Maybe this is already known, but just for the sake of clearance, the problem with comprehension is that it's an expression, before that only statements can hold a scope, that's why the ScopeStatement is a subclass of statement, and when comprehension requires a scope, we have a problem. Now the ComprehensionScope hangs to the actual comprehension, instead of being part of the tree.

PR #1127, opened 2021-03-10
Use expectedFailure instead of commenting out tests

slozier:


BCSharp:
This is a very good idea. However, sometimes the choice is not binary (ie. between expected pass and expected failure). Eg. `testLeaks` from `test_scope` sometimes passes, sometimes fails, depending on system load.

slozier:
Indeed. I'm perfectly happy to just comment out tests that are flaky or where the pass/fail matrix is too complex.

Issue #30, opened 2014-03-21
PEP 3155 -- Qualified name for classes and functions


jdhardy:
http://www.python.org/dev/peps/pep-3155/

PR #1128, opened 2021-03-11
Misc changes

slozier:
Some misc changes for 3.5. Also adds `__qualname__` to `function` and `type` (part of https://github.com/IronLanguages/ironpython3/issues/30)

Just for the sake of clarity, these are the changes that are 3.5 specific:
- `_ast.AsyncFunctionDef` (which is just a stub)
- `pow(-1, -2, 3)` throws a `ValueError` instead of a `TypeError`

Everything else should be applicable to 3.4.

### DIFF
--- a/Src/IronPython.Modules/_ctypes/_ctypes.cs
+++ b/Src/IronPython.Modules/_ctypes/_ctypes.cs
@@ -187,8 +187,10 @@ namespace IronPython.Modules {
             NativeFunctions.FreeLibrary(handle);
         }
 
-        private static object LoadDLL(string library, int mode) {
-            if (library.IndexOf((char)0) != -1) throw PythonOps.ValueError("embedded null byte");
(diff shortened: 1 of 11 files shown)


BCSharp:
Are we targeting 3.5 now?

slozier:
@BCSharp Not yet, however it seems like a lot of packages are dropping support for 3.4. So I'm looking at what sort of impact going to 3.5 would have and bringing in (hopefully) harmless changes to facilitate the eventual transition. If it wasn't for `async` (which I have no idea how to address) we would likely have targeted a more recent version from the get-go.

BCSharp:
Thanks for the explanation. I was wondering which version should I be using to verify the tests. I suppose, given the situation, the best would be to use both 3.4 and 3.5. If there are differences in behaviour, it would be visible which way IronPython goes.

PR #1130, opened 2021-03-13
Evaluate first comprehension iterator in outer scope

### LINKED ISSUE #809: dir does not work in list comprehension
For example:
```Python
assert "__builtins__" in [x for x in dir()]
```

Also fails in other comprehensions (e.g. `{x for x in dir()}`) and with other functions (e.g. `locals`). May be related to https://github.com/IronLanguages/ironpython3/issues/562.

This is causing `pickle.__all__` to not opcodes and so causes some test failures.

Edit: also causes `pickletools` to fail on import.

BCSharp:
Resolves #809

### DIFF
--- a/Src/IronPython/Compiler/Ast/PythonAst.cs
+++ b/Src/IronPython/Compiler/Ast/PythonAst.cs
@@ -720,6 +720,7 @@ namespace IronPython.Compiler.Ast {
 
         internal class LookupVisitor : MSAst.ExpressionVisitor {
             private readonly MSAst.Expression _globalContext;
+            private readonly Dictionary<MSAst.Expression, ScopeStatement> _outerComprehensionScopes = new();
             private ScopeStatement _curScope;
 
             public LookupVisitor(PythonAst ast, MSAst.Expression globalContext) {
@@ -737,6 +738,12 @@ namespace IronPython.Compiler.Ast {
             protected override MSAst.Expression VisitExtension(MSAst.Expression node) {
                 if (node == _globalContext) {
                     return PythonAst._globalContext;
+                }
+
+                // outer comprehension iterable is visited in outer comprehension scope
+                if (_outerComprehensionScopes.TryGetValue(node, out ScopeStatement outerComprehensionScope)) {
+                    _outerComprehensionScopes.Remove(node);
+                    return VisitComprehensionIterable(node, outerComprehensionScope);
                 }
 
                 // we need to re-write nested scopes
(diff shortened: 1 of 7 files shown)


Src/IronPython/Compiler/Ast/PythonAst.cs
```diff
@@ -815,14 +822,27 @@ internal class LookupVisitor : MSAst.ExpressionVisitor {
 
                 ScopeStatement prevScope = _curScope;
                 try {
-                    // rewrite the comprehension in a new scope
+                    // mark the first (outermost) "for" iterator for rewrite in the current scope
+                    _outerComprehensionScopes[((ComprehensionFor)comprehension.Iterators[0]).List] = _curScope;
```
slozier, line 826: The first element of `Iterators` is always a `ComprehensionFor`? Probably need to revisit that Ast at some point...

slozier (approved):
Looks good to me! Plus I learnt something new (didn't know this is how comprehensions worked). :smile:

Src/IronPython/Compiler/Ast/PythonAst.cs
BCSharp: Yes, the first element of `Iterators` is always a `ComprehensionFor`. Furthermore, there is always at least one element in `Iterators`. The [parser already ensures it](https://github.com/IronLanguages/ironpython3/blob/8b853833aef23d327c64f4655d78152cfc05df24/Src/IronPython/Compiler/Parser.cs#L2684) (ignore the grammar comments above the methods, they do not exactly describe what the methods do). Also `ast` [organizes its "iterators"](https://github.com/IronLanguages/ironpython3/blob/8b853833aef23d327c64f4655d78152cfc05df24/Src/IronPython/Modules/_ast.cs#L580) as a sort of list of lists, each sublist headed by one `ComprehensionFor` and followed by zero or more `ComprehensionIf`.

For peace of mind, maybe a debug assert in the comprehension constructor?

BCSharp:
> didn't know this is how comprehensions worked

Neither did I until I looked into this issue. The clue was in [PEP 289](https://www.python.org/dev/peps/pep-0289/): 

> the semantic definition of a list comprehension in Python 3.0 will be equivalent to `list(<generator expression>)`

Idem:

> Only the outermost for-expression is evaluated immediately, the other expressions are deferred until the generator is run:
>```python
>g = (tgtexp  for var1 in exp1 if exp2 for var2 in exp3 if exp4)
>```
>is equivalent to:
>```python
>def __gen(bound_exp):
>    for var1 in bound_exp:
>        if exp2:
>            for var2 in exp3:
>                if exp4:
>                    yield tgtexp
>g = __gen(iter(exp1))
>del __gen
>```

For comprehensions, the talk about deferred evaluations is irrelevant, but the scoping implications for the first and only the first `for` expresion remain. Apparently, [Guido van Rossum found it more intuitive](https://www.python.org/dev/peps/pep-0289/#id17). I personally find it counterintuitive, which just shows that intuition is a very personal thing... 😃

slozier:
Test failure appears to be an unrelated concurrency issue.

Issue #982, opened 2020-10-14
Inconsistent use of tabs and spaces in indentation no longer allowed


slozier:
Inconsistent use of tabs and spaces in indentation is no longer allowed. With Python 2 this was allowed, but could be guarded again using the -t (or -tt) command line argument.

This should always throw:
```Python
compile("try:\n\t1/0\n    \t1/0\nfinally:\n pass\n", '<string>', 'exec')
```

Things to do:
- [ ] Ignore `-t` and `-tt` command line arguments
- [ ] Remove `-t` and `-tt` from command line help
- [ ] Remove `sys.flags.tabcheck`
- [ ] Remove `PythonOptions.IndentationInconsistencySeverity` and related code

BCSharp:
> Remove `-t` and `-tt` command line arguments

I think better than remove is to ignore, like [CPython does](https://github.com/python/cpython/blob/b81c833ab51fb7d7f0f8eaace37f60ef7455aa85/Python/initconfig.c#L2002):

```C
        case 't':
            /* ignored for backwards compatibility */
            break;
```

slozier:
Fine by me. Will amend to "remove from command line help".

slozier:
Fixed by https://github.com/IronLanguages/ironpython3/pull/1131

PR #1131, opened 2021-03-13
Remove -t and -tt functionality from ironpython

dc366:
This is a PR for issue #982, which removes console flags for indentation inconsistency. As suggested by @BCSharp, the arguments parser ignores `-t` and `-tt` by taking no action on them.

I have not updated the test suite. I see that the relevant test is `test_t()` in `test_stdconsole.py`. Curiously, the test suite only fails on `net46` after I make these changes.

Your comments are appreciated!

- [x]  Ignore -t and -tt command line arguments
- [x] Remove -t and -tt from command line help
- [x] Remove sys.flags.tabcheck
- [x] Remove PythonOptions.IndentationInconsistencySeverity and related code


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/IronLanguages/ironpython3?pullRequest=1131) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/IronLanguages/ironpython3?pullRequest=1131) before we can accept your contribution.<br/><br/>:x: dc366 [sign now](https://cla.dotnetfoundation.org/IronLanguages/ironpython3?pullRequest=1131)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/IronLanguages/ironpython3?pullRequest=1131) it.</sub>

Src/IronPython/Compiler/Tokenizer.cs
```diff
@@ -1325,10 +1313,6 @@ private class TokenEqualityComparer : IEqualityComparer<object> {
         /// Returns whether the 
         /// </summary>
         private bool ReadNewline() {
-            // Check whether we're currently scanning for inconsistent use of identation characters. If
-            // we are we'll switch to using a slower version of this method with the extra checks embedded.
-            if (IndentationInconsistencySeverity != Severity.Ignore)
-                return ReadNewlineWithChecks();
```
slozier, line 1331: I believe `ReadNewlineWithChecks` is what we want to keep.

slozier (changes requested):
Thanks for the PR.

Unfortunately I think you removed some of the wrong code paths. I guess in the description it should have been noted that the expected behavior is what we would have with the `-tt` option turned on. That means we should keep the code that would run with `IndentationInconsistencySeverity == Severity.Error`.

slozier:
As for `test_stdconsole` only failing on `net46` it's because the whole suite is disabled in .NET Core.

Src/IronPython/Compiler/Tokenizer.cs
```diff
@@ -127,17 +126,6 @@ internal Tokenizer(ErrorSink errorSink, PythonCompilerOptions options, bool verb
             }
         }
 
-        public Severity IndentationInconsistencySeverity {
-            get { return _indentationInconsistencySeverity; }
-            set {
-                _indentationInconsistencySeverity = value;
-
-                if (value != Severity.Ignore && _state.IndentFormat == null) {
-                    _state.IndentFormat = new StringBuilder[MaxIndent];
```
slozier, line 136: Seeing a bunch of
> SystemError: Object reference not set to an instance of an object

in the CI log. Probably need to move this initialization to the constructor.

dc366:
I moved the initialization of indentFormat to the State constructor. In testing on my end, there are a few new test cases which fail as a result of this change. 

I realized I went beyond the scope of the original issue, from simply ignoring the command line flags for indentation warnings to forcing the tokenizer to check for inconsistent indentation (though this is the desired behavior in Python3). I am not sure I have the ability to debug the behavior of the new ReadNewline() function without spending a lot more time. Let me know what you think a reasonable stopping point is.

slozier (approved):
No problem! This is great (and exactly what I had in mind as resolution of the issue).

Most of the test failures were expected with this change and I amended the tests to pass. It looks like there was one bug in the original `ReadNewlineWithChecks` which I fixed.

slozier:
Thanks for the PR!

PR #1138, opened 2021-03-18
Replace sys.long_info with sys.int_info

### LINKED ISSUE #974: sys.int_info is missing
Not sure what the values should be, probably simplest to match CPython?
```Python
sys.int_info(bits_per_digit=30, sizeof_digit=4)
```

dc366:
Resolves #974

### DIFF
--- a/Src/IronPython/Modules/sys.cs
+++ b/Src/IronPython/Modules/sys.cs
@@ -406,12 +406,12 @@ Handle an exception by displaying it with a traceback on sys.stderr._")]
 
         #endregion
 
-        // These values are based on the .NET 2 BigInteger in Microsoft.Scripting.Math
-        public static longinfo long_info = new longinfo(32, 4);
+        // These values are based on the .NET BigInteger in System.Numerics.BigInteger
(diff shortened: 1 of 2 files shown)


Src/IronPython/Modules/sys.cs
```diff
@@ -406,12 +406,12 @@ public sealed class SysFlags : PythonTuple {
 
         #endregion
 
-        // These values are based on the .NET 2 BigInteger in Microsoft.Scripting.Math
-        public static longinfo long_info = new longinfo(32, 4);
+        // These values are based on the .NET BigInteger in System.Numerics.BigInteger
+        public static intinfo int_info = new intinfo(32, 4);
 
         [PythonType("long_info"), PythonHidden]
```
slozier, line 412:
```suggestion
        [PythonType("int_info"), PythonHidden]
```
This shows up as the type name: `type(sys.int_info)`

dc366: Oops!

slozier:
Thanks for the PR!

PR #1139, opened 2021-03-18
Generate list of generators in test_cgcheck

### LINKED ISSUE #902: Update test_cgcheck to automatically pick up new generate_*.py files
https://github.com/IronLanguages/ironpython3/blob/368e2c5f157094cfbd7e8295089f1f8fd6f1c3d0/Src/Scripts/test_cgcheck.py#L17-L18

dc366:
Resolves #902. `generate_indicetest` is blocked as it doesn't have a main function and breaks `test_main()`.


slozier:
Thanks!

Issue #1114, opened 2021-02-27
Add namereplace error handler on encode


slozier:
From the release notes:
> Added the "namereplace" error handlers. The "backslashreplace" error handlers now work with decoding and translating.

PR #1140, opened 2021-03-18
Add namereplace error handler

slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/1114

### DIFF
--- a/Src/IronPython/Runtime/Operations/StringOps.cs
+++ b/Src/IronPython/Runtime/Operations/StringOps.cs
@@ -1839,6 +1839,7 @@ namespace IronPython.Runtime.Operations {
                 case null:
                 case "backslashreplace":
                 case "xmlcharrefreplace":
+                case "namereplace":
                 case "strict": e = setFallback(e, new ExceptionFallback(e is UTF8Encoding)); break;
                 case "replace": e = setFallback(e, ReplacementFallback); break;
                 case "ignore": e = setFallback(e, new DecoderReplacementFallback(string.Empty)); break;
@@ -2091,6 +2092,11 @@ namespace IronPython.Runtime.Operations {
                 d["backslashreplace"] = BuiltinFunction.MakeFunction(
                     "backslashreplace_errors",
                     ReflectionUtils.GetMethodInfos(typeof(StringOps).GetMember(nameof(BackslashReplaceErrors), BindingFlags.Static | BindingFlags.NonPublic)),
+                    typeof(StringOps));
+
+                d["namereplace"] = BuiltinFunction.MakeFunction(
+                    "namereplace_errors",
+                    ReflectionUtils.GetMethodInfos(typeof(StringOps).GetMember(nameof(NameReplaceErrors), BindingFlags.Static | BindingFlags.NonPublic)),
                     typeof(StringOps));
 
                 d["surrogateescape"] = BuiltinFunction.MakeFunction(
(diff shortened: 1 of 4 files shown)


BCSharp:
Today I am AFK, but can look at it tomorrow.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -2638,6 +2644,69 @@ private class ExceptionFallbackBufferUtf8DotNet : ExceptionFallbackBuffer {
             }
         }
 
+        private static object NameReplaceErrors(object unicodeError) {
+            Modules.unicodedata.PerformModuleReload(null, null);
```
BCSharp, line 2648: There is a similar call in `LiteralParser`, which has been bothering me already for some time, since I had thought of enabling nullable annotations in `unicodedata.cs`. The old-style nullable annotations in the definition of `unicodedata.PerformModuleReload` mark both parameters as not-null. And indeed, `PythonContext` seems to be expected to be not null everywhere in the code base. It appears that in both and only cases of the explicit call to `PythonModuleReload` the goal is simply to make sure the private static variable `ucd_5_2_0` is properly initialized. So perhaps factor that out of `PythonModuleReload` into a parameterless internal method? I don't know what the `[SpecialName]` `PerformModuleReload` is supposed to represent in the general case to decide whether it is OK for the parameters to become nullable or not.

BCSharp:
Looks good to me, except for a small comment.

Src/IronPython/Runtime/Operations/StringOps.cs
slozier: I'm fine with making an internal method to initialize `ucd_5_2_0`. Presumably the analyzer should handle `[SpecialName] PerformModuleReload` in a `PythonModule` and check that the arguments are non-null (and don't have `[NotNull]` since it's not Python callable).

While on the subject of `NotNull` I've also been thinking about the conflict with the standard nullable annotation attribute. I wonder if we should create a new attribute to avoid the conflict. Unfortunately I can't come up with a good name for this new attribute. Perhaps `DlrNotNull`, although it's likely to prevent any future collisions it's not super descriptive.

Something to keep in mind for the future, if that if we want to mimic CPython, whenever we use the `namereplace` error handler or `\N` in a string we should load the `unicodedata` module in `sys.modules`. So something more along the lines of what we do with `PythonContext.EnsureEncodings` (although it's slightly different since `unicodedata` is embedded in our main library instead of being an external module).

BCSharp: > While on the subject of `NotNull` I've also been thinking about the conflict with the standard nullable annotation attribute. I wonder if we should create a new attribute to avoid the conflict. Unfortunately I can't come up with a good name for this new attribute. Perhaps `DlrNotNull`, although it's likely to prevent any future collisions it's not super descriptive.

I was wondering the same. Some time ago I have [made a suggestion](https://github.com/IronLanguages/ironpython3/pull/808#discussion_r426281282) naming it `NotNone` but it is Python-centric... DLR itself calls a null type `DynamicNull`, but `[NotDynamicNull]` looks too long... `NotDNull`?

BCSharp (approved):

Src/IronPython/Runtime/Operations/StringOps.cs
slozier: `NotDNull` seems a bit obscure. I think I could live with `NotDynamicNull`...

BCSharp: I think I could live with that too...

PR #1142, opened 2021-03-19
Properly bind references in comprehension scopes

### LINKED ISSUE #1082: Incorrect exception type thrown for referenced unassigned variables in comprehensions
Example:

```python
def foo():
    [x for y in range(2)]
    x = 2
foo()
```

CPython:
```
NameError: free variable 'x' referenced before assignment in enclosing scope
```

IronPython:
```
UnboundLocalError: Local variable 'x' referenced before assignment.
```

Maybe a remnant from the time when a list comprehension did not introduce a new scope.

### LINKED ISSUE #1132: dir in comprehensions omits closure variables
When run within a given non-global scope, `dir()` should report all non-global variables used, including referenced variables defined in encompassing scopes (closure variables). However, it does not happen as expected for comprehension scopes. Example:

```python
l1 = [1] # global
def f():
    l2 = [2] # outer local
    return [dir() for x in l1 for y in l2]
res = f()
assert 'l1' not in res[0] # passes, l1 is global
assert 'l2' in res[0] # fails, l2 is a closure var
```

The same thing happens with `locals`, `vars`, etc. and other comprehension types (set, dict).

Possibly related to #1082.

----
Originally discovered when writing tests for #1130.

BCSharp:
Resolves #1132 and at the same time fixes #1082.

### DIFF
--- a/Src/IronPython/Compiler/Ast/Comprehension.cs
+++ b/Src/IronPython/Compiler/Ast/Comprehension.cs
@@ -273,6 +273,7 @@ namespace IronPython.Compiler.Ast {
         }
 
         internal override PythonVariable BindReference(PythonNameBinder binder, PythonReference reference) {
+            // First try variables local to this scope
             if (TryGetVariable(reference.Name, out PythonVariable variable)) {
                 if (variable.Kind == VariableKind.Global) {
                     AddReferencedGlobal(reference.Name);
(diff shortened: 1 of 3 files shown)


BCSharp:
The last commit is not part of the fix but I included it here since I was at the comprehensions again.

It's meant to enforce the conditions checked in the constructors (added in #1130) throughout the lifetime of the comprehension object.

slozier (approved):
Looks good to me!

Issue #1143, opened 2021-03-19
Using a generator in a function adds an empty closure to functions defined in the same scope


BCSharp:
This works:

```python
def t():
    def f(): pass
    assert f.__closure__ == None

t()
```

This doesn't:

```python
def t():
    def f(): pass
    assert f.__closure__ == None
    (str(x) for x in [1, 2])

t()
```

The `__closure__` attribute of the nested function `f` is now an empty tuple. Since the assert failure happens before the extra generator call, it suggests the effect is created during compilation.

BCSharp:
It turns out this has nothing to do with the generators; any scope at the peer level that has free variables (local or global), produces the same result. This fails as well:

```python
def t():
    def f(): pass
    assert f.__closure__ == None, "f.__closure__ is " + str(f.__closure__)
    def g(): return str

t()
```

BCSharp:
There are two root causes of this strange behaviour, both are incompatibilities with CPython.

### Closure Sharing

The first is that in IronPython, all nested scopes **share** one closure. Therefore, the closure must contain all free local-scope variables accessed in any nested scopes. In CPython, each nested scope has a tailored closure containing only what is needed.

```python
def t():
    x, y = 1, 2
    def f(): return x
    def g(): return y

    print("Closure of f():", [cell.cell_contents for cell in f.__closure__] if f.__closure__ is not None else None)
    print("Closure of g():", [cell.cell_contents for cell in g.__closure__] if g.__closure__ is not None else None)

t()
```
Results from IronPython:
```
Closure of f(): [1, 2]
Closure of g(): [1, 2]
```
Results from CPython:
```
Closure of f(): [1]
Closure of g(): [2]
```

### Referenced globals trigger closure creation

This is because closure creation is triggered if `ContainsNestedFreeVariables` is true (among other things), and this variable does not discriminate between local and global references. Example:

```python
def t():
    def f(): return str
    print("Closure of f():", [cell.cell_contents for cell in f.__closure__] if f.__closure__ is not None else None)

t()
```
IronPython:
```
Closure of f(): []
```
CPython:
```
Closure of f(): None
```

BCSharp:
[...]

PR #1145, opened 2021-03-20
Prevent closure creation when all free variables are globals

BCSharp:
This solves one of the root causes of #1143. As such, it is merely an optimization, and compatibility improvement in some limited cases.


slozier (approved):
Looks good to me.

slozier:
@BCSharp Hmm, following this PR, when running in Debug mode, I'm getting some assertion failures from the `Debug.Assert(NeedsLocalContext);` in `ComprehensionScope.GetParentClosureTuple`. Presumably the assertion is no longer valid? Any ideas?

BCSharp:
It seems that the assertion is now too strong, but I'd rather replace it with a weaker one than remove it altogether.

PR #1150, opened 2021-03-27
Prevent nonlocal binding to global variables

BCSharp:
The existing code was already skipping the global scope when resolving nonlocal references, but failed to account for global variables introduced into local scopes with the `global` statement.


slozier (approved):
Looks good to me. Thanks!

PR #1151, opened 2021-03-28
Prefer local class namespace over lexical scoping

### LINKED ISSUE #1030: Name binding of locals and class does not as expected
From `test_namebinding`:
```py
def test():
    xyz = False
    class C:
        locals()["xyz"] = True
        passed = xyz
    
    assert C.passed == True

test()
```

The following (also from `test_namebinding`) might be the same issue:
```py
def f():
    a = 2
    class c:
        exec("a = 42")
        abc = a
    return c

assert f().abc == 42
```

BCSharp:
Resolves #1030.

The original bug got fixed in CPython 3.4 ([bpo-17853](https://bugs.python.org/issue17853)) because of an interesting metaclass scenario. I tried to add that scenario to the IronPython tests, but it fails, presumably because [PEP 3115](https://www.python.org/dev/peps/pep-3115/) is not implemented yet? Are the old-style metaclasses supposed to work?

### DIFF
--- a/Src/IronPython/Compiler/Ast/AstMethods.cs
+++ b/Src/IronPython/Compiler/Ast/AstMethods.cs
@@ -61,6 +61,7 @@ namespace IronPython.Compiler.Ast {
         public static readonly MethodInfo MakeClosureCell = GetMethod((Func<ClosureCell>)PythonOps.MakeClosureCell);
         public static readonly MethodInfo MakeClosureCellWithValue = GetMethod((Func<object, ClosureCell>)PythonOps.MakeClosureCellWithValue);
         public static readonly MethodInfo LookupName = GetMethod((Func<CodeContext, string, object>)PythonOps.LookupName);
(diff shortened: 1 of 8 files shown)


slozier:
> Are the old-style metaclasses supposed to work?

Hmm, old-style metaclasses should be gone and new-style should be mostly working (although they we went through a few hacky implementations to bootstrap things so I'm not sure anymore :smile:).

BCSharp:
So I think the reason my metaclass tests are failing is on the metaclass side, for which I've filed a new issue #1154. I've added the test here anyway, so that when #1154 is resolved, the test will not be forgotten.

Since there seems to be no issue with this PR, I'm marking is as ready for review, and, hopefully, will find some time next weekend to look into the metaclass issue (unless somebody beats me to it... 😄 ).

slozier (approved):
Looks good to me!

Issue #686, opened 2019-11-24
validation for star expressions should not be done in the parser


slozier:
For example, `x = ast.parse("a = 1,*b")` should succeed and only fail when compiling `compile(x, None, 'exec')`.

See https://github.com/IronLanguages/ironpython3/pull/662#discussion_r339225110

slozier:
https://github.com/IronLanguages/ironpython3/pull/1158 resolves the parsing, but the expression fails to compile with a TypeError:
```Python
import ast
x = ast.parse("a = 1,*b")
compile(x, "", 'exec')
```

> TypeError: Unexpected expr type: IronPython.Modules._ast+Starred

PR #1158, opened 2021-04-01
Move starred expression checks out of parser

slozier:
Moves the starred expression checks out of parser which resolves the following:
```Python
import ast
try:
    ast.parse("*a")
except SyntaxError:
    assert False
```

### DIFF
--- a/Src/IronPython/Compiler/Ast/PythonAst.cs
+++ b/Src/IronPython/Compiler/Ast/PythonAst.cs
@@ -149,6 +149,7 @@ namespace IronPython.Compiler.Ast {
         /// </summary>
         public void Bind() {
             PythonNameBinder.BindAst(this, _compilerContext);
+            StarredExpressionChecker.Check(this, _compilerContext);
         }
 
         public override string Name {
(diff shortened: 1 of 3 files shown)


slozier:
@BCSharp do you think this seems like a reasonable approach to post-parsing syntax checks?

BCSharp (approved):
I think it is a good approach. My only concern was about the ordering of the checks (it comes after namebinding syntax errors are reported), but it appears this is exactly what CPython does as well, so all is well.

A few other things that caught my eye, I'm pretty sure you know this, but just for the record:
1. Going through `compile(ast.parse(...), ...)` does not work, even for valid cases; I assume this is for another PR, maybe a one that finishes off #686?
2. Starting from Python 3.5, the error message "can use starred expression only as assignment target" is replaced by "can't use starred expression here", I assume this will get changed in #1153?

slozier:
> Going through `compile(ast.parse(...), ...)` does not work, even for valid cases; I assume this is for another PR, maybe a one that finishes off #686?

Ah, I knew there was a starred expression issue somewhere, good catch. I guess this solved half the issue. Will take a look at the second half in a separate PR (if I don't forget about it).

> Starting from Python 3.5, the error message "can use starred expression only as assignment target" is replaced by "can't use starred expression here", I assume this will get changed in #1153?

Indeed, 3.5 will change a bunch of things in this walker, but it seemed like a good idea to work this out separately from the other PR.

PR #1165, opened 2021-04-11
Release notes

slozier:
Draft of release notes for 3.4.0-alpha1.

@BCSharp can you think of anything else that should be included?


slozier:
Probably worth adding some notes about the future of IronPython 2.

BCSharp:
Maybe it would be a good idea to add a link to [differences-from-c-python.md](https://github.com/IronLanguages/ironpython3/blob/master/Documentation/differences-from-c-python.md), which spells out some minor known incompatibilities.

Also, what about adding links to the "WhatsNewInPython3x.md" series, so that people can easily look up the status of their favourite Python feature. Furthermore, since a release may increase usage of IronPython 3, making not implemented parts more visible can prevent unnecessary/duplicate issue submissions.

As for the release itself, I think it would be nice (though not necessary for an alpha) to implement [PEP 514](https://www.python.org/dev/peps/pep-0514/) in the MSI installer. Last time I checked (about a month ago), it wasn't creating the mandatory registry keys as per PEP 514, and this PEP is relied on by several development tools (most notably by PTVS).

And last but not least: congrats on the release!

slozier:
Thanks for the feedback. I've moved some of the info out of the announcement and into the readme+articles. The plan is to remove the readme.md from this PR and use its text in the release announcement.

slozier:
PEP 514: seems interesting, in particular (at least for me) if the py launcher can pick up on it and be used to launch IronPython.

BCSharp:
> PEP 514: seems interesting, in particular (at least for me) if the py launcher can pick up on it and be used to launch IronPython.

I think this will only work if IronPython 3.4 is the only 3.4 Python installed on the system. Then it can be launched by providing option `-3.4` to `py`. I see no other options to tie in py launcher configuration options ([PEP 397](https://www.python.org/dev/peps/pep-0397/)) to PEP 514. Otherwise, some settings in `py.ini` are required to [make py launch IronPython](https://github.com/IronLanguages/pylauncher/blob/master/Doc/launcher.rst).

Talking about py launcher, what are the plans for the project [IronPython/pylauncher](https://github.com/IronLanguages/pylauncher)?

BCSharp:
> The plan is to remove the release.md from this PR and use its text in the release announcement.

I understand. In that case the links in the release announcement to the internal documentation (i.e. the differences, what's new, etc.) would need to point to a specific revision (or label) on GitHub, since those files may be updated later on as the development continues after the release.

Issue #1163, opened 2021-04-04
keyword arguments not passed to metaclass __new__


slozier:
Ran into this issue while testing the `typing` module in the 3.5 stdlib:
```Python
class TypingMeta(type):
    def __new__(cls, name, bases, namespace, **kwargs):
        assert "_root" in kwargs

class _TypingBase(metaclass=TypingMeta, _root=True):
    pass
```

PR #1167, opened 2021-04-11
Separate positional and keyword arguments in CallExpression

BCSharp:
Code refactoring of `CallExpression`.

It came up when I was looking into keyword arguments for metaclasses (#1163), which most likely will have an impact on `CallExpression` as it is now. So before starting with metaclasses, I wanted to get this one out of the way, since it brings a public interface change.

### DIFF
--- a/Src/IronPython/Compiler/Ast/CallExpression.cs
+++ b/Src/IronPython/Compiler/Ast/CallExpression.cs
@@ -22,18 +22,20 @@ using MSAst = System.Linq.Expressions;
 namespace IronPython.Compiler.Ast {
 
     public class CallExpression : Expression, IInstructionProvider {
-        public CallExpression(Expression target, Arg[] args) {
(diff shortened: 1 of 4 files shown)


slozier (approved):
Looks good to me! I thought about doing this but was too lazy to go through with it. 😄

Not specifically an issue with this PR, but one thing that was bugging me while I was working on this code was the `Arg` construct. It doesn't seem particularly, especially now that we have split arguments. For example `args` could just be a list of expressions and we look for `StarredExpression`. Also `GetArgumentInfo` is terrible since it creates an `Argument` whenever call it on an argument that's not simple (and most of the time we're calling it just to check the `Kind`). Anyway, enough of my rant. 😺

BCSharp:
Thanks for the rant! Now the `Arg` construct has started to bug me too... 😄

Issue #821, opened 2020-05-04
Cannot import os/no implementation of stat


alexbuzzbee:
### Prerequisites

* [X] Are you running the latest version? (Latest Git `master`)
* [X] Are you reporting to the correct repository?
* [X] Did you perform a cursory search? (Searched for `os` and `stat`.)

### Description

The `os` module throws an `ImportError` when imported.

### Steps to Reproduce

1. Run `ipy`.
2. `import os`

**Expected behavior:**

`os` module is imported.

**Actual behavior:**

```
Traceback (most recent call last):
File "<stdin>", line 1, in <module>
File "<string>", line 539, in _check_name_wrapper
File "<string>", line 1614, in load_module
File "<string>", line 596, in _load_module_shim
File "<string>", line 1220, in load
File "<string>", line 1200, in _load_unlocked
File "<string>", line 1129, in _exec
File "<string>", line 1471, in exec_module
File "<string>", line 321, in _call_with_frames_removed
File "C:\path\to\ironpython3\bin\Release\netcoreapp3.1\Lib\os.py", line 27, in <module>
ImportError: No module named stat
```

### Versions

```
> bin/Release/netcoreapp3.1/ipy -V
IronPython 3.4.0 3.4.0.0 on .NET 3.1.3
> git log
commit 7c687497c6ce3a0a3d5db9c0d9b39d62d128f85c (HEAD -> master, origin/master, origin/HEAD)
...
```

slozier:
`stat` is a Python module, sounds like you need to include the standard library. You can do that, for example, by setting the `IRONPYTHONPATH` environment variable to `C:\path\to\ironpython3\Src\StdLib\Lib`.

PR #1168, opened 2021-04-12
Remove os.py from build Lib folder

slozier:
It seems like people are building the Release, running it and then trying to `import os` (e.g. https://github.com/IronLanguages/ironpython3/issues/821). This fails since `os` has dependencies on `stat` (and a number of other modules). This just removes `os.py` from the Lib folder so that people don't get the impression that they can `import os` without including the standard library (documentation is probably lacking about this).


slozier:
Closing this since os.py is used to probe for the standard library path folder (CPython does the same).

PR #1169, opened 2021-04-12
Optimize IronPython.Compiler.Ast.Arg

BCSharp:
This removes some of the bugginess... 😸


slozier (approved):
Indeed an improvement. 😄

PR #1170, opened 2021-04-13
Support starred arguments as class bases

BCSharp:
@slozier I know you are not a huge fan of `Compiler.Ast.Arg`, but the reason I changed the type of `ClassDefinition.Bases` from a ro list of `Expression` to a ro list of `Arg` is to align it more closely with `CallExpression`. [PEP 3115](https://www.python.org/dev/peps/pep-3115/) specifically declares that:

> the parameter list passed to a class definition will now support all of the features of a function call

so ideally this should be handled by the same code path as a function call. Indeed, this is what I think CPython does with routing class creation with a call to the internal `__build_class__` function. 

I looked first at implementing it here too, but eventually decided not to go ahead with such approach. The way how arguments are handled in `CallExpression` is tailored to what the DLR supports, which in turn does what it does to implement sophisticated overload resolution for .NET. On the other hand, `ClassDefinition` is optimized to call `PythonOps.MakeClass` directly, which in turn uses extra parameters to make a class efficiently.

Hence handling of class arguments is done separately from handling of function call arguments, except for some overlap in the parser. So unfortunately, when changes are made (e.g. Python 3.5 allows multiple starred arguments), this will have to be implemented in two places. To ease that, I wanted to keep both parts of the code aligned as much as possible without compromising on efficiency.


slozier (commented):
Looks good to me! Though when I find a bit of time I'll probably go through the codebase to eliminate `Arg`. 😄

PR #1178, opened 2021-04-18
Update after review

slozier:
This should address the comments from https://github.com/IronLanguages/ironpython3/pull/1175#pullrequestreview-638355520.

The only think I haven't changed yet is the `EmptyArray` pending feedback from @BCSharp.


BCSharp:
@slozier I am opting for removing `Expression.EmptyArray`.

PR #1179, opened 2021-04-18
Enable some metaclass tests

BCSharp:


Tests/test_metaclass.py
```diff
@@ -114,7 +114,7 @@ def method(self): return 10
             self.assertEqual(x.method(), 10)
 
         try_metaclass(type)
-        #try_metaclass(type(Old))  # bug 364938
+        try_metaclass(type(Old))
```
slozier, line 117: We could probably just remove any `Old` related code since it's for old-style classes which no longer exist.

slozier (approved):
Looks good to me!

Tests/test_metaclass.py
BCSharp: There is a different code path in a few places between no base classes specified vs. explicit `object` as a base. The code should compensate for this but I would keep the tests. Admittedly, the names `Old`/`New` are misnomers, any ideas for better names?

slozier: Can't think of good names, maybe something with `ExplicitBase` for the `(object)` version? We're going to need good comments for this otherwise I may forget about the different code paths and end up getting rid of it if I ever revisit.

This particular line though is identical to the line above since `type(Old) is type`.

slozier:
Thanks!

PR #1180, opened 2021-04-19
Enable 32-bit tests during CI

slozier:


Lamparter:
@slozier what's stopping this from being merged?

slozier:
I don't remember why I left this one as a draft. Maybe something was failing? Or Windows 11 being 64-bit only made testing it less interesting. I guess I can rebase and see if anything is failing.

Lamparter:
How is the check still running?
I'm surprised it's taking this long

slozier:
Ahh, a bunch of tests are not actually running in 32-bit mode because the process runner is launching ipy.exe. Seems vaguely familiar - probably why I moved it to draft in the first place. Guess I should have left a note. :smile:

slozier:
Closing this one as it just increases CI time considerably for little to no gain.

PR #1181, opened 2021-04-20
Updated C# snippet in the README.md to use top level statements.

Kuinox:
Hello,
Top level statements were introduced in C# with .NET 5.  
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/top-level-statements
For the comparison to be fair, I think the readme should use this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,8 @@ C#:
 
 ```cs
 using System;
-class Hello
-{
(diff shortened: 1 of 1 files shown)


slozier (approved):
Looks good to me, not sure why we're comparing with C# anyway. 😄

Issue #1154, opened 2021-03-29
Metaclass __prepare__ executed out of order


BCSharp:
According to [PEP 3115](https://www.python.org/dev/peps/pep-3115/), the `__prepare__` method defined in a metaclass should be executed early, before the main body of the class using that metaclass as its type. But currently `__prepare__` is executed after (and twice) so the dictionary-like objects it creates is not being used while the body of the class is being evaluated.

Example:

```python
class Meta(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        print("__prepare__ called")
        return dict(x="metaclass prepared")

class C(metaclass=Meta):
    print("class body executing")
    assert 'x' in dir() # IronPython fails

assert 'x' in dir(C) # passes if the previous assert is commented out
```

CPython:
```
__prepare__ called
class body executing
```

IronPython (after removing the assertions):
```
class body executing
__prepare__ called
__prepare__ called
```

slozier:
Looks like I'm the one to blame for the double `__prepare__` calls (https://github.com/IronLanguages/ironpython3/pull/174). Although I seemed to have been aware of the issues with it. 😃

PR #1189, opened 2021-04-23
Improve metaclass logic

BCSharp:
It addresses a part of the problem reported in #1154 (duplicate calls to `__prepare__`) and several other related issues. Off the top of my head:
* removes duplicate calls to `__init__`
* supports inheriting of `__prepare__`, `__new__`, `__init__` by metaclasses
* supports `__prepare__` on a function used as "metaclass"
* fixes interplay when mixing metaclasses provided classes and functions in various scenarios
* ...

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -1305,66 +1305,17 @@ namespace IronPython.Runtime.Operations {
             iwr.SetFinalizer(new WeakRefTracker(iwr, nif, nif));
         }
 
-        internal static object? CallPrepare(CodeContext/*!*/ context, PythonType meta, string name, PythonTuple bases, PythonDictionary? keywords, PythonDictionary dict) {
-            object? classdict = dict;
-
-            // if available, call the __prepare__ method to get the classdict (PEP 3115)
-            if (meta.TryLookupSlot(context, "__prepare__", out PythonTypeSlot pts)) {
-                if (pts.TryGetValue(context, null, meta, out object value)) {
-                    if (keywords is null || keywords.Count == 0) {
-                        classdict = PythonOps.CallWithContext(context, value, name, bases);
-                    } else {
-                        var args = new object[] { name, bases };
-                        classdict = PythonCalls.CallWithKeywordArgs(context, value, args, keywords);
-                    }
-                    // copy the contents of dict to the classdict
-                    foreach (var pair in dict)
-                        context.LanguageContext.SetIndex(classdict, pair.Key, pair.Value);
-                }
-            }
-
-            return classdict;
-        }
-
         public static object MakeClass(FunctionCode funcCode, Func<CodeContext, CodeContext> body, CodeContext/*!*/ parentContext, string name, PythonTuple bases, PythonDictionary? keywords, string selfNames) {
-            Func<CodeContext, CodeContext> func = GetClassCode(parentContext, funcCode, body);
+            Func<CodeContext, CodeContext> func = getClassCode(parentContext, funcCode, body);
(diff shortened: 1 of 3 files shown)


Tests/test_metaclass.py
```diff
@@ -380,38 +381,37 @@ class Foo(test=1, **{'metaclass':funcMeta, 'test2':2}): pass
...
 
-        msg = "MetaP() takes at most 4 arguments (5 given)" if is_cli else "type() takes 1 or 3 arguments"
-        with self.assertRaisesMessage(TypeError, msg):
+        with self.assertRaisesRegex(TypeError, r"\(\) takes .* arguments"):
             class Foo(test=True, metaclass=MetaP): pass
 
-        class MetaN(type):
-            def __new__(metacls, name, bases, attrdict, **kwargs):
-                return type.__new__(metacls, name, bases, attrdict)
+        if (sys.version_info < (3,6)):
```
slozier, line 401: What's the behavior for >= (3,6), is it something we could easily adopt? Should we have a negative test to make sure we fail when we get to 3.6?

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1390,36 +1358,82 @@ public static partial class PythonOps {
...
+                PythonDictionary vars = func(parentContext).Dict;
+                return PythonType.__new__(parentContext, TypeCache.PythonType, name, bases, vars, selfNames);
             }
 
-            object? classdict = vars;
-
-            if (metaclass is PythonType metatype) {
-                classdict = CallPrepare(context, metatype, name, bases, keywords, vars);
-            }
+            // Prepare classsdict
```
slozier, line 1365: Typo: classsdict

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1390,36 +1358,82 @@ public static partial class PythonOps {
...
 
+            // Ensure the class derives from `object`
             if (obj is PythonType newType && newType.BaseTypes.Count == 0) {
                 newType.BaseTypes.Add(DynamicHelpers.GetPythonTypeFromType(typeof(object)));
             }
 
             return obj;
+            // ------------------------------------------------------------------------
+
+            static Func<CodeContext, CodeContext> getClassCode(CodeContext/*!*/ context, FunctionCode funcCode, Func<CodeContext, CodeContext> body) {
```
slozier, line 1394: Not that you need to change this, but I usually PascalCase local functions. We should probably adopt a convention. I'll have to check if there's an editorconfig rule for this.

slozier (approved):
I'm glad one of us is good at writing tests! Just some minor comments, looks goot to me otherwise.

Src/IronPython/Runtime/Operations/PythonOps.cs
BCSharp: I usually use dromedaryCase for anything local (inc. functions) to distinguish it from regular methods. But I am happy to adopt any convention; .editorconfig check would be good. However, I may have committed my dromedaryCase in some other places already, though now I can't find it...

slozier: I managed to define an .editorconfig rule, unfortunately it seems like the file has to be open in VS for it to display an issue. Anyway, I opened a bunch of files to see if it would find rule violation and there are indeed a few. An exception to the rule might be the helper methods like `link.linkWindows` where the casing matches the public parent. Another interesting case is where the local function is essentially replacing lambda (e.g. in `SurrogatePassErrors`). I assume I would have dromedaryCased the lambda? I guess it's not as clear cut as I thought it was.

PR #1192, opened 2021-04-24
Update example in README.md

BCSharp:
I think it is safe to assume that visitors to the IronPython GitHub project page know what Python is. So the example to show how to print "Hello World!" in plain Python is kind of pointless. Also, they probably know C# and .NET, so the only reason for comparing IronPython code to C# would be to give an idea for people wondering "well, I know how to do .NET stuff in C#, how would that work in IronPython?"

I've come up here with two simple but-not-totally-trival examples: how to access .NET from IronPython and how to access IronPython form C#. Comments welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@ IronPython 3
 
 IronPython is an open-source implementation of the Python programming language which is tightly integrated with .NET. IronPython can use .NET and Python libraries, and other .NET languages can use Python code just as easily.
 
+IronPython 3 targets Python 3, including the re-organized standard library, Unicode strings, and all of the other new features.
+
+
 | **What?** | **Where?** |
 | --------: | :------------: |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,28 +6,49 @@ IronPython 3
...
+var eng = IronPython.Hosting.Python.CreateEngine();
+var scope = eng.CreateScope();
+eng.Execute(@"
+def greetings(name):
+    return 'Hello ' + name.title() + '!'
+", scope);
+dynamic greetings = scope.GetVariable("greetings");
+System.Console.WriteLine(greetings("world"));
+```
+This example assumes that `IronPython` has been added the the C# project as a NuGet package.
```
slozier, line 51: Typo: the the

slozier:
I think these are good examples, definitely an improvement over the print example which doesn't really highlight why IronPython is interesting.

PR #1233, opened 2021-05-23
Add methods to deque

slozier:
Adds `copy`, `index`, `insert`, `__contains__`, `__add__` and `__mul__` to `deque`.


slozier:
Looks like there's a repr issue:
```py
from collections import deque
d = deque([1, 2], 5)
assert list(d*2) == [1, 2, 1, 2]
assert repr(d*2) == "deque([1, 2, 1, 2], maxlen=5)"
```

PR #1314, opened 2022-02-12
Separate intermediate item directories for .NET 5 and 6 builds

BCSharp:
Fixes intermittent build failures of _IronPython.Wpf_ and _IronPython.SQLite_ when parallel builds are enabled with multiple target frameworks.


slozier:
Thanks!

PR #1315, opened 2022-02-12
Type name

BCSharp:
Prepares the ground for #52.

A handful of calls to `PythonTypeOps.GetName` remain where they are used for debugging purposes.

I also have left a few calls to `DynamicHelpers.GetPythonType(o).Name` in `ctypes`. I don't know that code well enough to judge whether the change would be justified, and since it is `ctypes` I thought it might be better to keep using the actual type name there.


BCSharp:
This is it, for now. As you said, we can iron out the kinks as we go.

slozier (approved):
Nice! Looks good to me.

Issue #1309, opened 2022-01-21
dir is much slower than with ipy2


slozier:
For example, creating this test.py file:
```py
import time

for i in range(10000):
    exec("a" + str(i) + " = 1")

t = time.time()
dir()
print(time.time() - t)
```
and executing it is quite slow:
```
> ipy2 test.py
0.00299835205078
> ipy3 test.py
2.7853622436523438
```

PR #1318, opened 2022-02-16
Improve perf while iterating over globals

slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/1309

### DIFF
--- a/Src/IronPython/Runtime/CustomDictionaryStorage.cs
+++ b/Src/IronPython/Runtime/CustomDictionaryStorage.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 using Microsoft.Scripting;
 using Microsoft.Scripting.Runtime;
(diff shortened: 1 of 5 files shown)


BCSharp:
Is this a regression?

slozier:
@BCSharp Yes and no I guess. Yes in that perf is horrible compared to ipy2, no in that the only change compared to ipy2 is return value of `dict.keys` changed from a list to a `dict_keyiterator`. This is just a quick fix for this specific scenario and doesn't really resolve the underlying perf issue.

The underlying issue is that the dictionary iterators call `DictionaryStorage.Count` to check for size changes at every step and this is not cheap for certain storages.

BCSharp:
Would it help to cache the size and invalidate it only on dictionary changes?

slozier:
That might work, but I'm not super comfortable with this part of the codebase so not something I want to mess with right now. Will probably need to revisit if I ever get to making dicts preserve insertion order.

PR #1319, opened 2022-02-16
Disable warning about using end-of-life .NET Core 2.1

BCSharp:
.NET Core 2.1 is end of life but is used to test .NET Standard 2.0 assemblies.


slozier (approved):
Looks good to me.

PR #1322, opened 2022-02-18
Apply customary type conversions when assigning values to ndim Array

### LINKED ISSUE #1320: Inconsistent type conversion rules for CLI arrays
### Description

When assigning integer values to an element of a multi-dimensional array, the customary type conversions are not performed. However, they are performed when the array is 1-dimensional.

### Steps to Reproduce

```python
IronPython 3.4.0a1 (3.4.0.0001)
[.NETCoreApp,Version=v6.0 on .NET 6.0.1 (64-bit)] on darwin
>>> import clr
>>> clr.AddReference("System.Numerics")
>>> from System import Array, Int32
>>> from System.Numerics import BigInteger
>>> rank1arrayOfInt32 = Array[Int32](3)        
>>> rank1arrayOfBigInteger = Array[BigInteger](3) 
>>> rank1arrayOfBigInteger[0] = 1  # assigning Int32 to an element of type BigInteger: OK, widening conversion
>>> print(rank1arrayOfBigInteger)
Array[long]((1, 0, 0))
>>> rank1arrayOfInt32[0] = 1 << 64 >> 64 # assigning BigInteger to an element of type Int32: OK, narrowing conversion
>>> print(rank1arrayOfInt32)
Array[int]((1, 0, 0))
>>> rank2arrayOfInt32 = Array[Int32](3, 2)
>>> rank2arrayOfBigInteger = Array[BigInteger](3, 2)
>>> rank2arrayOfBigInteger[0, 0] = 1   # assigning Int32 to an element of type BigInteger: not OK?
TypeError: Object cannot be stored in an array of this type.
>>> rank2arrayOfInt32[0, 0] = 1 << 64 >> 64   # assigning BigInteger to an element of type Int32: not OK?      
TypeError: Object cannot be stored in an array of this type.
>>> rank2arrayOfInt32[0, 0] = 1  # assignments that do not require conversion work OK
>>> rank2arrayOfInt32[0, 0]                     
1
>>> rank2arrayOfBigInteger[0, 0] = 1 << 64 >> 64
>>> rank2arrayOfBigInteger[0, 0]
1
```

**Expected behavior:**

Multi-dimensional arrays behave like one-dimensional arrays.

### Related Case

Something similar (though not the same) is happening with arrays of `Single` and `Double`: assigning the other type to a 1-dimensional array works in both ways, but when 2-dimensional arrays are used, assigning a `Double` to an array of `Single` fails (but with a different message than for integer types), but assigning a `Single` to an array of `Double` works fine.

```python
>>> from System import Array, Single, Double
>>> rank1arrayOfSingle = Array[Single](3)
>>> rank1arrayOfDouble = Array[Double](3)   
>>> rank1arrayOfSingle[0] = Double(1.2)   # narrowing conversion
>>> rank1arrayOfDouble[0] = Single(1.2)   # widening conversion
>>> rank2arrayOfSingle = Array[Single](3, 2)
>>> rank2arrayOfDouble = Array[Double](3, 2)
>>> rank2arrayOfSingle[0, 0] = Double(1.2)  # no narrowing conversion?
ValueError: Cannot widen from source type to target type either because the source type is a not a primitive type or the conversion cannot be accomplished.
>>> rank2arrayOfDouble[0, 0] = Single(1.2)  # widening conversion
```

I will look into this issue in the coming days because it is blocking #52.

BCSharp:
Resolves #1320

### DIFF
--- a/Src/IronPython/Runtime/Operations/ArrayOps.cs
+++ b/Src/IronPython/Runtime/Operations/ArrayOps.cs
@@ -164,8 +164,7 @@ namespace IronPython.Runtime.Operations {
         public static void SetItem(Array a, params object[] indexAndValue) {
             if (indexAndValue == null || indexAndValue.Length < 2) throw PythonOps.TypeError("__setitem__ requires at least 2 parameters");
 
-            int iindex;
-            if (indexAndValue.Length == 2 && Converter.TryConvertToInt32(indexAndValue[0], out iindex)) {
+            if (indexAndValue.Length == 2 && Converter.TryConvertToInt32(indexAndValue[0], out int iindex)) {
                 SetItem(a, iindex, indexAndValue[1]);
                 return;
             }
@@ -180,7 +179,9 @@ namespace IronPython.Runtime.Operations {
             if (a.Rank != args.Length) throw PythonOps.ValueError("bad dimensions for array, got {0} expected {1}", args.Length, a.Rank);
 
             for (int i = 0; i < indices.Length; i++) indices[i] += a.GetLowerBound(i);
-            a.SetValue(indexAndValue[indexAndValue.Length - 1], indices);
+
+            Type elm = a.GetType().GetElementType()!;
+            a.SetValue(Converter.Convert(indexAndValue[indexAndValue.Length - 1], elm), indices);
         }
 
         [SpecialName]
--- a/Src/IronPython/Runtime/Operations/ArrayOps.cs
+++ b/Src/IronPython/Runtime/Operations/ArrayOps.cs
@@ -180,7 +180,7 @@ namespace IronPython.Runtime.Operations {
 
             for (int i = 0; i < indices.Length; i++) indices[i] += a.GetLowerBound(i);
 
-            Type elm = a.GetType().GetElementType()!;
+            Type elm = t.GetElementType()!;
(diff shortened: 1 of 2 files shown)


Src/IronPython/Runtime/Operations/ArrayOps.cs
```diff
@@ -180,7 +179,9 @@ public static class ArrayOps {
             if (a.Rank != args.Length) throw PythonOps.ValueError("bad dimensions for array, got {0} expected {1}", args.Length, a.Rank);
 
             for (int i = 0; i < indices.Length; i++) indices[i] += a.GetLowerBound(i);
-            a.SetValue(indexAndValue[indexAndValue.Length - 1], indices);
+
+            Type elm = a.GetType().GetElementType()!;
```
slozier, line 183: Minor nit: could use `t.GetElementType()` (or we could get rid of `t`).

slozier (approved):
Looks good to me!

Src/IronPython/Runtime/Operations/ArrayOps.cs
BCSharp: Oh yeah, indeed. I've just mindlessly copied the line from the next overload (got distracted by error handling here; see incoming issue report)

Issue #1295, opened 2021-09-10
Cleanup and add tutorials from main repo


slozier:
Looks like there are some samples at https://github.com/IronLanguages/main/tree/ipy-2.7-maint/Languages/IronPython/Public which never made it over to the repo. Might be worth seeing if there's anything salvageable.

slozier:
Also samples: https://github.com/jdhardy/ironpython/tree/master/Languages/IronPython/Samples

PR #1329, opened 2022-02-26
PEP 237: int/long unification

### LINKED ISSUE #894: Get rid of __long__ throughout the codebase
`__long__` is no longer used in Python 3

BCSharp:
This PR implements PEP 237, as discussed under #52. I have tried to keep it functionally self-contained (i.e. working consistently) but still as small as possible. As a result, there is still a dozen or so smaller updates coming up, mostly cleanup (notably, in `LongOps`/`IntOps`). Therefore I checked off PEP 237 as implemented in "WhatsNewInPython30" but want to keep #52 open for a while until everything gets into the mainline.

Despite that, it is still a sizable PR, though most of the changes are in the tests. The old tests used `long` profusely, which was aliased to `int`. This made the tests passing, but since `int` was `Int32`, `BigInteger` codepaths were not really well tested and the tests were often testing just the same as for old `int`. I've reviewed each case of `long` usage one by one and replaced `long` with either `int` or `big`, which is a function creating `BigInteger` instances for small integer values that normally would be `Int32`. Some tests that became obsolete are removed, and some new tests are added, but in most cases the existing tests are repurposed to test both `Int32` and `BigInteger`.

Before going full scale with this implementation, I did some prototyping of the remaining possible scenarios. All together, I've considered the following scenarios:

1. `int` is pure `BigInteger` and `Int32` is treated like all other unmanaged .NET types (e.g. `Int64`).
2. `int` type is `BigInteger` but instances of both `BigInteger` and `Int32` are treated as `int` for performance reasons (this proposal).
3. `int` type stays `Int32` but instances of both `BigInteger` and `Int32` are treated as `int` for performance reasons.
4. There are two separated Python types for `BigInteger` and `Int32` but both named `int` and further indistinguishable from the Python level. 

After going back and forth between them, I am happy to conclude that the chosen scenario seems to be the best choice. Each of the scenarios has some strengths and weaknesses, but scenario 2 seems like the best compromise.

For instance, scenario 1 is the most clean and logical, but gives in some performance and is not so convenient for interop, since lots of .NET API return just `Int32` and not `BigInteger`.

Scenario 3 comes close and has the advantage that the existing IronPython 2.x code would be easier to migrate to 3.x, but is difficult to keep free of surprises, mostly because not all of `int` instances would fit in objects/collections/generic methods strongly typed for `int`. I wouldn't even try to explain it to the people.

Scenario 4 becomes convoluted to get full compatibility with CPython.

This leaves scenario 2 which is fairly clean, straightforward and (mostly) free of surprises. For the migration, one must remember that Python 2.x `long` is renamed to `int` in Python 3.x, and Python 2.x `int` is "renamed" to `System.Int32`. In practice it seems that the only times that the developer has to pay attention to that distinction is when dealing with generics, since using `int` will imply `BigInteger` rather than `Int32`.

There is one regression: instances of subclasses of `int` behave differently during overloaded method resolution. The root cause is not so much this PR as the fact that the method resolution with arguments of type `BigInteger` and `Int32` works differently. I think something is wrong here. For instance, given argument `BigInteger(-200)`, and a method group with overloads taking either `Int32` or `Byte`, why would the resolver choose the latter one and then choke on it with an overflow exception? The tests for those cases are in `test_methodbinder2`, which, according to the code comment, are originally algorithmically generated, but maybe never got properly reviewed and simply froze incorrect behaviour. In any case, this is one of the loose ends I want to investigate in a separate PR, so for now the `myint` tests in that test module are commented out.

Incidentally, this PR also resolves #894.


Tests/test_bool.py
```diff
@@ -90,8 +88,6 @@ class MyLong(long): pass
         cases = []
         cases += [(ClassWithBool(x), y) for x, y in bool_cases]
         cases += [(ClassWithLen(x), y) for x, y in len_cases]
-        cases += [(ClassWithLen(long(x)), y) for x, y in len_cases if isinstance(x, int)]
```
slozier, line 93: No value in testing with `big(x)` and `myint(x)`?

Tests/test_class.py
```diff
@@ -35,7 +33,9 @@ def test_common_attributes(self):
             # Modifying __class__ causes a TypeError
             self.assertRaises(TypeError, i.__setattr__, "__class__")
 
-            self.assertEqual(type(i), i.__getattribute__("__class__"))
+            if i != 0:
```
slozier, line 36: I thought I saw some code handling `__class__` . Do we need to skip?

Tests/test_dlrkwarg.py
```diff
@@ -4,7 +4,7 @@
 
 import unittest
 
-from iptest import IronPythonTestCase, is_cli, is_mono, is_netcoreapp, is_posix, long, run_test, skipUnlessIronPython
+from iptest import IronPythonTestCase, is_cli, is_mono, is_netcoreapp, is_posix,  run_test, skipUnlessIronPython
```
slozier, line 7: Nit: leftover space.

Tests/test_ironmath.py
```diff
@@ -108,31 +109,31 @@ def test_big_2(self):
...
-            self.assertRaises(OverflowError,getattr(BigInteger(-1 - x ** a ), m))
+            self.assertRaises(OverflowError,getattr((2 ** a ).ToBigInteger(), m))
+            self.assertRaises(OverflowError,getattr((-1 - x ** a ).ToBigInteger(), m))
 
     def test_complex(self):
         from System.Numerics import BigInteger, Complex
         self.assertEqual(
             Complex.Add(
-                Complex(BigInteger(long(9999)), -1234),
+                Complex((big(9999)), -1234),
```
slozier, line 134: Nit: extra parens

Tests/test_methodbinder1.py
```diff
@@ -11,28 +11,32 @@
...
 from iptest.type_util import *
 
+from System import Int32
+
 myint1,     myint2      = myint(20),    myint(-20)
-mylong1,    mylong2     = mylong(3),    mylong(-4)
 myfloat1,   myfloat2    = myfloat(4.5), myfloat(-4.5)
 mycomplex1              = mycomplex(3)
 
+def bint(s):
```
slozier, line 20: Maybe `def bint(*s): ...` would make a nicer call site?

Tests/test_methodbinder1.py
```diff
@@ -919,11 +940,11 @@ class UserObject(object): pass
             # simple single type tests, no constraints
             # public static PythonType M0<T>(T x) -> pytypeof(T)
             # target method,   args,                            Result,     KeywordCall,      Exception
-            (target.M0,        (1, ),                          int,        True,             None),
+            (target.M0,        (1, ),                          Int32,      True,             None),
+            (target.M0,        ((1).ToBigInteger(), ),         int,        True,             None),
```
slozier, line 944: Use `bint`?

Tests/test_methodbinder1.py
```diff
@@ -968,21 +995,28 @@ class UserObject(object): pass
...
             (target.M7,        ([], ),                          list,       False,           None),
             (target.M7,        (objArray, ),                    type(objArray),False,        None),
         
             # simple dependent constraints
             # public static PythonTuple M8<T0, T1>(T0 x, T1 y) where T0 : T1 -> (pytype(T0), pytype(T1))     
-            (target.M8,        (1, 2),                         (int, int),   False,           None),
+            (target.M8,        (1, 2),                         (Int32, Int32), False,           None),
+            (target.M8,        bint((1, 2)),                   (int, int),   False,           None),
+            (target.M8,        (big(1), 2),                    None,         False,           TypeError),
+            (target.M8,        (1, big(2)),                     None,         False,           TypeError),
```
slozier, line 1015: Nit: alignment

Src/IronPython/Runtime/ConversionWrappers.cs
```diff
@@ -197,10 +250,35 @@ public class DictionaryGenericWrapper<K, V> : IDictionary<K, V> {
...
+                val = new BigInteger(i32);
+            }
+            return (V)val;
+        }
+        private static bool IsValidKey32(K key, out object key32) {
+            if (IsBigIntWrapperK && key is BigInteger bi && bi >= int.MinValue && bi <= int.MaxValue) {
+                key32 = (int)bi;
+                return true;
+            }
+            key32 = 0;
```
slozier, line 272: `null`?

slozier (commented):
Alright, turns out I had a bit more time. Here are some more comments after my first pass.

Tests/test_metaclass.py
```diff
@@ -151,7 +151,7 @@ class A2(object): pass
         class B1(metaclass=dash_attributes): pass
         class B2(object, metaclass=dash_attributes): pass
 
-        meta = lambda *args: 100
+        meta = lambda *args: big(100)
```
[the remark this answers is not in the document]
BCSharp: OK

Src/IronPython/Modules/Builtin.cs
```diff
@@ -656,7 +656,7 @@ public static object eval(CodeContext/*!*/ context, [NotNull]FunctionCode code,
             return (BigInteger)res;
         }
 
-        public static PythonType @int => DynamicHelpers.GetPythonTypeFromType(typeof(int));
+        public static PythonType @int => DynamicHelpers.GetPythonTypeFromType(typeof(BigInteger));
```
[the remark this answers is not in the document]
BCSharp: Good point. I'll add it to my follow-up list.

Src/IronPython/Lib/iptest/type_util.py
```diff
@@ -44,7 +52,8 @@ def remove_clr_specific_attrs(attr_list):
     
     # CLR array shortcut
     array_cli       = System.Array
-    array_int       = System.Array[int]
+    array_int       = System.Array[System.Int32]
```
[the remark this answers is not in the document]
BCSharp: I know, me too. I didn't notice [Upgrading from IronPython 2 to 3](https://github.com/IronLanguages/ironpython3/blob/master/Documentation/upgrading-from-ipy2.md#upgrading-from-ironpython-2-to-3) before. It will put a comprehensive note there (separate PR).

Src/IronPython/Runtime/Operations/LongOps.cs
```diff
@@ -14,106 +14,218 @@
...
-        public static object __new__(CodeContext context, PythonType cls, string s, int radix) {
-            var value = LiteralParser.ParseBigIntegerSign(s, radix, Int32Ops.FindStart(s, radix));
 
-            if (cls == TypeCache.BigInteger) {
-                return value;
-            } else {
-                return cls.CreateInstance(context, value);
+        #region Constructors
+
+        private static object FastNew(CodeContext/*!*/ context, object o, int @base = 10) {
```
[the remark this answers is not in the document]
BCSharp: Yes, I was planning to handle it in the follow-up `LongOps`/`IntOps` unification. It is kind of a separate issue since there are several ways of doing it. The idea I had was to make `BigIntegerOps` the main version and call it from `Int32Ops` as appropriate. This also applies to other methods in `Int32Ops`.

Src/IronPythonTest/EngineTest.cs
```diff
@@ -1079,7 +1077,7 @@ class ns(object):
     def __int__(self): return 42
     def __float__(self): return 42.0
     def __str__(self): return 'Python'
-    def __long__(self): return long(42)
+    def __long__(self): raise TypeError
```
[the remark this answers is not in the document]
BCSharp: I'm fine with removing it. I kept it here because it is a test and wanted to make sure that no code path calls `__long__` anymore. Although I grepped for `__long__`, sometimes the reflection code composes a method name from parts, so you never know...

Tests/modules/misc/test_math.py
```diff
@@ -412,12 +412,12 @@ def test_rounding(self):
         self.assertEqualAndCheckType((-2.4).__round__(), -2, int)
         self.assertEqualAndCheckType((-2.5).__round__(), -2, int)
         self.assertEqualAndCheckType((-2.6).__round__(), -3, int)
-        self.assertEqualAndCheckType(sys.float_info.max.__round__(), long(sys.float_info.max), long)
-        self.assertEqualAndCheckType(-sys.float_info.max.__round__(), long(-sys.float_info.max), long)
+        self.assertEqualAndCheckType(sys.float_info.max.__round__(), big(int(sys.float_info.max)), int)
```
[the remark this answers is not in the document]
BCSharp: No. its a no-op here.

Tests/modules/misc/test_math.py
```diff
@@ -615,43 +615,45 @@ def test_long_log(self):
...
+        self.assertRaises(ValueError, math.log, big(2), -1e666)
+        self.assertRaises(ZeroDivisionError, math.log, big(2), 1.0)
 
         #Make sure that an object is converted to float before being passed into log funcs
         class N(object):
             def __float__(self):
                 return 10.0
+            def __int__(self):
+                return 100
             def __long__(self):
```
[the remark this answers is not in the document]
BCSharp: OK. I kept it here out of abundance of caution.

Tests/test_bigint.py
```diff
@@ -46,14 +46,10 @@ def test_misc(self):
...
         # Test hex conversions. CPython 2.5 uses capital L, lowercase letters a...f)
-        s = hex(long(27))  # 0x1b
+        s = hex(big(27))  # 0x1b
         self.assertTrue(s == "0x1b", "27: Expect lowercase digits. Received: %s." % (s));
 
-        s = hex(-long(27))
+        s = hex(-big(27))
         self.assertTrue(s == "-0x1b", "-27: Expect lowercase digits. Received: %s." % (s));
 
-    def test_negative_misc(self):
```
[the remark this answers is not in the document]
BCSharp: I thought so, but now I can't find it anywhere (not even in StdLib). I will add the test back.

Tests/test_bool.py
BCSharp: I suppose there is. There is already a test with a big int (to test overflow) but not a successful conversion.

Tests/test_class.py
BCSharp: NO! Good catch!

I have skipped this test because I had originally intended `(1).__class` to be `Int32` as a convenient way to check the actual type without importing `System`, but it turned out it has to be `int` because StdLib assumes it is always so for integers (see: `@functools.singledispatch`). I guess I forgot about this test afterwards.

Tests/test_methodbinder1.py
BCSharp: OK

BCSharp: This is also a good moment to come up with a nicer name...

slozier (approved):
Looks good to me!

Side note, guess we'll have to update https://ironpython.net/documentation/dotnet/dotnet.html#mapping-between-python-builtin-types-and-net-types

Tests/test_tuple.py
```diff
@@ -47,13 +47,13 @@ def test_add_mul(self):
         self.assertEqual((1,2,3) * 2, (1,2,3,1,2,3))
         self.assertEqual(2 * (1,2,3), (1,2,3,1,2,3))
 
-        class mylong(long): pass
+        class mylong(int): pass
```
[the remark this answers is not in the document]
BCSharp: I thought about it too but tried to restraint myself from too much cleanup in this PR to keep is as small as possible, but also because the idea of cleaning up tests can become a big distraction, there is so much that can be cleaned up and it is tempting.

In this particular case, if you just object to the name `mylong`, I have planned a pass on that in my follow ups. Since Python `long` is gone, `long` can only be read as `Int64` and better not used at all except as a C# type.

If you meant using `myint` from `type_util`, then I was not planning of using it here, because the test also defines/uses `mylong2` and I think it is more explicit about the test intentions to have both types defined side-by-side. So here probably I would use names `myint1`/`myint2` instead but keep the class definitions.

Src/IronPython/Runtime/ConversionWrappers.cs
```diff
@@ -5,18 +5,28 @@
...
+        private static readonly bool IsBigIntWrapper = typeof(T) == typeof(BigInteger) || typeof(T) == typeof(BigInteger?);
 
         public ListGenericWrapper(IList<object> value) { _value = value; }
 
         #region IList<T> Members
 
         public int IndexOf(T item) {
-            return _value.IndexOf(item);
+            int pos = _value.IndexOf(item);
+            if (IsBigIntWrapper && item is BigInteger bi && bi >= int.MinValue && bi <= int.MaxValue) {
```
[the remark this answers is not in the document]
BCSharp: I'm not sure it will get inlined or even compiled in an optimal way. For instance, even if `T` is `BigInteger`, going from `T item` to `BigInteger bi` likely involves boxing, similarly (in other places) casting a `BigInteger` result to `T`. I have put on my follow-up list to look into the prefromance of the conversion wrappers, but right now, short of creating dedicated classes for `BigInteger` and `Nullable<BigInteger>` or dropping down to IL, I don't have any ideas.

Tests/test_file.py
```diff
@@ -721,7 +721,7 @@ def test_buffering_kwparam(self):
         with self.assertRaises(ValueError): # can't have unbuffered text I/O
             open(file=fname, mode='w', buffering=0)
 
-        self.assertRaisesMessage(TypeError, "expected int, got float" if is_cli else "integer argument expected, got float",
+        self.assertRaisesMessage(TypeError, "expected Int32, got float" if is_cli else "integer argument expected, got float",
```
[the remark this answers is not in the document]
BCSharp: Interesting, I had the same reaction here. But I do like seeing the actual type in error messages from calling regular .NET methods, but then again how to tell the difference between them and methods used to implement Python operations? Maybe check if the method comes from one of IronPython assemblies?

Frankly at this moment, the whole error message creation in DLR is half-broken and would benefit from a redesign, though it is not high on my list of interests... If it comes to that point, I can keep this case in mind.

BCSharp:
> Looks good to me!
> 
> Side note, guess we'll have to update https://ironpython.net/documentation/dotnet/dotnet.html#mapping-between-python-builtin-types-and-net-types

Is this page generated from some source, like RST? If so, where is it? The style sheets look rather outdated. I would be OK to review and update the page, and probably learn a few things for myself in the process.

Another side note: the link "Tools" at the top of that page is broken.

And more side notes: I thought I saw somewhere an issue report about some documentation that didn't get properly migrated from project _main_ (or _ironpython2_) but now I can't locate it. Any ideas? If indeed there is some documentation missing, it could be taken/migrated in the same cleanup action (though probably separate PRs).

slozier:
Unfortunately that particular page seems to be in [html](https://github.com/IronLanguages/ironpython.net/blob/gh-pages/documentation/dotnet/dotnet.html) instead of RST.  ~I wonder if it was RST once upon a time...~ Found this https://github.com/IronLanguages/main/blob/ipy-2.7-maint/Languages/IronPython/Public/Doc/dotnet-integration.rst which we might be able to salvage... The site hasn't gotten much attention in the past years.

Were you thinking of this issue? https://github.com/IronLanguages/ironpython3/issues/1295

BCSharp:
> Were you thinking of this issue? #1295

Yes! Thank you.

PR #1332, opened 2022-03-06
Eliminate Extensible<int>

BCSharp:
Follow-up on #1329 (issue #52).


slozier:
Looks good to me. Thanks!

Issue #104, opened 2016-11-13
PEP 498: Formatted string literals


moto-timo:
https://docs.python.org/3.6/whatsnew/3.6.html#pep-498-formatted-string-literals

https://www.python.org/dev/peps/pep-0498

https://docs.python.org/3.6/reference/lexical_analysis.html#f-strings

PR #1334, opened 2022-03-06
Formatted string literals (PEP 498)

slozier:
There are still some differences related to the `SyntaxError` messages which I may not fix in the PR (https://github.com/IronLanguages/ironpython3/issues/1346):
- `f'{lambda x:x}'` reports `invalid syntax` instead of `unexpected EOF while parsing`. Note that `eval('(lambda x)')` also reports the wrong message.
- `eval("f'''{\xa0}'''")` reports `invalid syntax` instead of `invalid character in identifier` (and the message is again different in 3.9).
- eval("fu''") reports `invalid syntax` instead of `unexpected EOF while parsing`.
- `f'\N'` is missing the `'unicodeescape' codec can't decode bytes in position .*` part of the message. (https://github.com/IronLanguages/ironpython3/issues/1348)

Other issues:
- `f'{3.14!s:10.10}'` doesn't produce the expected result but that's because the float repr (https://github.com/IronLanguages/ironpython2/issues/102).
- The CPython test_fstring comes from 3.6 and expects `eval(r"f'\{6*7}'")` to raise a `DeprecationWarning` (https://github.com/IronLanguages/ironpython3/issues/1343).
- Support for `yield`/`yield from` (https://github.com/IronLanguages/ironpython3/issues/1344).
- AST locations are all wrong (https://github.com/IronLanguages/ironpython3/issues/1345).

Resolves https://github.com/IronLanguages/ironpython3/issues/104


BCSharp:
F-strings? Awesome!

Are you going to support `yield`/`yield from` too?

slozier:
> F-strings? Awesome!

Yeah, one of the few 3.5+ features that I use regularly so thought I'd give it a go.

> Are you going to support `yield`/`yield from` too?

Have not given it any thought (didn't even know it was a thing until I tried running the unit tests😄). Will take a look at what's involved once I get the "basic" stuff working.

slozier:
@BCSharp No idea if you have any interest in reviewing this - but I'm sure it could use a good look-over. I guess the one open issue that's bugging me the most (and that falls in an area you looked at before) is the one with the bad unicode escapes (e.g. `f"\N"` or `f"\U1234"`) resulting in `encoding problem`. So if you have any ideas about that (or anything else) I'd be glad to hear them.

BCSharp:
Yes, I will look at it today. I was kind of looking along as you went but not very thoroughly. The remaining outstanding issues seem less consequential (with maybe the exception of AST locations) since they are mostly error messages. I personally never used `yield` in an f-string, but have used the "="-modifier, though it is from a higher Python version (3.8).

BCSharp:
Looks good to me!

> I guess the one open issue that's bugging me the most (and that falls in an area you looked at before) is the one with the bad unicode escapes (e.g. `f"\N"` or `f"\U1234"`) resulting in `encoding problem`

Well, there are a couple things going on here. First, `Parser.BadSourceError` handles conversion from `DecoderFallbackException` to `SyntaxError`. It is filled with heuristics to match CPython messages as close as possible, though it is not perfect because CPython reports a slightly different message if a Python file is imported rather than run on command line. Discrepancy considered benign. So I would modify this part to detect your f-string decoding exception and handle is as you wish.

The second issue is in `LiteralParset.TryParseFString.handleUnicodeError`. It uses `PythonExceptions.CreatePythonThrowable` from `UnicodeDecodeError`, but when `UnicodeDecodeError` is converted to a CLR exception it only takes the first argument as the message (while the most informative message is the last argument). Perhaps this is a bug but you do not need to create a Python error here, since the it will be caught in  `Parser.BadSourceError` and converted to `SyntaxError`, you can simply throw `PythonOps.UnicodeDecodeError`.

PR #1339, opened 2022-03-08
Restore locale after testing

BCSharp:
Prevents bogus test failures in subsequent tests under some circumstances (e.g. locale "en_CA" and .NET 6.0).


slozier (approved):
Looks good to me!

PR #1340, opened 2022-03-10
Fix strptime

BCSharp:
Fun fact: since .NET 5.0, AM/PM in Canada becomes a.m./p.m. CPython doesn't buy it.


slozier (approved):
Looks good to me! (though it's localization stuff so I don't know what's going on) :smile:

PR #1342, opened 2022-03-11
BytesWarning on bytes/str comparison

### LINKED ISSUE #1327: CPython.test_bytes fails with -bb
Noticed a warning that `test_compare` in `CPython.test_bytes` was being skipped because it requires `BytesWarning` to be enabled. When running with the option enabled it fails because `b'' == ''` (for example) does not warn.

BCSharp:
It's been on my todo list for too long. Resolves #1327.


slozier (approved):
Thanks!

PR #1347, opened 2022-03-12
Cleanup of constructors in BigIntegerOps/IntOps

BCSharp:

### DIFF
--- a/Src/IronPython/Runtime/Operations/BigIntegerOps.cs
+++ b/Src/IronPython/Runtime/Operations/BigIntegerOps.cs
@@ -184,9 +184,7 @@ namespace IronPython.Runtime.Operations {
         private static object ReturnObject(CodeContext context, PythonType cls, object value)
             => cls == TypeCache.BigInteger ? value : cls.CreateInstance(context, value);
 
-        #endregion
-
-        internal static int FindStart(string s, int radix) {
+        private static int FindStart(string s, int radix) {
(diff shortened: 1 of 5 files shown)


BCSharp:
@slozier do you know what BigIntegerV2 is and do we still need to support it? `BigIntegerOps` contains a bunch of `PythonHidden`, public methods under the "Backwards compatibility with BigIntegerV2" banner. AFACS, the history goes beyond IronPython2 first GitHub commit. If we can get rid of them, it will simplify alignment with `Int32Ops`. Otherwise, all this "compatibility" stuff has to be replicated in `Int32Ops` or we accept the fact that different instances of `int` may have different methods.

slozier:
>  do you know what BigIntegerV2 is and do we still need to support it?

BigInteger was added to .NET in .NET Framework 4.0 so I assume BigIntegerV2 refers to the previous baked-in DLR [BigInteger](https://github.com/IronLanguages/main/blob/ipy-2.7-maint/Runtime/Microsoft.Dynamic/Math/BigIntegerV2.cs) class. Seems like they added those to preserve the api surface that V2 provided. Feel free to yank it.

BCSharp:
Okay, I'm going to remove BigIntegerV2 API in a separate PR, in case somebody wants to look in up in git history.

slozier (approved):
Looks good to me!

PR #1354, opened 2022-03-15
Define constructors for System.Char

BCSharp:
For some reasons, `System.Char` was lacking constructors. It made it convoluted to create instances of this type:

```pycon
>>> from System import Char, IConvertible
>>> Char()
TypeError: Char() takes at least 2147483647 arguments (0 given)
>>> Char(65)
TypeError: cannot create instances of Char
>>> IConvertible.ToChar(65, None)   
A
>>> type(IConvertible.ToChar(65, None))
<class 'Char'>
```

The constructors bring `Char` more in line with other .NET primitive value types:

```pycon
>>> from System import Char
>>> Char('A')
A
>>> Char(65)
A
```

### DIFF
--- a/Src/IronPython/Runtime/Operations/CharOps.cs
+++ b/Src/IronPython/Runtime/Operations/CharOps.cs
@@ -7,6 +7,8 @@
 using System.Runtime.CompilerServices;
 
 using Microsoft.Scripting.Runtime;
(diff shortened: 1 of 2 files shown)


slozier (approved):
Looks good to me.

PR #1357, opened 2022-03-16
Remove obsolete BigIntegerV2 API from BigIntegerOps

BCSharp:
Follow-up on #1347 and https://github.com/IronLanguages/ironpython3/pull/1329 (issue https://github.com/IronLanguages/ironpython3/issues/52).

Special point of interest: `op_Division`.


slozier (approved):
Thanks!

PR #1358, opened 2022-03-18
.NET 7.0

slozier:


slozier:
@BCSharp Are you able to reproduce the failure on macOS with .NET 7? I think it's occurring in `test_re_stdlib` but I don't have a macOS system to test with. If it's a bug with .NET 7 would be good to report it before they do their release.

BCSharp:
Yes, I was able to reproduce the failure. The test crashes on `test_big_codesize` with a stack overflow. To be more specific, it crashes on the first `match` on the compiled expression. However, replacing `match` with `re.search`, which uses uncompiled expression, does not trigger the overflow. Also, the overflow only happens when run by the testrunner, not if the test is executed directly by the ipy console program. When I see an effect like this, my first suspect is IronPython code rewriting. Not that it is incorrect, but may simply generate inefficient code. For instance, reducing `range(10000)` to `range(7000)` and modifying the match patterns accordingly, is sufficient to make the test passing on my machine. On the other hand, increasing the range to, say, 20000, makes the test crash even on .NET 6.0.

All in all, I don't have a feeling that it is caused by a problem in .NET 7.0, more likely, the implementation in .NET 7.0 changed somehow that it uses more method calls on the stack than before, which, when accumulated, push the process over the edge.

To make the test pass temporarily, you may set `IsolationLevel=PROCESS`. Unfortunately, the root cause is elusive and will require more time to investigate, which I don't have at the moment.

slozier:
Thanks for the tip. I was able to reproduce the failure on Windows by bumping the range to 25000 and on .NET 6 (with 50000). I managed to get the same stack overflow running directly in C# so that rules out an IronPython issue:
```c#
var sb = new StringBuilder();
sb.Append(0);
for(int i=1; i<25000; i++)
{
    sb.Append('|');
    sb.Append(i);
}

Regex.Match("1000", sb.ToString(), RegexOptions.Compiled);
```

But as you say we must be right near the limit on macOS and IronPython's stack is pushing it over the edge. I will try `IsolationLevel=PROCESS` to see if it helps.

DamianReeves:
Any timeline around when we can expect net7.0 support?

slozier:
@DamianReeves This PR only enables the test suite. Is there a problem running the existing (.NET 6) DLLs on .NET 7?

PR #1361, opened 2022-03-19
Fix serialization of non-int-based enum instances

BCSharp:
Also applies to pickling.

### DIFF
--- a/Src/IronPython/Runtime/ClrModule.cs
+++ b/Src/IronPython/Runtime/ClrModule.cs
@@ -1146,7 +1146,8 @@ import Namespace.")]
             }
 
             string data, format;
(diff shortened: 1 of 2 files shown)


slozier (approved):
Seems fine. I'm surprised there isn't a better way to check for primitive types (e.g. `type.IsPrimitive`).

BCSharp:
There is. It's `type.IsPrimitive` 😄

BCSharp:
Well, it turns out the test `!type.IsPrimitive` is not right by itself. `decimal` and `DBNull` are not primitive. I'll submita PR soon.

slozier:
Ah right, didn't notice those were the list.

Issue #524, opened 2019-01-27
TypeError when __index__ returns a subclass of int


slozier:
Note: For the record, I don't think we should spend any effort adding support for this since it's deprecated. I'm just filing it so I have an issue to reference against.

The following throws a `TypeError` in IronPython, but gives a `DeprecationWarning` in CPython 3.4.
```Python
class BadInt:
    def __index__(self):
        return True

operator.index(BadInt())
```

Edit: This comes from `test_index`.

PR #1365, opened 2022-03-20
Cleanup indexing

slozier:
Changes the behavior of indexing (e.g. `operator.index`) to always return an `int` (`Int32`/`BigInteger`) even for subclasses of int. This simplifies some code paths and is the behavior since 3.10:
> The result of [PyNumber_Index()](https://docs.python.org/3/c-api/number.html#c.PyNumber_Index) now always has exact type [int](https://docs.python.org/3/library/functions.html#int). Previously, the result could have been an instance of a subclass of int. (Contributed by Serhiy Storchaka in [bpo-40792](https://bugs.python.org/issue40792).)

Example:
```py
import _operator
class myint(int): pass

class myindex:
    def __init__(self, val):
        self.val = val
    def __index__(self):
        return self.val

assert type(_operator.index(myindex(myint(1)))) == int
```

Resolves https://github.com/IronLanguages/ironpython3/issues/524

### DIFF
--- a/Src/IronPython.Modules/_opcode.cs
+++ b/Src/IronPython.Modules/_opcode.cs
@@ -165,7 +165,7 @@ namespace IronPython.Modules {
                     throw PythonOps.ValueError("stack_effect: opcode requires oparg but oparg was not specified");
                 }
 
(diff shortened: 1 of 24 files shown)


BCSharp:
Why reverting whitespace changes?

slozier:
Reverted some changes and the only thing left over in those files were whitespace changes.

PR #1366, opened 2022-03-20
Finish IConvertible members in BigIntegerOps

BCSharp:
Follow-up on https://github.com/IronLanguages/ironpython3/pull/1347 and https://github.com/IronLanguages/ironpython3/pull/1329 (issue https://github.com/IronLanguages/ironpython3/issues/52).

### DIFF
--- a/Src/IronPython/Runtime/Operations/BigIntegerOps.cs
+++ b/Src/IronPython/Runtime/Operations/BigIntegerOps.cs
@@ -655,7 +655,7 @@ namespace IronPython.Runtime.Operations {
 
         #endregion
 
-        #region Mimic some IConvertible members
+        #region Mimic IConvertible members
 
         [PythonHidden]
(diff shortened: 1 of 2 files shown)


slozier (approved):
Seems fine to me.

I guess this is for parity with the Int32 type? Looks like we used to hide all the IConvertible methods from the int type before?

BCSharp:
> I guess this is for parity with the Int32 type?

Yes.

> Looks like we used to hide all the IConvertible methods from the int type before?

Yes, it changed with #1360. Right now, .NET interfaces are either always visible on non-Python-types, or never visible/resolved on Python types, there is nothing in between. I wish there were a mechanism to make them visible/resolvable only after `import System`, just like what `[PythonHidden]` does for methods.

I am thinking of adding such a mechanism to the type system but have no concrete ideas yet. Maybe I just special-case it for `Int32`.

PR #1368, opened 2022-03-21
Separate FloatOps.Generated from IntOps.Generated

BCSharp:


Src/Scripts/generate_alltypes.py
```diff
@@ -164,7 +168,7 @@ def gen_unaryops(cw, ty):
         cw.write(identity_method, method_name="Abs")
         cw.write(unsigned_negate_or_invert, method_name="OnesComplement")
 
-    if (ty.type is not complex) and (ty.type is not bigint):
+    if ty.name not in ['BigInteger', 'Complex']:
```
slozier, line 171: Do we need this check?

BCSharp: No, and such checks are done in more places than here. I assume at some time this script was used to generate code for `BigInteger` and `Complex` too. I was hesitating removing those checks in case code generation for those types comes back (?), but am ready to remove them if you green-light it.

In any case, I plan another pass though this script to modernize the code it generates (separate PR).

slozier (approved):

Src/Scripts/generate_alltypes.py
slozier: Feel free to yank any dead code. If I had more time on my hands I'd probably look into trying to use C# Source Generators instead of this. 😃

PR #1373, opened 2022-03-23
Improve clr.Serialize

BCSharp:
Improvement on  #1361.

### DIFF
--- a/Src/IronPython/Runtime/ClrModule.cs
+++ b/Src/IronPython/Runtime/ClrModule.cs
@@ -1147,39 +1147,20 @@ import Namespace.")]
 
             string data, format;
             Type type = CompilerHelpers.GetType(self);
-            switch (type.GetTypeCode()) {
-                // For the primitive non-python types just do a simple serialization.
-                // Object of type Int32 normally do not go though this code path when pickled
-                // (that is, unless __reduce_ex__ is called directly), being handled as Python int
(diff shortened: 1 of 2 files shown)


slozier (approved):

Src/IronPython/Runtime/ClrModule.cs
```diff
@@ -1147,39 +1147,20 @@ private class PALHolder : MarshalByRefObject {
...
-
-                    data = self.ToString();
-                    format = type.FullName;
-                    break;
-
-                default:
-                    // something more complex, let the binary formatter handle it                    
-                    BinaryFormatter bf = new BinaryFormatter();
-                    MemoryStream stream = new MemoryStream();
+            if ((type.IsPrimitive && type.GetTypeCode() != TypeCode.Object)
```
slozier, line 1150: Are there primitive types with TypeCode.Object?

BCSharp: `IntPtr` and `UIntPtr`

slozier: Interesting, wonder why they made that choice...

BCSharp: If I had to guess, perhaps because those types do not implement `IConvertible`.

PR #1383, opened 2022-03-28
Use __index__ if needed to pack integer values in struct

### LINKED ISSUE #1381: struct.pack does not respect __index__ operator
Packing using `struct` fails if the argument to be packed is not convertible to the desired type using `Converter.TryConvertToXXX`. If the argument defines `__index__`, CPython uses this operator to obtain an integer value.

This affect all data getter helpers for integer types in `_struct.cs`, which do not call `PythonOps.TryToIndex`.

```python
import struct

class Index:
    def __index__(self):
        return 1

assert struct.pack("i", Index()), b'\x01\x00\x00\x00'
```

BCSharp:
Resolves #1381.

### DIFF
--- a/Src/IronPython.Modules/_struct.cs
+++ b/Src/IronPython.Modules/_struct.cs
@@ -141,12 +141,12 @@ namespace IronPython.Modules {
                             break;
                         case FormatType.UnsignedInt:
                             for (int j = 0; j < curFormat.Count; j++) {
-                                WriteUInt(res, _isLittleEndian, GetULongValue(context, curObj++, values, "unsigned int"));
+                                WriteUInt(res, _isLittleEndian, GetULongValue(context, curObj++, values));
                             }
                             break;
                         case FormatType.UnsignedLong:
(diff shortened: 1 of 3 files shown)


slozier (approved):
Looks good to me. Interesting that CPython has two different error messages for some of these.

Random thought, I wonder if trying to cast the `BigInteger` and catching the failure would be more performant than checking the range.

BCSharp:
Intuitively, a direct cast should be faster for cases without overflow, but  I was curious by how much, so I did some measurements.

The results show that going through `AstInt32`  adds ca. 550% to the execution of the cast (i.e., the whole operation is 6.5 times slower). About half of this overhead comes from the two `BigInteger` comparisons, and the other half from the static method call (and whatever else that method does).

That said, those things are extremely fast (and hard to measure accurately). Converting by calling `AstInt32` on my machine is less than 8 nanoseconds. So in the end, it is a matter of judgment if shaving off a few nanoseconds is worth increased code verbosity.

PR #1385, opened 2022-03-31
Fix complex.__new__

BCSharp:
A thought: move `TryToFloat` to `PythonOps`?

### DIFF
--- a/Src/IronPython/Runtime/Operations/FloatOps.cs
+++ b/Src/IronPython/Runtime/Operations/FloatOps.cs
@@ -72,17 +72,17 @@ namespace IronPython.Runtime.Operations {
             } else if (value is int i) {
                 result = i;
             } else if (value is BigInteger bi) {
-                result = BigIntegerOps.ConvertToDouble(bi);
+                result = BigIntegerOps.ToDouble(bi);
             } else if (TryInvokeFloat(context, value, out result)) {
                 // pass
             } else if (value is Extensible<double> ed) {
                 result = ed.Value;
             } else if (value is Extensible<BigInteger> ebi) {
-                result = BigIntegerOps.ConvertToDouble(ebi.Value);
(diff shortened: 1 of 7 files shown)


Src/IronPython/Runtime/Operations/FloatOps.cs
```diff
@@ -53,39 +53,66 @@ public static partial class DoubleOps {
...
+            }
+        }
+
+        internal static bool TryToFloat(CodeContext context, object/*?*/ value, out double result) {
+            if (value is double d) {
+                result = d;
+            } else if (value is int i) {
+                result = i;
+            } else if (value is BigInteger bi) {
+                result = (double)bi;
```
slozier, line 75: I guess these cast is where testing fails. The BigInteger to double cast returns infinity instead of overflowing. There's a `MathUtils.ToFloat64` that throws (but we should consider a helper at the ipy level so we can match the error message).

BCSharp: Indeed. The fast track just cut too many corners. `__float__` and `__complex__` on `int` do it correctly, whith a call to `MathUtils.ToFloat64`. Funny thing, I started working on `complex.__new__` as a spin-off from an investigation whether `int.__complex__` can be safely removed (it can, assuming that `complex.__new__` works correctly) and then completely forgot about that.

I like the idea of our own helper class with a Python-specific error message.

BCSharp:
I looked into the CPython's code and roughly speaking the functionality of `TryToFloat` resides inside `PyNumber_Float`, except that the latter also does string parsing, so I suppose is more or less equivalent to `NewFloat`. `PyNumber_Float` is used in a few places but it seems that apart from of the `float` and `complex` constructors, only to parse a string. So I suppose no need to move `TryToFloat` to `PythonOps`.

slozier (approved):
Looks good to me!

PR #1390, opened 2022-04-04
Cleanup IntOps

BCSharp:
Follow-up on #1347 and https://github.com/IronLanguages/ironpython3/pull/1329 (issue https://github.com/IronLanguages/ironpython3/issues/52).

There are many changes in this PR but most of them is just reordering of methods in `IntOps` and `BigIntegerOps` to align them with the order in `IntOps.Generated.cs`. Makes browsing, comparing, and reviewing easier.

From the functional changes:

* in `IntOps`:
  * `__abs__` removed (was erroneous)
  * `__float__` removed (was redundant)
* in `BigIntegerOps`, to align with the other numeric Ops:
  * `__pos__` renamed `Plus` (although actually it is redundant)
  * `__abs__` renamed `Abs` (could have been internal)

There are possibly more redundant methods or methods that could have been internal, but I am not going to hunt them down.


slozier (approved):
Looks good to me.

Issue #1182, opened 2021-04-21
StackOverflowException with continue from try..except


slozier:
When using `continue` in the `except` clause of a try/except a `StackOverflowException` may occur. This seems to occur more frequently when running in 64-bit. Here's a repro which causes the issue somewhat consistently:
```py
def test_gh1628():
    def error():
        raise Exception()

    for _ in range(100000):
        try:
            error()
        except Exception:
            continue
```

I believe the issue only occurs when the code is interpreted. To reproduce consistently the following change can be applied to the DLR:
```C#
LightCompiler.DefaultCompilationThreshold = int.MaxValue;
```

The issue has been reported a number of times:
https://github.com/IronLanguages/ironpython2/issues/40
https://github.com/IronLanguages/ironpython2/issues/42
https://github.com/IronLanguages/ironpython2/issues/182

As well as a proposed fix:
https://github.com/IronLanguages/main/pull/1167

PR #1392, opened 2022-04-05
Implement to_bytes for all integer types

BCSharp:
Normally I would not add a Python-specific method to a non-Python type (except for dunder methods), but this method can be quite useful on integral types.

### DIFF
--- a/Src/IronPython/Runtime/Operations/BigIntegerOps.cs
+++ b/Src/IronPython/Runtime/Operations/BigIntegerOps.cs
@@ -749,7 +749,7 @@ namespace IronPython.Runtime.Operations {
                 val = -self;
             }
             string digits;
-            
+
             switch (spec.Type) {
                 case 'n':
                     CultureInfo culture = context.LanguageContext.NumericCulture;
@@ -932,9 +932,9 @@ namespace IronPython.Runtime.Operations {
             return __new__(context, type, val);
         }
 
-#endregion
+        #endregion
 
-#region Mimic IConvertible members
+        #region Mimic IConvertible members
 
         [PythonHidden]
         public static bool ToBoolean(BigInteger self, IFormatProvider? provider) {
@@ -1060,9 +1060,9 @@ namespace IronPython.Runtime.Operations {
             return TypeCode.Object;
         }
 
-#endregion
+        #endregion
 
-#region Helpers
+        #region Helpers
 
         internal static double ToDouble(BigInteger self) {
             // Unlike ConvertToDouble, this method produces a Python-specific overflow error messge.
@@ -1080,7 +1080,7 @@ namespace IronPython.Runtime.Operations {
             return ToDigits(val, 8, lowercase);
         }
 
-        internal static string ToBinary(BigInteger val) {            
+        internal static string ToBinary(BigInteger val) {
             string res = ToBinary(BigInteger.Abs(val), true, true);
             if (val.IsNegative()) {
                 res = "-" + res;
@@ -1092,7 +1092,7 @@ namespace IronPython.Runtime.Operations {
             Debug.Assert(!val.IsNegative());
 
             string digits = ToDigits(val, 2, lowercase);
-            
+
             if (includeType) {
                 digits = (lowercase ? "0b" : "0B") + digits;
             }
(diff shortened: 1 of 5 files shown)


slozier (approved):
Seems fine to me. Here are some random thoughts:

- I noticed that .NET Core has a `public byte[] ToByteArray (bool isUnsigned = false, bool isBigEndian = false);` overload. I wonder if it's worth having a branch and using that.
- Do you know if there is a significant performance difference between your `Int64`/`UInt64` implementations vs cast to `BigInteger` and using that implementation throughout?
- Wondering if we should create a dummy `[KeywordOnlyAttribute]` that we can use to start annotating.

BCSharp:
> * I noticed that .NET Core has a `public byte[] ToByteArray (bool isUnsigned = false, bool isBigEndian = false);` overload. I wonder if it's worth having a branch and using that.

I assume that by "having a branch" you mean `#if NETCOREAPP`.
Yes, this would be improvement in some scenarios, though I'd use `TryWriteBytes` which can write to a span, thus avoiding a potential array resize. Will commit an update, probably in two days.

> * Do you know if there is a significant performance difference between your `Int64`/`UInt64` implementations vs cast to `BigInteger` and using that implementation throughout?

Yes, especially on .NET Framework, `BigInteger.ToByteArray` is quite inefficient. It creates two temporary arrays before allocating the final one to return. And even that final one may have to be resized to satisfy the required length. I have optimized `BigIntegerOps.to_bytes` to call `BigInteger.ToByteArray` only once (by eliminating a call to `bit_length`, which also calls `ToByteArray`), but it is still far from how `(U)Int64Ops.to_bytes` performs.

On .NET Core, it is less significant; they have optimized `BigInteger.ToByteArray`. But if a value does not fit in `Int32`, going through `BigInteger` will trigger a bits array allocation so still comparatively quite expensive. If a value does fit in `Int32` then this would have been acceptable (though still a few times more expensive than the current `to_bytes` implementation).

> * Wondering if we should create a dummy `[KeywordOnlyAttribute]` that we can use to start annotating.

After finishing int/long integration, my plan was to finish keyword-only support, as well as positional-only a.k.a non-keyword a.k.a prohibit-keyword support. I have some draft code in a stash, sitting there for over a year, waiting for some improvements in error reporting in DLR, but I don't see it happening so I better submit my code soon anyway. As long as the relevant places are commented with `keyword[ -]*only`, this is good enough for me. But don't hold me to that, if you feel like it, go ahead with keyword-only support.

Src/IronPython/Runtime/Operations/BigIntegerOps.cs
```diff
@@ -715,8 +715,17 @@ private static bool IsInt32(this BigInteger self)
             return 1;
         }
 
-        public static int bit_length(BigInteger self) {
+        public static object bit_length(BigInteger self) {
+#if NET
+            long length = self.Sign switch {
```
slozier, line 720: I guess this is needed since we don't have a .NET 5+ assembly for the DLR?

Src/IronPython/Runtime/Operations/BigIntegerOps.cs
```diff
@@ -897,9 +932,9 @@ private static bool IsInt32(this BigInteger self)
             return __new__(context, type, val);
         }
 
-        #endregion
+#endregion
```
slozier, line 935: Editor accidentally messing with the alignment? Otherwise seems fine to me, feel free to merge.

Src/IronPython/Runtime/Operations/BigIntegerOps.cs
BCSharp: Do you think it should be moved to the DLR?

slozier: I'm assuming it produces exactly the same result as `MathUtils.BitLength` does, so from that point of view it would makes sense to have it there - but since this appears to be the only call site and we're not publishing a .NET 5+ DLR right now it's probably not worth it.

slozier:
Ahh, looks like `test_to_bytes_clrint` is hitting the intermittent https://github.com/IronLanguages/ironpython3/issues/1182 and causing a `StackOverflowException` during CI.

slozier:
Added a workaround for the stackoverflow in https://github.com/IronLanguages/ironpython3/pull/1397

BCSharp:
Funny thing, that's how I had it originally (modulo spelling error), and then I "simplified" it. Something learned.

PR #1393, opened 2022-04-05
Don't overflow on complex constructor

slozier:
Noticed this while reviewing https://github.com/IronLanguages/ironpython3/pull/1385


BCSharp (commented):
Glad you noticed that. While reviewing this PR, I've noticed that `ParseComplex` doesn't handle null chars correctly like `ParseFloat` does.

```pycon
>>> complex("1111\x00")
(1111+0j)
>>> complex("1111\x00j")
1111j
```

The above should raise `ValueError: complex() arg is a malformed string`. Perhaps replacing `ParseFloadNoCatch` in `ParseComplex` with `ParseFloat` would do the trick.

BCSharp (approved):

Issue #1395, opened 2022-04-08
.NET Overloads appearing in dir(0)


slozier:
Some of the .NET Overloads are now appearing in `dir(0)`:
>['CompareTo', 'Equals', 'GetHashCode', 'GetTypeCode', 'MaxValue', 'MinValue', 'Parse', 'ToBigInteger', 'ToBoolean', 'ToByte', 'ToChar', 'ToDateTime', 'ToDecimal', 'ToDouble', 'ToInt16', 'ToInt32', 'ToInt64', 'ToSByte', 'ToSingle', 'ToString', 'ToType', 'ToUInt16', 'ToUInt32', 'ToUInt64', 'TryParse', ...]

Expected behaviour is to not see them until someone steps out of the Python world (e.g. via `import System`).

@BCSharp I guess this is related to some of the recent `int` changes?

BCSharp:
I have it on my to-do list. The issue exists because now `Int32` does not fit in the dichotomy _Python type_ vs. _non-Python type_, which the `PythonType`/`PythonTypeOps` mechanisms support.

Ideally, I'd hide .NET members on `Int32` instances (until `import System`) but not on the actual `Int32` type. Such a distinction is not supported by the Python type mechanism and adding it would likely significantly increase the complexity of the code, so I am currently inclined not to go that far. Anyway, I going to submit my changes so far in a draft PR so that we can have a discussion supported by real code. @slozier

PR #1396, opened 2022-04-08
Hide .NET methods in Int32 until CLR visible

BCSharp:
Addresses #1395.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonTypeOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonTypeOps.cs
@@ -656,7 +656,17 @@ namespace IronPython.Runtime.Operations {
                         break;
                     }
                 }
-
+            } else if (type.IsAssignableFrom(typeof(int))) { // GH #52
+                // only show methods defined outside of int
+                 foreach (MethodInfo mi in methods) {
+                    Debug.Assert(!mi.DeclaringType.IsInterface || typeof(int).GetInterfaces().Contains(mi.DeclaringType));
+                    if (mi.DeclaringType == typeof(int) ||
+                        mi.DeclaringType.IsInterface ||
+                        PythonHiddenAttribute.IsHidden(mi)) {
+                        alwaysVisible = false;
+                        break;
+                    }
+                }
             }
             return alwaysVisible;
         }


BCSharp:
The fields are still visible. Will require additional handling to hide them.

BCSharp:
Stuff is hidden but still resolvable.

slozier:
I poked around the code a bit but did not come up with anything better than what you have here. The only other idea I had was to cut the interfaces from the `Int32` mro and add the members to `Int32Ops` with `PythonHidden` but that sort of feels like a step backwards?

BCSharp:
I first tried that too, but it felt like a somewhat convoluted way to go. Moreover, some use cases remained elusive, like:
```python
System.IComparable[System.Int32].CompareTo(1, 2)
```

BCSharp:
I've introduced a concept of a Python-supporting type in an attempt to make the changes a little less hacky. Currenly, `Int32` is the only case of such type so I think it does not warrant creating a new flull-blown category of types in `PythonBinder` (like `MakeExtensionTypes` and `MakeSystemTypes` do), but I can envision this concept being used more in the future. For instance, when [PEP 393](https://peps.python.org/pep-0393/) is implemented.

BCSharp:
Another idea: the name of `PythonBinder.IsExtendedType` is really confusing. Perhaps rename to `IsExtendedPythonType` (more accurate and clear) or `IsSystemType` (more consistent with the terminology used throughout `PythonBinder` but potentially confusing with .NET `System` types)?

slozier:
> Another idea: the name of `PythonBinder.IsExtendedType` is really confusing. Perhaps rename to `IsExtendedPythonType` (more accurate and clear) or `IsSystemType` (more consistent with the terminology used throughout `PythonBinder` but potentially confusing with .NET `System` types)?

I agree `PythonBinder.IsExtendedType` is a bit confusing since there are other methods like `PythonBinder.HasExtensionTypes` which mean something else. I think `IsExtendedPythonType` is probably fine - better to avoid `IsSytemType` since we have `PythonType.IsSystemType`.

slozier:
I must have enabled some unstable tests recently - getting a lot of failures...

Tests/test_int.py
```diff
@@ -3,10 +3,36 @@
...
+    def test_instance_set(self):
+        i = 1
+        j = big(1)
+        from System import Int32
+
+        self.assertSetEqual(set(dir(i)) - set(dir(j)), {'MaxValue', 'MinValue'})
+        self.assertSetEqual(set(dir(Int32)) - set(dir(int)), {'MaxValue', 'MinValue'})
+
+    @unittest.expectedFailure
+    def test_instance_set_todo(self):
```
slozier, line 28: Did you want to keep this todo in the PR?

slozier (approved):
Looks good to me!

Tests/test_int.py
BCSharp: Yes, so that the test fails when I fix the issue on one of the next PRs and I don't forget about this test.

PR #1399, opened 2022-04-13
Mimic BigInteger members on Int32

BCSharp:

Instances of `Int32` behave as if being instances of an anonymous subtype of `int`. Therefore, for the sake of type system consistency, any attributes of `int` should be also present on instances of both `int` and `Int32`. This was already the case for regular Python attributes, but not for Python-hidden attributes. This PR sets it straight. Arguably, there is little chance that the static attributes are going to be accessed through an instance rather than the type, but here they are to avoid surprises. Who knows, maybe some code-generating use case assumes that it is possible, as it should be.

Ideally I would like to see those Python-hidden properties and method only on `Int32` instances and not pollute the namespace of the `Int32` type. However, the existing resolver mechanism does not support such distinction and I am not convinced it is worth extra complexity implementing an additional mechanism just for that.

### DIFF
--- a/Src/IronPython/Runtime/Operations/IntOps.cs
+++ b/Src/IronPython/Runtime/Operations/IntOps.cs
@@ -466,16 +466,48 @@ namespace IronPython.Runtime.Operations {
 
 #if NETCOREAPP
         [PythonHidden]
+        public static byte[] ToByteArray(int self, bool isUnsigned = false, bool isBigEndian = false) => new BigInteger(self).ToByteArray(isUnsigned, isBigEndian);
+
+        [PythonHidden]
         public static int GetByteCount(int self, bool isUnsigned = false) => new BigInteger(self).GetByteCount(isUnsigned);
 
         [PythonHidden]
         public static bool TryWriteBytes(int self, Span<byte> destination, out int bytesWritten, bool isUnsigned = false, bool isBigEndian = false)
             => new BigInteger(self).TryWriteBytes(destination, out bytesWritten, isUnsigned, isBigEndian);
+
+#elif NETSTANDARD
+        [PythonHidden]
+        public static byte[] ToByteArray(int self, bool isUnsigned = false, bool isBigEndian = false) {
+            if (self < 0 && isUnsigned) throw new OverflowException("Negative values do not have an unsigned representation.");
+            byte[] bytes = ToByteArray(self);
+            int count = bytes.Length;
+            if (isUnsigned && count > 1 && bytes[count - 1] == 0) Array.Resize(ref bytes, count - 1);
+            if (isBigEndian) Array.Reverse(bytes);
+            return bytes;
+        }
+
+        [PythonHidden]
+        public static int GetByteCount(int self, bool isUnsigned = false) => ToByteArray(self, isUnsigned).Length;
+
+        [PythonHidden]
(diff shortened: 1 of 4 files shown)


Src/IronPython/Runtime/Operations/IntOps.cs
```diff
@@ -422,14 +419,155 @@ public static object from_bytes(CodeContext context, PythonType type, object byt
...
+        public static byte[] ToByteArray(int self, bool isUnsigned = false, bool isBigEndian = false) => new BigInteger(self).ToByteArray(isUnsigned, isBigEndian);
+
+        [PythonHidden]
+        public static int GetByteCount(int self, bool isUnsigned = false) => new BigInteger(self).GetByteCount(isUnsigned);
+
+        [PythonHidden]
+        public static bool TryWriteBytes(int self, Span<byte> destination, out int bytesWritten, bool isUnsigned = false, bool isBigEndian = false)
+            => new BigInteger(self).TryWriteBytes(destination, out bytesWritten, isUnsigned, isBigEndian);
+
+#elif NETSTANDARD
```
slozier, line 478: These feel a bit icky since depending on what runtime is using the .NET Standard assemblies you will get different methods. For example, if the .NET Standard assemblies are used with .NET Framework you'll might get extra methods on `Int32` based `int` which won't appear when backed by `BigInteger`.

Anyway, I don't have a good solution to propose right now so this is probably fine.

Tests/test_int.py
```diff
@@ -21,17 +20,22 @@ def test_instance_set(self):
...
-        self.assertSetEqual(set(dir(int)) - set(dir(Int32)), set())
+        if not is_mono:
+            self.assertSetEqual(set(dir(j)) - set(dir(i)), set())
+            self.assertSetEqual(set(dir(int)) - set(dir(Int32)), set())
+        else:
+            self.assertSetEqual(set(dir(j)) - set(dir(i)), {'GetByteCount', 'TryWriteBytes'})
+            self.assertSetEqual(set(dir(int)) - set(dir(Int32)), {'GetByteCount', 'TryWriteBytes'})
+
+        # these two assertions fail on IronPython compiled for .NET Standard
+        # if not iptest.is_netstandard: # no such check possible
```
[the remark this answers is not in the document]
BCSharp: OK, will give it a try.

Src/IronPython/Runtime/Operations/IntOps.cs
BCSharp: Indeed, but I too have no better ideas. However I am not too concerned about this case: from the perspective of the Python type system consistency it is totally acceptable for some instances to have some additional attributes that other instances or their type don't have.

On the other hand, it is not OK if a type has attributes which are missing on (some) instances (without any explicit deletion, that is). Unfortunately, this is now the case on Mono. After spending an hour or so trying to fix it, I decided to simply fix the tests on Mono and will submit an issue report once this PR is merged. Maybe it will be picked up at some point in the future (or Mono becomes obsolete by .NET vNext). My enthusiasm to work around Mono quirks is quite limited, and in any case, this particular issue is not on the top of the list.

slozier: > I decided to simply fix the tests on Mono and will submit an issue report once this PR is merged.

Fine by me. Mono is less relevant these days with .NET having good cross-platform support.

slozier (approved):

PR #1402, opened 2022-04-16
Remove ToBigInteger()

BCSharp:
It was mainly used for testing and there is a simpler alternative.

### DIFF
--- a/Src/IronPython/Lib/iptest/type_util.py
+++ b/Src/IronPython/Lib/iptest/type_util.py
@@ -6,15 +6,10 @@
 
 from .test_env import is_cli
 
-if is_cli:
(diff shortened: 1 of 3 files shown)


slozier (commented):
It seems to me that having a method that guarantees the underlying object is a `BigInteger` could be useful in the interop world. I can't think of a great real-world scenario, but for example trying to call a .NET method with an `object` argument and ensuring a BigInteger is passed it (why anyone would do that I don't know - but who knows). Obviously there are workarounds (such as the `int.__pos__` that you used) but they are not as discoverable.

BCSharp:
All right then, let’s keep it.

PR #1403, opened 2022-04-17
Rename ConvertToLong to ConvertToInt

BCSharp:

Since `__long__` is gone, to avoid confusion, `XxxConvertToLong` is renamed to `XxxConvertToInt`, while the old `XxxConvertToInt` is gone. Plus some cleanup. A couple of functional changes are fixes:

* If `__int__` returns a `BigInteger` instance, a conversion call site with return type `Int32` succeeds or throws `OverflowException` if the value does not fit in 32 bits. Previously it was always throwing `TypeErrorException` with a rather cryptic message: `__int__ returned non-int (int)`, regardless of the value.
* If `__complex__` returns non-complex, `TypeErrorException` with a proper message is returned `__complex__ returned non-complex (xxx)` rather than `InvalidCastException`.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -2833,13 +2833,12 @@ namespace IronPython.Runtime.Operations {
         }
 
         public static object? ConvertFloatToComplex(object value) {
-            if (value == null) {
-                return null;
-            }
-
-            if (value is double d) return new Complex(d, 0.0);
-            if (value is Extensible<double> ed) return new Complex(ed.Value, 0.0);
-            throw new InvalidOperationException();
+            return value switch {
+                null => null,
+                double d => new Complex(d, 0.0),
+                Extensible<double> ed => new Complex(ed.Value, 0.0),
+                _ => throw new InvalidOperationException(),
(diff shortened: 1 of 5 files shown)


Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -2833,38 +2833,44 @@ public static partial class PythonOps {
...
                 _ => throw new InvalidOperationException(),
             };
         }
 
-        internal static bool CheckingConvertToInt(object value) {
-            return value is int || value is BigInteger || value is Extensible<BigInteger>;
+        public static object? ConvertIntToInt32(object? value) {
+            return value switch {
+                null => null,
+                int i => i,
```
slozier, line 2857: Probably insignificant optimization, but maybe return `value` here? I guess the same applies to `ConvertIntToBigInt`.

BCSharp: I thought about it and took the lead from `ConvertToPythonPrimitive` (a few lines above), which does more of this kind of stuff, presumably for readability (?).

OK to optimize it.

slozier (approved):
Looks good to me!

Issue #1404, opened 2022-04-17
ctypes .from_buffer only works for array.array


orent:
The .from_buffer[_copy] methods of ctypes types only accept an array.array argument.  Other implementations such as cpython, pypy also accept bytes, bytearray, mmap, memoryview etc.

The struct and re module accept a wider variety of buffer-like objects. Strangely, re accepts mmap objects, but stuct does not.

orent:
Correction: from_buffer_copy() does accept bytes, but not bytearray, memoryview, etc.

BCSharp:
Thanks for the submission. This seems to be related to #21, which is only partially finished. What is still needed is a review of all modules and converting their API to use `IBufferProtocol` as needed. `struct` and `ctypes` are the two prominent ones. Especially getting `ctypes` up to speed will be exciting because it is a prerequisite for supporting `numpy`.

slozier:
`from_buffer` is probably blocked by https://github.com/IronLanguages/ironpython3/issues/1297. `from_buffer_copy` should be simple to fix with `IBufferProtocol` (thanks to @BCSharp).

BCSharp:
Yes, `from_buffer_copy` will be simple. I can pick it up since the int/long unification stuff is drying up.
For the `from_buffer`, indeed #1297 needs to be done first.

slozier:
@BCSharp I guess I should have waited a bit! https://github.com/IronLanguages/ironpython3/pull/1405 could use a review.

BCSharp:
No problem! I wouldn't be able to get at it till tonight at best. Will review your PR instead, hopefully this time pushing all the right buttons... 😄

orent:
Thanks for the fast response. Two more minor inconveniences that caused my code not to work out-of-the-box on IronPython are:

1.  struct.unpack_from not accepting an mmap object
2. ctypes.Structure.\_pack\_ attribute accepts only int, not bool

Should I open issues for these?

what is the easiest way to get a build of something newer than 3.4alpha?

slozier:
> Should I open issues for these?

I think `struct.unpack_from` will take a bit of effort to implement properly and might be worth its own issue. I can take a look at `ctypes.Structure._pack_` later today, I think it should be easy.

> what is the easiest way to get a build of something newer than 3.4alpha?

You should be able to get the latest build from the main page. If you search for "builds" there are badges for the Azure and Github CI. They should both publish build artifacts that you can download.

BCSharp:
[...]

PR #1405, opened 2022-04-17
Update from_buffer_copy to accept an IBufferProtocol

slozier:
Part of https://github.com/IronLanguages/ironpython3/issues/1404

### DIFF
--- a/Src/IronPython.Modules/_ctypes/ArrayType.cs
+++ b/Src/IronPython.Modules/_ctypes/ArrayType.cs
@@ -109,35 +109,16 @@ namespace IronPython.Modules {
                 return res;
             }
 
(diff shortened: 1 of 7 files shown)


BCSharp (commented):
The changes are fine except for `WriteSpan` used now in several places. The use is OK but the implementation of method itself  caught my attention. It is much less efficient than what spans allow. Perhaps it can be rewritten to something like:

```C#
internal unsafe void WriteSpan(int offset, ReadOnlySpan<byte> value) {
    Span<byte> dest = new Span<byte>((void*)_data, _size).Slice(offset);
    value.CopyTo(dest);
    GC.KeepAlive(this);
}
```

BCSharp (commented):
Looks good to me!

Issue #1297, opened 2021-09-15
Support pinning of IPythonBuffer


slozier:
Currently the `IPythonBuffer` interface only provides access to the buffer via a Span<byte>. Adding `IPythonBuffer.AsMemory<byte>()` would enable some scenarios such as [`ctypes._CData.from_buffer`](https://docs.python.org/3/library/ctypes.html#ctypes._CData.from_buffer) which requires(?) pinning of the memory for use in unmanaged code.

@BCSharp since you implemented `IPythonBuffer`, any thoughts or concerns?

BCSharp:
I did not have the opportunity to check `ctypes._CData.from_buffer` in depth, but from what I remember about `IPythonBuffer`, there is no inherent reason to use `Span` over `Memory`. It all boils down to a trade-off: using`Span` allows more sources but puts more constraints on the consumers, using `Memory` requires the source to be on the heap but allows more consuming cases.

I started with `Span` and was curious to see if we hit the wall somewhere. It seems that you have found the case where `Span` is not good enough. If so, `Memory` is the way to go; there is no need to keep `Span` around in the API if it gets replaced by `Memory`.

BCSharp:
After dusting off this issue I now remember why `IPythonBuffer` provides access only by span: it allows access to both managed and unmanaged memory. I expect unmanaged buffers being created by `numpy`, `pillow`, and the like.

slozier:
I can rename this issue to something like "Support pinning of IPythonBuffer" or something of the sort if you think `AsMemory` is not the way to go (I think I had run into some annoyances when I initially tried to run with it so I'm fine with going with another solution). Or we can close this issue and make a new one with a better description of the issue. The `array` module already does something of the sort (exposes its address when you call `buffer_info`) so it might offer some insight.

BCSharp:
> I can rename this issue to something like "Support pinning of IPythonBuffer"

Sounds good.

PR #1409, opened 2022-04-18
Convert internal field CData._memHolder to property CData.MemHolder

BCSharp:
Prepares ground for #1297.

It is a simple, not-functional change but because of the rename it touches many places so I submit it as a separate PR to keep subsequent PRs small (hopefully).


slozier (approved):
Looks good to me!

PR #1411, opened 2022-04-19
Support pinning of IPythonBuffer

### LINKED ISSUE #1297: Support pinning of IPythonBuffer

BCSharp:
Resolves #1297 though more can be done. Currently the lifetime of `MemoryHolder` is not managed efficiently when used from C#. It was like that before this change and it does not affect Python code, which relies on the finalizers and the GC, but for C# calls, it can be done more efficiently using `Dispose()`. I think there are more improvements/modernizations/cleanups possible here, and likely in the whole `ctypes` module, but that is for another PR (probably several).

I have locally other changes in `ctypes` that would address the remaining part from #1404 and I used them for testing, not sure should they go in the separate PR (since they address a separate issue) or here (since they validate the code changes).


slozier:
Looks like a pretty clean solution. I like it.

> I have locally other changes in `ctypes` that would address the remaining part from #1404 and I used them for testing, not sure should they go in the separate PR (since they address a separate issue) or here (since they validate the code changes).

I'm fine either way if the changes are small and doesn't make it hard to review this PR.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -930,7 +931,7 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
                 throw PythonOps.BufferError("memoryview: underlying buffer is not writable");
 
             if (flags.HasFlag(BufferFlags.CContiguous) && !_isCContig)
-                throw PythonOps.BufferError("memoryview: underlying buffer is not C-contiguous");
+                throw PythonOps.BufferError("memoryview: underlying buffer is not C contiguous");
```
slozier, line 934: I thought Python used "C-contiguous" with a dash?
```py
struct.unpack("H", memoryview(b"AA")[::-1])
```
> BufferError: memoryview: underlying buffer is not C-contiguous

slozier (commented):
Generally looks good to me (other than the change in memoryview error messages).

Src/IronPython/Runtime/MemoryView.cs
BCSharp: `memoryview` does, but `ctypes` doesn't.
```python
c_short.from_buffer(memoryview(bytearray(b"AA"))[::-1])
```
> TypeError: underlying buffer is not C contiguous

Unfortunately, tests for `ctypes` check for this form. Will see if I can get it both ways.

slozier: Took a peek at the CPython ctypes implementation and it seems like `from_buffer` is the odd one:
```
ctypes.c_int.from_buffer(memoryview(bytearray(b"aaaa"))[::-1])
```
> TypeError: underlying buffer is not C contiguous

but `from_buffer_copy` gives the `memoryview` error:
```
ctypes.c_int.from_buffer_copy(memoryview(bytearray(b"aaaa"))[::-1])
```
> BufferError: memoryview: underlying buffer is not C-contiguous

slozier:
Thanks!

PR #1420, opened 2022-04-24
Introduce NotNoneAttribute alias

### LINKED ISSUE #1412: Confusing use of NotNullAttribute
`NotNullAttribute` comes from two sources:
1. Originally from the DLR (namespace [_Microsoft.Scripting.Runtime_](https://github.com/IronLanguages/dlr/blob/master/Src/Microsoft.Scripting/Runtime/NotNullAttribute.cs)) to mark parameters that do not accept a dynamic null value.
2. Newer use to support nullable annotations (namespace [_System.Diagnostics.CodeAnalysis_](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullattribute?view=net-6.0)).

This creates a name clash when both attributes need to be used in the same file or even when only one of the attributes is used but both namespaces are imported. Various compilation units in the codebase handle the name clash in various ways: sometimes `NotNull` for the code analysis is renamed to `NotNullOnReturn`, and sometimes the DLR `NotNull` is renamed to `NotDynamicNull`. If there is no clash but just ambiguity, the attribute may not be renamed at all, just redeclared.

Even when there is no name clash or ambiguity, this is still confusing: when reading the code, it is not immediately obvious which attribute (hence which meaning) is applied. This has been bugging me already for a while.

The issue [has been discussed before](https://github.com/IronLanguages/ironpython3/pull/1140#discussion_r597998663) resulting in a compromise of using `NotDynamicNull` for the DLR `NotNull`. This requires renaming the attribute in the prologue of every _.cs_ file where such attribute is used (and it is used frequently). My understanding was that, with time, `NotDynamicNull` will be supported directly by the DLR and all those renames polluting _.cs_ files could be removed without requiring further changes to the IronPython code. Because this was supposed to be a DLR attribute, that was my reason for rejecting the alternative name `NotNoneAttribute`.

But now with `global using`, the whole issue can be easily handled exclusively at the IronPython level; no need to drag DLR into this. [As I have stated before](https://github.com/IronLanguages/ironpython3/pull/808#discussion_r426281282), I prefer `NotNone` over `NotDynamicNull`, primarily because is as short as `NotNull` and length matters since this attribute is used in method declarations where most often the whole declaration is put on one line.

Therefore I propose the following solution:
1. `NotNull` **always** means `System.Diagnostics.CodeAnalysis.NotNullAttribute`.
2. `Microsoft.Scripting.Runtime.NotNullAttribute` is renamed **globally** to `NotNone`. The code uses **solely** the renamed version.
3. If possible, _IronPythonAnalyzer_ is used to spot usage of not-renamed DLR `NotNull`.
4. All existing other aliases for those two attributes are abolished, together with the corresponding file-scope `using` declarations for those aliases.
5. The rename of DLR `NotNnull` is done with `global using` so that:
   1. It is readilly available without extra file-scope `using` (this attribute is much more often used than the one for code analysis)
   2. If the DLR ever does change the name for `NotNull`, it will require update in only one place (per project).

If this proposal gets a go ahead, I am prepared to implement the changes throughout the whole codebase.

BCSharp:
Resolves #1412.


slozier:
`NotNullOnReturn` is something we want to keep? Or it's just to ease the PR review process?

BCSharp:
It is an artifact of the way I was making the changes and testing them. I have it already removed locally. In general, I am not a big fan of renames, unless there is a really compelling reason to do so, like `NotNone`.

But to play devil's advocate: one possible reason to use `NotNullOnReturn` would be to be able to do that rename as `global using`, next to `NotNone`. Right now, if a file uses `[NotNull]` (from `System.Diagnostics.CodeAnalysis`) but also uses `Microsoft.Scripting.Runtime` namespace in the same file, there is a name ambiguity. To resolve that, the file still needs to declare 
`using NotNullAttribute = System.Diagnostics.CodeAnalysis.NotNullAttribute` (or spell out the fully-qualified name). But since it only happens in a handful of files, I don't think it is worth to introduce a non-standard alias.

slozier (approved):
Thanks!

PR #1423, opened 2022-04-26
Update notes on upgrading from ipy2 to cover PEP 237

BCSharp:
Concludes #52.

### DIFF
--- a/Documentation/upgrading-from-ipy2.md
+++ b/Documentation/upgrading-from-ipy2.md
@@ -33,7 +33,19 @@ False
 'BigInteger'
 ```
 
-This means that in interop cases, when the `int` type is used (think generics), it will mean `BigInteger` and not `Int32` (which was the case in IronPython 2). For example, `System.Collections.Generic.List[int]` in IronPython 3 is equivalent to `System.Collections.Generic.List[System.Numerics.BigInteger]`. To retain IronPython 2 semantics, replace `int` with `System.Int32`.
+This means that in interop cases, when the `int` type is used (think generics), it will mean `BigInteger` and not `Int32` (which was the case in IronPython 2). To retain IronPython 2 semantics, replace `int` with `System.Int32`.
+
+Example:
+
+```python
+# IronPython 2
+System.Collections.Generic.List[int]
+```
+
+```python
+# IronPython 3
+System.Collections.Generic.List[System.Int32]
+```
 
 Overview of `int` type equivalency:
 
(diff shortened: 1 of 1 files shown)


Documentation/upgrading-from-ipy2.md
```diff
@@ -14,3 +14,167 @@ In an effort to improve compatibility, `sys.platform` no longer returns `cli`. I
 if sys.implementation.name == "ironpython":
     print("IronPython!")
 ```
+
+## `int` Type
+
+One of the major backward incompatible changes in Python 3 is [PEP 0237][]: Essentially, `long` renamed to `int`. That is, there is only one built-in integral type, named `int`; but it behaves mostly like the old `long` type. From the pure Python perspective this means that `int` should be used wherever previously `long` was used. More consideration has to be applied in interop cases with .NET.
```
slozier, line 20:
```suggestion
One of the major backward incompatible changes in Python 3 is [PEP 237 – Unifying Long Integers and Integers][PEP 0237]: Essentially, `long` renamed to `int`. That is, there is only one built-in integral type, named `int`; but it behaves mostly like the old `long` type. From the pure Python perspective this means that `int` should be used wherever previously `long` was used. More consideration has to be applied in interop cases with .NET.
```

Documentation/upgrading-from-ipy2.md
```diff
@@ -14,3 +14,167 @@ In an effort to improve compatibility, `sys.platform` no longer returns `cli`. I
...
+
+As for instances of `int`, mostly for performance reasons, IronPython may use instances of `System.Int32` to hold smaller integers, while `BigInteger` instances are used for large integers. This is done transparently from the Python side, but again the distinction may become relevant for interop cases. Examples:
+
+```python
+i = 1        # instance of Int32
+j = 1 << 31  # instance of BigInteger
+k = j - 1    # still BigInteger, as one of the arguments makes the result type BigInteger
+```
+
+This means that the type of `Int32` objects is always reported as `int` (which is the same as `BigInteger`). If it is important to check what is the actual type of a given integer object, test for the presence of `MaxValue` or `MinValue`. For those properties to be visible, `System` has to be imported first.
```
slozier, line 55: Maybe something like `isinstance(i, System.Int32)` would be a "better" check? The intent of using `hasattr(i, 'MaxValue')` may not be clear and would probably need to accompanied by a comment if used in code.

Documentation/upgrading-from-ipy2.md
```diff
@@ -14,3 +14,167 @@ In an effort to improve compatibility, `sys.platform` no longer returns `cli`. I
...
+
+The creation of either `Int32` or `BigInteger` instances happens automatically by the `int` constructor. If for interop purposes it is important to create a `BigInteger` (despite the value fitting in 32 bits), use method `ToBigInteger`. It converts `Int32` values to `BigInteger` and leaves `BigInteger` values unaffected.
+
+```pycon
+>>> bi = i.ToBigInteger()
+>>> hasattr(bi, 'MaxValue')
+False
+```
+
+In the opposite direction, if it is essential to create `Int32` objects, either use constructors for `int` or `Int32`. The former converts an integer to `Int32` if the value fits in 32 bits, otherwise it leaves it as `BigInteger`. The latter throws an exception is the conversion is not possible.
```
slozier, line 79: I wonder if we should avoid suggesting the `int` constructor in case future selves decide that we want to change the behaviour (to not force to `Int32` when in range). Or at least put some sort of disclaimer (like "in the current implementation").

Documentation/upgrading-from-ipy2.md
```diff
@@ -14,3 +14,167 @@ In an effort to improve compatibility, `sys.platform` no longer returns `cli`. I
...
+Array[UInt32]((0, 0, 1))
+>>> i.GetWords()
+Traceback (most recent call last):
+  File "<stdin>", line 1, in <module>
+AttributeError: 'int' object has no attribute 'GetWords'
+>>> Microsoft.Scripting.Utils.MathUtils.GetWords(i)
+Array[UInt32]((1))
+```
+
+Another set of Python-hidden methods on `long` in IronPython 2 that not available on `int` in IronPython 3 are conversion methods with names like `ToXxx`. The recommended way to perform type conversions like those is to use type constructors. The exception is the conversion to `BigInteger` itself, for the reasons explained above.
```
slozier, line 160: typo: missing "are" in "that not available"

I thought you had added all the `IConvertible` like methods to `int`? Though I guess those all take an `IFormatProvider` as an argument...

slozier (commented):
Generally looks good. Maybe make the change to generics a little bit more prominent with an example:
```python
# IronPython 2
System.Collections.Generic.List[int]
```

```python
# IronPython 3
System.Collections.Generic.List[System.Int32]
```
A lot of people are not necessarily great at reading. :smile:

Documentation/upgrading-from-ipy2.md
BCSharp: This looks indeed much more readable.,

Documentation/upgrading-from-ipy2.md
BCSharp: Being able of "normalizing" an instance in some way is important. From the CLR perspective, `BigInteger(1)` is not the same object as `Int32(1)`. I was thinking of using `__pos__` (operator `+`) for that purpose, but somehow the constructor felt like the most appropriate way. Indeed, it may change in the future, like if `Int64` is added to the list of types we use to accelerate `int`. Or, going in the other direction, if `Int32` is dropped and `int` is just `BigInteger` all the time.

I am fine with the disclaimer "in the current implementation".

Documentation/upgrading-from-ipy2.md
BCSharp: Yes, they are different `ToXxx` methods. Keeping them on `BigInteger` seemed superfluous since the job can be easily done using relevant constructors (which feels more Pythonic) and also if they stayed in `BigInteger` they would have to be added on `int32` too. Looked just like pollution.

Documentation/upgrading-from-ipy2.md
```diff
@@ -52,51 +64,51 @@ j = 1 << 31  # instance of BigInteger
...
 
 ```pycon
 >>> bi = i.ToBigInteger()
->>> hasattr(bi, 'MaxValue')
+>>> isinstance(j, System.Int32)
 False
 ```
 
-In the opposite direction, if it is essential to create `Int32` objects, either use constructors for `int` or `Int32`. The former converts an integer to `Int32` if the value fits in 32 bits, otherwise it leaves it as `BigInteger`. The latter throws an exception is the conversion is not possible.
+In the opposite direction, if it is essential to create `Int32` objects, either use constructors for `int` or `Int32`. In the current implementation, the former converts an integer to `Int32` if the value fits in 32 bits, otherwise it leaves it as `BigInteger`. The latter throws an exception is the conversion is not possible. Although the behavior of the constructor `int` may or may not change in the future, it is always guaranteed to convert the value to the "canonical form" adopted for that version of IronPython.
```
BCSharp, line 91: The declaration about the constructor always producing the canonical form is debatable. Although a documented and reliable way of producing a canonical representation is necessary, it doesn't have to be the constructor. Looking at the example of Pyhon `Decimal`, it uses three operations to produce some form of a normalized value:
* `__pos__()`, `+x` (on instance), `plus(x)` (on context) — "This operation applies the context precision and rounding, so it is not an identity operation."
* `normalize()` — "Used for producing canonical values for attributes of an equivalence class."
* `canonical()` — "Return the canonical encoding of the argument." It also comes with test `is_canonical()`.

According tho the documentation, current implementation of `Decimal` keeps all instances canonical at all times anyway, although it does not explain what that means. To make the matter more confusing:

> Q. There are many ways to express the same value. The numbers 200, 200.000, 2E2, and 02E+4 all have the same value at various precisions. Is there a way to transform them to a single recognizable canonical value?
>
> A. The normalize() method maps all equivalent values to a single representative:

Perhaps `canonical()` refers only to some internal representation. If so, this is exactly what we are looking for in the case of our `int`.

slozier: I think it's fine (in particular since it's quoted). If someone has an issue with the wording they can file an issue! 😄

slozier (approved):

PR #1424, opened 2022-04-27
Enable test_numtypes

BCSharp:
It looks like this test has never been run on IronPython 3 before — it was still using Python 2 syntax. After upgrading, it exposed some gaps in the operators on the non-Python integers.

I am torn between enabling and not enabling this test by default. It was originally disabled because it "takes way too long". Indeed, it takes about 30 seconds on my rig, which I think means it can be some extra 3 minutes on the CI server. In the end I enabled it because I have some resolver-related tinkering planned ahead and would prefer the test to catch as much mistakes as possible.


slozier (commented):
Seems fine. I guess it matches the behaviour we had before (not that I necessarily like what we do).

PR #1433, opened 2022-05-01
Add Install-IronPython.ps1

BCSharp:
This is meant to bridge the install functionality till we have `dotnet tool` supported. I originally tried to have a script that can install ItonPyton/.NET 6 from just the zipfile packaged by the `package` target, but it turned out that the content of it is not sufficient (esp. on Unix). So the script now pulls additional binaries from the `bin` directory. Ugly, as there is no guarantee  thatthey are in sync. Perhaps a better option would have been to create an additional build target, say `install`, that installs a standalone IronPython runtime in a designated location, but at the moment, this script is sufficient for me. Now I can start testing out my favourite Python packages under IronPython 😃

### DIFF
--- a/Src/Scripts/Install-IronPython.ps1
+++ b/Src/Scripts/Install-IronPython.ps1
@@ -8,7 +8,7 @@
     Install IronPython 3 for .NET 6 from an official zip file distributable.
 
 .DESCRIPTION
-    This script facilitates installation of IronPython on .NET 6 binaries from a zip file. The zip file is assumed to have content as published on the IronPython's downolad page. The zip file is produced by IronPython's "package" build target.
+    This script facilitates installation of IronPython on .NET 6 binaries from a zip file. The zip file is assumed to have content as published on the IronPython's download page. The zip file is produced by IronPython's "package" build target.
 
 .EXAMPLE
     PS>./make
@@ -32,7 +32,7 @@ Param(
     [Parameter(Position=0, Mandatory)]
     [string] $Path,
 
-    # The path to the downoladed zip file with IronPython binaries. If empty, the script will try to grab the package directly produced by the local build.
+    # The path to the downloaded zip file with IronPython binaries. If empty, the script will try to grab the package directly produced by the local build.
     [string] $ZipFile,
 
     # If the target path exists, it will be wiped clean beforehand.
@@ -42,11 +42,11 @@ Param(
 $ErrorActionPreference = "Stop"
 
 $defaultVersion = "3.4.0-beta1"
-$projectRoot = $PSScriptRoot | Split-Path | Split-Path
 
 if (-not $ZipFile) {
     # If zipfile path not given, assume that the script is in the standard location within the source tree
     # and locate the zip archive in the standard location of the package target.
+    $projectRoot = $PSScriptRoot | Split-Path | Split-Path
     $ZipFile = Join-Path $projectRoot "Package/Release/Packages/IronPython-$defaultVersion/IronPython.$defaultVersion.zip"
 }
 
@@ -61,13 +61,18 @@ if (Test-Path $Path) {
     }
 }
 
+# Unzip archive and move files into place
 $unzipDir = Join-Path $Path "zip"
 
 Expand-Archive -Path $ZipFile -DestinationPath $unzipDir
 Move-Item -Path (Join-Path $unzipDir "Lib") -Destination $Path
 Move-Item -Path (Join-Path $unzipDir "net6.0/*") -Destination $Path -Exclude "*.xml","*.dll.config"
-Remove-Item -Recurse -Path $unzipDir
+Remove-Item -Path $unzipDir -Recurse
 
+# Remove garbage files
+Remove-Item -Path (Join-Path $Path "Lib/__pycache__") -Recurse
+
+# Create a startup script
 $ipyPath = Join-Path $Path "ipy.ps1"
 Set-Content -Path $ipyPath -Value @'
 #!/usr/bin/env pwsh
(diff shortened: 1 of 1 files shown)


Src/Scripts/Install-IronPython.ps1
```diff
@@ -0,0 +1,104 @@
...
+# Unzip archive and move files into place
+$unzipDir = Join-Path $Path "zip"
+
+Expand-Archive -Path $ZipFile -DestinationPath $unzipDir
+Move-Item -Path (Join-Path $unzipDir "Lib") -Destination $Path
+Move-Item -Path (Join-Path $unzipDir "net6.0/*") -Destination $Path -Exclude "*.xml","*.dll.config"
+Remove-Item -Path $unzipDir -Recurse
+
+# Remove garbage files
+Remove-Item -Path (Join-Path $Path "Lib/__pycache__") -Recurse -ErrorAction Ignore
```
slozier, line 73: Hmm, you have `__pycache__` files making their way to the package? Probably an issue that needs to be fixed.

Src/Scripts/Install-IronPython.ps1
```diff
@@ -0,0 +1,104 @@
...
+if ($IsMacOS -or $IsLinux) {
+    chmod +x $ipyPath
+}
+
+# Add items that are missing in the zip file, directly from the bin directory
+if ($projectRoot) {
+    $binPath = Join-Path $projectRoot "bin/Release/net6.0"
+    Copy-Item (Join-Path $binPath "ipy") $Path
+    if ($IsMacOS -or $IsLinux) {
+        Copy-Item (Join-Path $binPath "Newtonsoft.Json.dll") $Path
```
slozier, line 90: Are you using `Newtonsoft.Json` for something?

Src/Scripts/Install-IronPython.ps1
```diff
@@ -0,0 +1,104 @@
...
+        Write-Error "Path already exists: $Path"
+    }
+}
+
+# Unzip archive and move files into place
+$unzipDir = Join-Path $Path "zip"
+
+Expand-Archive -Path $ZipFile -DestinationPath $unzipDir
+Move-Item -Path (Join-Path $unzipDir "Lib") -Destination $Path
+Move-Item -Path (Join-Path $unzipDir "net6.0/*") -Destination $Path -Exclude "*.xml","*.dll.config"
```
slozier, line 69: Guess we could prune the `.dll.config` from the net6.0 - I think it's only needed for .NET Framework...

slozier (commented):
Seems fine to me if it's helpful to you. Any value in putting it in the root instead of `Src/Scripts`?

Src/Scripts/Install-IronPython.ps1
BCSharp: Right now it is happening only in my macOS working directory, but if I check it out afresh, `__pycache__` is not being included. A similar thing was happening in my Windows workarea on the day of the release, when a fresh checkout helped. It seems that with use, something is happening to the workarea that `make clean` cannot clean. Not a problem for CI builds, but since I intend to use this script from my regular workarea, I want to automatically skip over this stuff without requiring a fresh checkout.

Src/Scripts/Install-IronPython.ps1
BCSharp: No, but I wasn't sure if it wasn't being dynamically loaded by `ipy`, like `Mono.Unix.dll`. If you know for sure it's not needed, I'd rather not include it. But if it is not used, why is it being put in `bin` for normal builds? Used in testing?

Src/Scripts/Install-IronPython.ps1
BCSharp: > I think it's only needed for .NET Framework...

That is my understanding too. What is more, I think that `.dll.config` is only used during build, to be assembled into `.app.config`, and only `.app.config` has any effect on a running .NET Fmwk app.

BCSharp:
> Any value in putting it in the root instead of `Src/Scripts`?

I am very reluctant to put anything in the project root, user home etc.  unless absolutely necessary. This is the one thing that can't be collapsed when viewing the source tree. The project root is already packed, pushing the `README.md` in GitHub off screen. From the usability point of view, I simply create an alias `isipy` for this script right in my PS `$profile` and have it under my fingertips. From the discoverability point of view: don't get too attached to it, it is meant as a temporary band-aid, until we have `IronPython.Tool` on NuGet.

Src/Scripts/Install-IronPython.ps1
slozier: Ok, managed to reproduce the failure. I think there might be a few issues:
- Exclusion glob pattern for `__pycache__` in `StdLib.pyproj` is incorrect.
- Staging folder is not getting cleared. This might cause other junk to make its way into the package...
- `msi\obj` is not getting cleaned? Somehow I managed to hit the same issue you had and zapping this folder "fixed" the issue.

Src/Scripts/Install-IronPython.ps1
slozier: Looks like a transitive dependency for `Microsoft.NET.Test.Sdk` so not needed. I wonder if there's a way to not have it end up in the `bin`...

Src/Scripts/Install-IronPython.ps1
BCSharp: Does it mean you are going to address these issues? Or shall I? Or we park it for now and leave the script as is? I see you are making great progress towards 3.6 and I don't want to distract you with my stuff.

slozier: Glob pattern is a quick fix so I was going to do that one. Didn't have any plans for the other two.

BCSharp: > * `msi\obj` is not getting cleaned?

I have noticed before that none of the 17 `obj` dirs are ever being cleaned, which on some occasion caused me problems. They usually contain both `Release` and `Debug` subdirs with often some files directly in `obj`. Thinking about a fix, it is not clear which target should clean them. Maybe a new target `clean-all`?

slozier: I guess part of the question is how much cleaning is expected of "clean". Like if you create a brand new .NET console program then run `dotnet build` and `dotnet clean` there's still a lot of stuff left over in the `obj` folder. Anyway, if we want to force clean everything maybe it could be called `purge`?

BCSharp: I am going to file a separate issue for this.

PR #1437, opened 2022-05-03
Add Enter-IronPythonEnvironment.ps1

BCSharp:
An IronPython tailored script to activate a given IronPython installation. Currently can handle the MSI installations and the ones created by `Install-IronPython.ps1` (#1433), but in the future also installations createdby `dotnet tool`. So unlike `Install-IronPython.ps1`, this script is meant to stay, and possibly be part of the installation itself.

This script is something between Python's _venv_ and Anaconda. It is more flexible than the _venv_'s `Activate.ps1` since it is not tied to a specific installation, with paths hard-coded in the scripts. But is more light-weight than Aanaconda's `Enter-CondaEnvironment.ps1` because it is self-contained and does not have to be loaded in PowerShell's `$profile`.

I personally set up an alias `etipy` in `$profile` pointing to this script in a well-known and stable location. and use it for all environments I have.

In some respects it can do more than either _venv_ or Anaconda, esp. saving/restoring `Env:IRONPYTHONPATH`, which I find quite useful given my way of working with many different environments, paths and aliases to `ipy` commands. I hope it will be useful to others as well.

### DIFF
--- a/Src/Scripts/Enter-IronPythonEnvironment.ps1
+++ b/Src/Scripts/Enter-IronPythonEnvironment.ps1
@@ -1,0 +1,139 @@
+#!/usr/bin/env pwsh
+# Licensed to the .NET Foundation under one or more agreements.
+# The .NET Foundation licenses this file to you under the Apache 2.0 License.
+# See the LICENSE file in the project root for more information.
+
+<#
+.SYNOPSIS
+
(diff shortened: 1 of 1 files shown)


BCSharp:
Reverted the lowercase "lib" change because it is a more complex issue, deserving its own Issue report and discussion.

slozier (approved):
Looks good to me if it works for you.

Issue #1439, opened 2022-05-04
pyserial throws exception when creating serial port


maBarabas:
### Description

A strange exception is produced when trying to use pyserial on windows. The same version of pyserial works correctly on the official python implementation.

Is this because of pyserial using ctypes? I see that numpy is not supported because it is implemented in C.

```
(.venv) C:\Users\barabas\snip>ipy
IronPython 3.4.0b1 (3.4.0.0010)
[.NETFramework,Version=v4.6 on .NET Framework 4.8.4470.0 (64-bit)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import serial
>>> serial.Serial("COM4")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\barabas\snip\.venv\lib\site-packages\serial\serialwin32.py", line 33, in __init__
  File "C:\Users\barabas\snip\.venv\lib\site-packages\serial\serialutil.py", line 244, in __init__
TypeError: expected int or long, got bool
```

The lines are:

```
    def __init__(self, *args, **kwargs):
        self._port_handle = None
        self._overlapped_read = None
        self._overlapped_write = None
        super(Serial, self).__init__(*args, **kwargs) # here
```

and

```
        if port is not None:
            self.open() # here        
```

### Steps to Reproduce

1. Install pyserial==3.5, IronPython 3.4.0b1 on Windows
2. Connect a USB to serial adapter
3. Start ipy and run the commands from above

**Expected behavior:** Serial port is created

**Actual behavior:** Exception thrown

### Versions

You can get this information from executing `ipy -V`.

    IronPython 3.4.0b1

pip freeze:

    pyserial==3.5
    robotframework==4.1.3
    typing==3.10.0.0

slozier:
Thanks for the report. I think this occurs whenever a boolean is assigned to ctypes field with a bit count. Managed to hit the exception with code like this:
```py
import ctypes

class Test(ctypes.Structure):
    _fields_ = [("x", ctypes.c_bool, 1)]

assert Test(True).x is True
```

BCSharp:
This seems to be easy to fix. @slozier are you working on it or shall I pick it up?

slozier:
@BCSharp go for it

BCSharp:
#1441 fixed the issue as reported by @slozier but I can't verify it solves the original issue as I don't have a serial port or a serial adapter. @maBarabas, can you confirm it is working now?

[...]

PR #1441, opened 2022-05-07
Improve handling of bitfields in ctypes

BCSharp:
Addressed #1439 and a few other issues that caught my attention. The most interesting is a fix for python/cpython#90914, which was reported just 3 months ago and is still not resolved. Funny thought that IronPython on one point will be ahead of CPython. Not that this functionality is used anywhere in the wild, since it is not working yet in CPython.

### DIFF
--- a/Src/IronPython.Modules/_ctypes/Field.cs
+++ b/Src/IronPython.Modules/_ctypes/Field.cs
@@ -68,13 +68,13 @@ namespace IronPython.Modules {
             internal override bool TryGetValue(CodeContext context, object instance, PythonType owner, out object value) {
                 if (instance != null) {
                     CData inst = (CData)instance;
-                    value = _fieldType.GetValue(inst.MemHolder, inst, _offset, false);
                     if (_bits == -1) {
+                        value = _fieldType.GetValue(inst.MemHolder, inst, _offset, false);
+                        return true;
+                    } else {
+                        value = ExtractBits(inst);
                         return true;
                     }
-
-                    value = ExtractBits(value);
-                    return true;
                 }
 
                 value = this;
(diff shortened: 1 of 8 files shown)


Src/IronPython.Modules/_ctypes/Field.cs
```diff
@@ -42,51 +42,39 @@ public sealed class Field : PythonTypeDataSlot, ICodeFormattable {
...
                 _fieldName = fieldName;
 
-                // if the number of bits is the full type width, we don't
-                // need to do any masking anywhere, we may want to revisit this
-                // if the __repr__ is an issue
-                if (bits != null && bits.Value != (_fieldType.Size * 8)) {
+                if (bits != null) {
                     _bits = bits.Value;
                     _bitsOffset = bitOffset.Value;
+                    if (((SimpleType)_fieldType)._swap) {
```
slozier, line 48: I guess it's guaranteed to be a `SimpleType` if `bits is not null`?

Src/IronPython.Modules/_ctypes/Field.cs
```diff
@@ -112,110 +100,138 @@ public sealed class Field : PythonTypeDataSlot, ICodeFormattable {
...
+                            }
+                        } else {
+                            if (biVal >= long.MinValue) {
+                                newBits = (ulong)(long)biVal;
+                            } else {
+                                newBits = (ulong)(biVal & ulong.MaxValue);
+                            }
+                        }
+                    } else {
+                        throw PythonOps.TypeErrorForTypeMismatch("int", value);
```
slozier, line 219: I think this bit is unreachable so you can throw InvalidOperationException?

Src/IronPython.Modules/_ctypes/Field.cs
```diff
@@ -112,110 +100,138 @@ public sealed class Field : PythonTypeDataSlot, ICodeFormattable {
...
+                            // need to sign extend if high bit is set
+                            if ((bits & (1UL << (_bits - 1))) != 0) {
+                                bits |= ulong.MaxValue ^ validBits;
+                            }
+                            value = (BigInteger)(long)bits;
+                        } else {
+                            value = (BigInteger)bits;
+                        }
+
+                    } else if (value is bool bVal) {
```
slozier, line 166: Is this reachable or `bool` will be caught by the `IsBoolType` check above?

Src/IronPython.Modules/_ctypes/Structure.cs
```diff
@@ -50,6 +50,10 @@ public abstract class _Structure : CData {
                 INativeType nativeType = NativeType;
                 StructType st = (StructType)nativeType;
 
+                if (args is null) { // None passed as a sole argument
```
slozier, line 53: Interesting, `None` gives `null` here? Would adding the `NotNoneAttribute` solve this?

slozier (commented):
Looks good to me! Good thing you took this on since I almost certainly wouldn't have been as thorough! 😄

Src/IronPython.Modules/_ctypes/Field.cs
BCSharp: Yes, and the code depended on this constraint already before but it was not that visible (see `IsSignedType` and its use). I'm gonna drop in an explicit debug assert here.

Src/IronPython.Modules/_ctypes/Field.cs
BCSharp: Right. It's a left-over from a pre-3.8 code version.

Src/IronPython.Modules/_ctypes/Field.cs
BCSharp: No, not reachable. Again a leftover from a previous version.

Src/IronPython.Modules/_ctypes/Structure.cs
BCSharp: Yes, `None` gives `null`. This is the consequence that the DLR allows to bind to the `param object[]` array directly (just like C#). We have discussed it before (IronLanguages/dlr#249). Although then I was arguing against it, I now think that for params arrays it will be better to keep the behaviour as it is now.

But indeed, `[NotNone]` prevents it from happening, pushing `null` into the params array. Good insight! It is an elegant solution.

BCSharp:
Thanks for the review!

slozier (approved):

PR #1443, opened 2022-05-08
Remove "unreachable code" condition in target 'package'

BCSharp:


Build/Tasks.Targets
```diff
@@ -35,10 +35,10 @@
                 fileStream.CopyTo(fileStreamInZip);
             }
           }
-          return true;
+          Success = true;
```
slozier, line 38: Is `Success` some sort of magic variable?

BCSharp: It looks like it is an output parameter that is implicitly created on every inline task by the boilerplate code generated by MSBuild. It is `true` by default so strictly speaking it is only needed to be set to `false` in case of a failure, but I think if the task can fail, then for readability it is good to explicitly indicate when the tasks succeeds. Also there is a guideline that a task should set all output parameters in its `Execute` method, although this may not apply to implicit parameters. The part of the boilerplate code that comes after the snippet ends with `return Success;` so if the snippet does `return` on all code paths it triggers the "unreachable code" warning.

PR #1461, opened 2022-05-17
Support str subclass from repr

BCSharp:
I was hoping this would go away in some new version of CPython, but it's still here in 3.10.


slozier (approved):
Looks good to me!

PR #1463, opened 2022-05-18
Implement structural equality test of memoryview

BCSharp:


BCSharp:
Oh, I forgot about nans. Good catch. Well, let's file down this edge.

Src/IronPython/Runtime/MemoryView.cs
```diff
@@ -851,20 +852,70 @@ public sealed class MemoryView : ICodeFormattable, IWeakReferenceable, IBufferPr
...
+                return true;
+            }
+
+            // compare item by item
+            TypecodeOps.DecomposeTypecode(value._format, out char theirByteorder, out char theirTypecode);
+
+            if ((ourTypecode is 'c') && TypecodeOps.IsByteCode(theirTypecode) ||
+                (theirTypecode is 'c') && TypecodeOps.IsByteCode(ourTypecode)) {
+                if (context.LanguageContext.PythonOptions.BytesWarning != Microsoft.Scripting.Severity.Ignore) {
+                    PythonOps.Warn(context, PythonExceptions.BytesWarning, "Comparison between bytes and int");
```
slozier, line 905: I think this warning may be unnecessary? It's only since 3.5 and should be triggered in the 3.6 branch by:
https://github.com/IronLanguages/ironpython3/blob/26767ee5e7de1b361acf75acca55715930476640/Src/IronPython/Runtime/Bytes.cs#L1189-L1194

slozier: and I guess this also triggers incorrectly (at least compared to CPython) on an empty buffer...
```py
memoryview(b'') == memoryview(b'').cast('c')
```

BCSharp: Oh, you have it already in 3.6. How about I revert the last commit here and save the test patch to apply once 3.6 is merged into master? Or I leave the test in since it is only run with the `-b` option which is not enabled by default in standard test runs?

BCSharp: On the other hand, the warning from 3.6 will not be triggered when hitting one of the fast tracks in `memoryview.__eq__`. (To be honest, neither my latest update would do it but this can easily be fixed).

slozier: > Oh, you have it already in 3.6. How about I revert the last commit here and save the test patch to apply once 3.6 is merged into master? Or I leave the test in since it is only run with the `-b` option which is not enabled by default in standard test runs?

I guess you could leave the test in and remove the `is_cli` bit, will automatically work in 3.6 since the version check will pass.

> On the other hand, the warning from 3.6 will not be triggered when hitting one of the fast tracks in `memoryview.__eq__`. (To be honest, neither my latest update would do it but this can easily be fixed).

Doesn't the fast track only compare same formats so there should never be a warning?

BCSharp: > Doesn't the fast track only compare same formats so there should never be a warning?

You are right, my mistake.

slozier (approved):

PR #1464, opened 2022-05-21
Use www.python.org for testing _ssl

BCSharp:
This replaces www.microsoft.com with www.python.org as the site used for the SSl tests. It will hopefully stop the intermittent `test__ssl` failures during CI. 

The failures were in `test_SSLType_read_and_write` during an attempt to read. At first I thought this might be a bug in the implementation of `_ssl` but it turned out that CPython exhibits the same behaviour (i.e. hangs indefinitely). My best guess now is that www.microsoft.com deploys some sophisticated anti-bot defences that hinder the automatic test. www.python.org seems to respond just fine, even during a repeated test run.


slozier (approved):
Awesome! Hopefully that resolves the SSL failures.

PR #1465, opened 2022-05-21
Use MemberNotNullAttribute in PythonEncoding

BCSharp:
I'm cleaning out my stash of old unfinished work of low importance that should not interfere with changes on 3.6.


slozier (approved):
LGTM!

PR #1467, opened 2022-05-22
Improve exception messages when Windows fails to load an extension module

BCSharp:
Probably I would not have noticed it otherwise but it is explicitly listed in "Whats New In Python 3.0". I have traced the change in CPython regarding this announcement to just two commits: [d6179e12](https://github.com/python/cpython/commit/d6179e120d3ed98ec28cb5e4b6feabed2d670235) (code) and [bccd63c3](https://github.com/python/cpython/commit/bccd63c389d72d22882264b8e9cf6e8e5911106c) (NEWS entry).

The update in this PR reflects the change in IronPython. The old error message in IronPython was generic but contained the file name. The CPython message is specific about the cause of the failure but contains _%1_ instead of the file name. I remember encountering both in the past and being puzzled in both cases. So this update combines the specific error code/message with the file name (library path) used.


Src/IronPython.Modules/_ctypes/_ctypes.cs
```diff
@@ -172,6 +174,15 @@ public static partial class CTypes {
             return new Win32Exception(errorCode).Message;
         }
 
+        private static string FormatError(int errorCode, string? fileName) {
+            string msg = FormatError(errorCode);
+            // error codes: https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes
+            if (errorCode is not (< 0 or >= 8200 or 34 or 106 or 317 or 718)) {
```
slozier, line 180: How'd you pick these error codes for formatting?

slozier (approved):
Looks good to me.

Seems like we've got a bunch of versions of the win32 exception handling code. Might be interesting to have something available in the core at some point (like `PythonNT.GetLastWin32Error`).

Src/IronPython.Modules/_ctypes/_ctypes.cs
BCSharp: The link in the comment shows all defined error codes and their messages, grouped by ranges. I searched each range for % to see what variable fields are being used. Luckily, % placeholders are not often used. If a filename is used, it is always on %1 with the four exceptions listed. Above code 8200, filename is not used at all and %1 means something else.

BCSharp:
> Seems like we've got a bunch of versions of the win32 exception handling code. Might be interesting to have something available in the core at some point (like `PythonNT.GetLastWin32Error`).

I saw that too. `ctypes` deserves a thorough review and possibly localized redesign. This is not going to happen for me before the summer recess, I'm now just picking up some loose ends that I've noticed.

PR #1470, opened 2022-05-23
Consolidate winerror message handling in modules _ctypes and nt

BCSharp:
Follow-up on #1467. As it turns out, I had some time.


slozier (approved):
Looks good to me. Didn't know about the `new Win32Exception(errorCode).Message` simplification!

PR #1480, opened 2022-05-28
Accept bytes as filename arg to compile

BCSharp:
CPython also accepts any buffer protocol aware object but with a deprecation warning starting from 3.7, so I think it is more an implementation artifact slated for removal than a legitimate use case. I also haven't met such usage in real life (unlike bytes, which are being passed).


slozier (approved):
Looks good to me!

Issue #1483, opened 2022-05-29
IronPython.SQLite not working on macOS/Linux


BCSharp:
While testing some PyPI packages on a recent ipy 3.4 I have discovered that the `sqlite3` package bundled with IronPython is not working correctly. After looking into the source code and some debugging I've discovered two major issues ans a few smaller ones.

Major issue #1 is that there is only one "platform compatibility" layer: `os_win_c.cs`, which, according to the embedded docs is Windows-specific. Nevertheless, it is being unconditionally compiled on all platforms. The compilation succeeds, but the code compatibility alternatives are limited to Windows 95, Windows CE, Windows RT, Windows Phone, Silverlight, and Windows NT (maybe I have missed some other Windows variant...). When run on macOS/Linux, the compatibility methods run out of all platform detection checks and return an error. I can't see how it ever was working on macOS/Linus in any serious way. When I hacked the code to treat macOS as WinNT, it went a little further, but still hit another snag.

Major issue #2: the code uses .NET API that is not supported on macOS. Warnings about that have been recently disabled (#1330). This pertains file locking, which is essential functionality, effectively making `sqlite3` unable to open any file. An in-memory database should work though. 

The warning itself mentions only problems on macOS, so presumably Linux should be OK (after some hacks). I haven't tested it on Linux yet, but if I find no more snags, I will submit the changes to get at least this platform supported. There is also a possibility to use _Mono.Unix_ compatibility layer on macOS to get partial file locking, but I am not sure if this is fully implemented; the Mono documentation is lacking.

On a wider note, the SQLite implementation that IronPython uses looks very old, unmaintained, supporting obsolete platforms, and not supporting the platforms that IronPython and .NET currently target. So I wonder whether instead of spending time trying to get this code to work, it wouldn't be better to replace it with a modern implementation that is actively maintained (like [SQLite.raw](https://github.com/ericsink/SQLitePCL.raw)).

slozier:
> On a wider note, the SQLite implementation that IronPython uses looks very old, unmaintained, supporting obsolete platforms, and not supporting the platforms that IronPython and .NET currently target. So I wonder whether instead of spending time trying to get this code to work, it wouldn't be better to replace it with a modern implementation that is actively maintained (like [SQLite.raw](https://github.com/ericsink/SQLitePCL.raw)).

[...]

PR #1486, opened 2022-05-30
Support sqlite3 on Linux

BCSharp:
Part of #1483.

### DIFF
--- a/Src/IronPython.SQLite/Cursor.cs
+++ b/Src/IronPython.SQLite/Cursor.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Numerics;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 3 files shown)


BCSharp:
I tried a little to get it work on macOS but it was too much effort. If I had the time to work more on `sqlite3`, I'd rather replace the backend than spend time hacking the current code that is already heavily hacked.

slozier (approved):
Little bit odd to have `isNT` returning `true` on Unix, but the code is already a mess, so looks good to me.

BCSharp:
> Little bit odd to have `isNT` returning `true` on Unix, but the code is already a mess, so looks good to me.

I agree and actually had a commit changing it to `isNTorPosix` but it did not look cleaner: a lot of code calling `isNT` is commented out so either the rename would leave it using a function that is no longer there or if comments were renamed too then the code in those block would really look weird as it is doing sometimes some pretty non-Posix stuff. So in the end, I opted for a minimal change hence no rename.

PR #1490, opened 2022-06-03
Implement module resource

BCSharp:
Module `resource` is Posix-only. The implementation here is fairly complete as far as I know, except for a small incompatibility when attempting to increase a maximum limit while not being a superuser. CPython raises `ValueError` while this implementation raises `OSError`. I think CPython's choice is a mistake. A `ValueError` is supposed to indicate an invalid value, something that the caller should be able to prevent from happening with proper value checks and preconditions. For instance, trying to set a a pair of limits where _current_ is higher than _maximum_ is a legitimate `ValueError`. But to decide whether the caller is or is not allowed to raise the maximum is the domain of the OS. Some lightweight OS implementations may allow it in some circumstances or for some users. Therefore `OSError` is more appropriate, especially since the call already can raise `OSError` when the systems says "ni" for some reason.

But this is not the main reason I haven't implemented the CPython's behaviour in such case. In general, I prefer to implement all CPython's quirkiness, as unreasonable as it may be — for the sake of compatibility. In this case, however, I think it is not worth the effort, performance, and extra complexity. First, such situations should be rare. Second, to actually properly raise `ValueError`, the implementation would need to add another system call (at least); all this just to report the expected error type for a rare case. Finally, when I looked up how real life modules on PyPI handle such calls, in all cases I found they were catching both `ValueError` and `OSError` together, which is not surprising, given that the caller cannot predict which one might happen. (This by itself is another indication that the CPython's choice is unfortunate). So unless some real-life problem gets reported, I leave it as it is.

Still to do: make working on macOS.

### DIFF
--- a/Src/IronPython.Modules/resource.cs
+++ b/Src/IronPython.Modules/resource.cs
@@ -3,20 +3,16 @@
 // See the LICENSE file in the project root for more information.
 
 // TODO: use LightThrowing
-// TODO: use RuntimeInformation.IsOSPlatform(OSPlatform.OSX)
 // TODO: Port to maxOS
 
 #nullable enable
 
 using System;
-using System.Collections.Generic;
 using System.Numerics;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
-using static System.Environment; // TODO: remove
 
-using IronPython;
 using IronPython.Runtime;
 using IronPython.Runtime.Exceptions;
 using IronPython.Runtime.Operations;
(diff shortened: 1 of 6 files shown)


Tests/modules/system_related/test_resource.py
```diff
@@ -0,0 +1,155 @@
+# Licensed to the .NET Foundation under one or more agreements.
+# The .NET Foundation licenses this file to you under the Apache 2.0 License.
+# See the LICENSE file in the project root for more information.
+
+# tests fro module 'resource' (Posix only)
```
BCSharp, line 5: Typo 'fro' => 'for'

Tests/modules/system_related/test_resource.py
```diff
@@ -0,0 +1,155 @@
...
+    except ImportError:
+        pass
+    else:
+        raise AssertionError("There should be no module resource on Windows")
+
+class ResourceTest(IronPythonTestCase):
+    def setUp(self):
+        self.RLIM_NLIMITS = 16 if is_linux else 9
+
+    @unittest.skipUnless(is_posix, "Posix-specific test")
```
slozier, line 25: I think you can put the skipUnless on the class.

Src/IronPython.Modules/resource.cs
```diff
@@ -0,0 +1,427 @@
...
+[assembly: PythonModule("resource", typeof(IronPython.Modules.PythonResourceModule), PlatformsAttribute.PlatformFamily.Unix)]
+namespace IronPython.Modules;
+
+
+[SupportedOSPlatform("linux")]
+[SupportedOSPlatform("macos")]
+public static class PythonResourceModule {
+    public const string __doc__ = "Provides basic mechanisms for measuring and controlling system resources utilized by a program. Unix only.";
+
+    [Obsolete("Deprecated in favor of OSError.")]
```
slozier, line 29: Does having the Obsolete tag produce any side effects in Python?

Src/IronPython.Modules/resource.cs
```diff
@@ -0,0 +1,427 @@
...
+    public static int RUSAGE_SELF => 0;
+
+    public static int RUSAGE_CHILDREN => -1;
+
+    [PythonHidden(PlatformID.MacOSX)]
+    public static int RUSAGE_THREAD => 1;
+
+    #endregion
+
+    [LightThrowing]
```
slozier, line 112: Any reason why chose to go `LightThrowing` throughout? Is it common to hit a throwing codepath?

slozier (approved):
I have any idea what this module does so if you're happy with it then so am I! :smile:

Src/IronPython.Modules/resource.cs
BCSharp: I haven't noticed any and I expect not. In all cases I saw `resource` used, the code was catching `OSError` rather than `resource.error`. It is here just for compatibility, but Python docs call it deprecated. There is a subtle difference between CPython and IronPython here, e.g.: in CPython, `resource.error(2)` creates `OSError`, but in IronPython it creates `FileNotFoundError`. The same applies to `OSError(2)` by the way. Not a big deal, I think.

Src/IronPython.Modules/resource.cs
BCSharp: The code I saw was trying to increase the soft limit of number of open files, and ignoring the error if it failed. I suspect in many cases increasing limits is an optional optimization, so the try-ignore pattern may be common.

I was debating with myself a little whether I should only raise `OSError` as a light exception, or also `ValueError`. While there is a legitimate use case for light-ignoring an `OSError here, a `ValueError` (usually) means that the caller has made an erroneous call and perhaps the call should fail loud and clear. But I don't expect this module to be used a lot from the C# side; I wrote it primarily because other packages depend on it (directly or indirectly). And given that the functions can already return a light exception and the confusing usage of `ValueError` by `resource` in CPython, I decided to return a light exception in all cases, to keep the door open for full CPython compatibility, if it ever becomes a need. But I could definitely reconsider this choice.

Src/IronPython.Modules/resource.cs
```diff
@@ -0,0 +1,427 @@
...
+        public ulong rlim_max;
+    }
+
+#pragma warning disable CS0649 // Field is never assigned to, and will always have its default value
+    internal struct rusage {
+        // /usr/include/x86_64-linux-gnu/bits/types/struct_rusage.h
+        // /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/resource.h
+
+        public Mono.Unix.Native.Timeval ru_utime;  // user CPU time used
+        public Mono.Unix.Native.Timeval ru_stime;  // system CPU time used
```
BCSharp, line 356: I am not happy with using a type from _Mono.Unix_, here; well, I am not happy with _Mono.Unix_ in general: it does not work correctly on macOS. Although there are no issues with `Mono.Unix.Native.Timeval`, I think that by avoiding using _Mono.Unix_ altogether, the dependency could be removed at some point (at least on macOS, it seems to work OK on Linux).

Src/IronPython.Modules/resource.cs
slozier: I meant side effects as in something like a `DeprecationWarning`. Not finding a `Obsolete` attribute on a property to, but `clr.GetDynamicType` appears to raise the warning. Not a big deal I guess...

> in IronPython it creates `FileNotFoundError`

Ah interesting. Seems like a bug - only the 2 argument OSError constructor returns a subclass (e.g. `OSError(2, '')` becomes `FileNotFoundError` whereas `OSError(2)` should be `OSError`).

Src/IronPython.Modules/resource.cs
BCSharp: Since _Mono.Unix_ is proven not to work correctly in all cases, assuming that any call through this library will work as expected without any testing is asking for trouble. I myself got burned on this assumption a few times. So in practical terms, _Mono.Unix_ must be considered untested (on macOS), and we need good test coverage regardless whether the call goes through the library or directly through P/Invoke. The advantage of P/Invoke is that it is readable/debuggable; I can check every signature and types used for correctness. On top of that, quite some of the missing functionality is not supported by _Mono.Unix_ anyway (example: `resource`), and what is supported is usually also (and better) accessible through the regular .NET API. Further, in some cases that Mono does offer some additional and needed functionality and it happens to work, it tweaks the interface slightly to make it a little bit more convenient to use from C#. But this is not the best for IronPython: CPython's modules expose the C lib API at its raw low level, and accessing C lib directly though P/Invoke guarantees that the implementation does what CPython would do, no need to reverse-engineer Mono API. All in all, I think there will be only a small number of cases where _Mono.Unix_ will be practical. This "small number" is not a scientific number, more my gut feeling, we'll see how it is after implementing all the missing modules. But everything is pointing in the direction that there will be "a bunch  of platform specific P/Invokes" which will have to be covered by tests.

BCSharp: > Although there are no issues with `Mono.Unix.Native.Timeval`

I have spoken too soon. On Darwin, `timeval` is `struct { long, int }`. `Mono.Unix.Native.Timeval` is `struct { long, long }` (Linux layout).

BCSharp:
OK, I did some reading of the source of _Mono.Unix_ and that explained a lot. It turns out I was using the library incorrectly. Mono types, flags, error codes defined in _Mono.Unix.Native_ are not (necessarily) the same as the corresponding native symbols of the underlying OS. Mono defines its own API that is the superset of all OS APIs, and maps them with a series of `#ifdef`s to the OS values/types. This is convenient for Mono applications, as they can simply use Mono types and constants with Mono calls and forget about the OS. This is less convenient for IronPython, because values passed from/to the Python user code are expected to be OS compliant. So to make a call to a Mono function, a user-supplied value has to be converted to the Mono language before the call, then Mono converts it back to the OS language and makes a P/Invoke. Similarly, in the opposite direction. The conversion functions in both directions are available in class `Mono.Unix.Native.NativeConvert`.

So I take back that Mono is badly tested on macOS. It is just that the conversions are (sometimes) missing. Mono "is working OK" on Linux because most constants and types are based on Linux. But since macOS uses sometimes different values and types, things were going wrong. With careful conversions, macOS should work fine too. I will probably review all places that IronPython makes a Mono call to check if correct conversions are in place.

Another consequence of the above is that Mono types and constants must not be used with direct P/Invoke calls. Since `resource` has to use P/Invoke, it was incorrect to use `Timeval` from Mono. The last committed version is correct.

PR #1494, opened 2022-06-15
Fix a potential variable misuse bug

LostBenjamin:
Hi,

This pull request is a fix to a potential variable misuse bug at `Src/StdLib/Lib/test/test_functools.py`. Please check the changes.

Best,
Jingxuan


slozier:
@LostBenjamin Thanks for the PR. This code comes from the standard CPython library so I have no idea what the intent of the test is. May be worth filing an issue on https://github.com/python/cpython to see if it is indeed a bug...

slozier:
Thanks!

PR #1496, opened 2022-06-17
Resolve failure with jedi module

slozier:
The following hello world example is failing:
```py
import jedi
script = jedi.Script(print('Hello World!')
completions = script.complete(1, 3)
```

The failure appears to be caused by `types.MemberDescriptorType` being `<class 'getset_descriptor'>` instead of `<class 'member_descriptor'>`.

In the initial report (by @SebastianBolte on gitter) the sample was run in a hosted scenario and failing with a `TypeError: object of 'NoneType' has no len()`:
```cs
engine.SetSearchPaths(seachPaths);
engine.Execute(@"
import jedi
script = jedi.Script(print('Hello World!')
completions = script.complete(1, 3)
");
```

This is because `sys.executable` is `None` by default. The solution to this is to set its value with `engine.SetHostVariables`. For example:
```cs
engine.SetHostVariables(@"C:\Program Files\IronPython 3.4", @"C:\Program Files\IronPython 3.4\ipy.exe", string.Empty);
```

### DIFF
--- a/Src/StdLib/Lib/types.py
+++ b/Src/StdLib/Lib/types.py
@@ -38,7 +38,7 @@ except TypeError:
 
 # For Jython, the following two types are identical
 GetSetDescriptorType = type(FunctionType.__code__)
-MemberDescriptorType = type(FunctionType.__globals__)
+MemberDescriptorType = type(ModuleType.__dict__["__dict__"]) # ironpython: type(FunctionType.__globals__) is getset_descriptor
 
 del sys, _f, _g, _C,                              # Not for export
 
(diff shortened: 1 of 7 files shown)


Src/StdLib/Lib/types.py
```diff
@@ -38,7 +38,7 @@ def _m(self): pass
 
 # For Jython, the following two types are identical
 GetSetDescriptorType = type(FunctionType.__code__)
-MemberDescriptorType = type(FunctionType.__globals__)
+MemberDescriptorType = type(ModuleType.__dict__["__dict__"]) # ironpython: type(FunctionType.__globals__) is getset_descriptor
```
BCSharp, line 41: From the comment above it seems that Jython has the same problem as IronPython. Maybe submit a patch upstream?

slozier: > Maybe submit a patch upstream?

Definitely worth considering. I'm actually not sure what the technical differences are between the two types. Most of out descriptors seem to be of the type `getset_descriptor`. The codebase does have `member_descriptor` so I just poked around until I found one, but I don't know what makes this one special...

Issue #266, opened 2017-03-02
Create separate posix module for Unix based platforms


slide:
We currently just remove builtins['nt'] and assign builtins['posix'] to be the nt module. There is some overlap in functionality, but there are some things that would be much easier to implement in separate modules. Another option would be to use a #if check on Mono builds and have to separate releases, one for Windows and one for Unix based systems. This also doesn't even bring in the concept of running on OSX, which may have some additional changes.

slozier:
Separate modules sounds fine as long as we can share the common code.

PR #1500, opened 2022-07-04
Use P/Invoke for os.strerror

BCSharp:
Besides producing correct error message on Linux and macOS, the main benefit is the elimination of the switch on constants from `PythonErrorNumber`. The values in module `errno` vary depending on OS so cannot be implemented as constants.

On Windows, P/Invoke calls into _ucrtbase_, which may not be available on Windows 7 by default. I don't know if IronPython is still supposed to support Windows 7, if so, the the WIX installer may have to install _msvcrt_ on that system.

This PR also contains a few small fixes of `Mono.Unix` usage (ref. https://github.com/IronLanguages/ironpython3/pull/1490#issuecomment-1157003535)

On a side note: I'm wondering if it wouldn't be cleaner to factor module `posix` out of `nt` (and putting common functionality in an additional static class, e.g. `PythonOS`).

### DIFF
--- a/Src/IronPython.Modules/nt.cs
+++ b/Src/IronPython.Modules/nt.cs
@@ -1496,7 +1496,7 @@ namespace IronPython.Modules {
         public static object stat(CodeContext context, int fd)
             => fstat(context, fd);
 
-        public static string strerror_r(int code) {
+        public static string strerror(int code) {
 #if FEATURE_NATIVE
             const int bufsize = 0x1FF;
             var buffer = new StringBuilder(bufsize);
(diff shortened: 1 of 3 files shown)


slozier:
Sorry for the delay in reviewing. Am away from my dev PC for the week so just doing this on my phone.

Microsoft isn't supporting win7 so I don't see why we should (although apparently a lot of people are still using it). Is there no API that can give us the same result without ucrtbase? How does the pinvoke call on Windows compare to `new Win32Exception(errorCode). Message`? Interestingly they're adding `Marshal.GetPInvokeMessage(errorCode)` to .NET 7. In any case, if it's too much effort then too bad for win7...

As for the split posix/nt modules I'm not opposed to it - though it may be hard to untangle. We have issue #266 which is suggesting the same.

BCSharp:
> Microsoft isn't supporting win7 so I don't see why we should (although apparently a lot of people are still using it).

Fine with me. A helpful note in a readme, release notes or something to install the needed library should be sufficient. 

> Is there no API that can give us the same result without ucrtbase?

Not that I am aware of. Of course, it is always possible to run a script and generate code that has all Windows messages hardcoded. Looks like a waste of memory to me but it would remove the dependency. Another (marginal, untested) benefit of using ucrtbase is that it should produce different messages depending on locale.

> How does the pinvoke call on Windows compare to `new Win32Exception(errorCode). Message`?

`Win32Exception(errorCode).Message` gives the message for Windows error (aka `winerror`), while the p/invoke call I introduced gives the message for C error (aka `errno`). `PythonExceptions._OSError.WinErrorToErrno` can convert from `winerror` to `errno` but there is no way to unambiguously convert in the opposite direction, simply because `winerror` is more specific.

For instance `winerror` `ERROR_FILE_NOT_FOUND` (value 2) has message _The system cannot find the path specified._; `ERROR_PATH_NOT_FOUND` (value 3) has message _The system cannot find the path specified._. However both, plus some six more `winerror` codes are mapped to `errno` `ENOENT` (value 2), which has message _No such file or directory_.

> Interestingly they're adding `Marshal.GetPInvokeMessage(errorCode)` to .NET 7.

I didn't know that; too bad it comes so late. Luckily all supported Windows versions have ucrtbase (if I am not mistaken).

> As for the split posix/nt modules I'm not opposed to it - though it may be hard to untangle.

Yes, it may be. The main reason I brought it up is that any time `Mono.Unix` is used it is being factored out to a separate method, with a comment _Isolate Mono.Unix from the rest of the method so that we don't try to load the Mono.Unix assembly on Windows._ I have been following this approach in my changes though I do not fully understand the mechanics; if I don't factor the code out it still seems to work OK. As module `posix` gets more complete, this approach may create quite some clutter.

slozier:
> I have been following this approach in my changes though I do not fully understand the mechanics; if I don't factor the code out it still seems to work OK.

I think this only occurs outside of the dev environment since the Mono.Unix dll isn't packaged on Windows. Although I haven't checked this since we switched from Mono.Posix to the Mono.Unix package.

PR #1502, opened 2022-07-06
Support module errno on Linux and macOS

BCSharp:
This PR requires #1500.


slozier (approved):
Looks good to me.

I think this might cause some merge conflicts when trying to merge into 3.6 since https://github.com/IronLanguages/ironpython3/pull/1491 changes some stuff in `PythonSocket.MakeException`... Hopefully I can just add the `WouldBlock` case without breaking any of your changes.

BCSharp:
Hmm, it wasn't my intention to work on socket and impact 3.6, but it turned out that it was dependent on errno. After a cursory reading of #1491 I think that a special case for `WouldBlock` will not be needed anymore. In a couple of days I may try to merge master into 3.6 and resolve the conflicts, while my memory of the socket code changes is still fresh.

PR #1503, opened 2022-07-06
Implement make target "purge"

### LINKED ISSUE #1436: Working directory not fully cleaned by ./make clean
The impact of this issue has been reported in #1422 [(comment)](https://github.com/IronLanguages/ironpython3/pull/1422#issuecomment-1114059297) and #1433. Below is a copy of the discussion under #1433 about a possible solution.

----
Ok, managed to reproduce the failure. I think there might be a few issues:

[...]  
* Staging folder is not getting cleared. This might cause other junk to make its way into the package...
* `msi\obj` is not getting cleaned? Somehow I managed to hit the same issue you had and zapping this folder "fixed" the issue.

_Originally posted by @slozier in[ #1433 (comment)](https://github.com/IronLanguages/ironpython3/pull/1433#discussion_r862866416)_

----
I have noticed before that none of the 17 `obj` dirs are ever being cleaned, which on some occasion caused me problems. They usually contain both `Release` and `Debug` subdirs with often some files directly in `obj`. Thinking about a fix, it is not clear which target should clean them. Maybe a new target `clean-all`?

_Originally posted by @BCSharp in [#1433 (comment)](https://github.com/IronLanguages/ironpython3/pull/1433#discussion_r863233788)_

----
I guess part of the question is how much cleaning is expected of "clean". Like if you create a brand new .NET console program then run `dotnet build` and `dotnet clean` there's still a lot of stuff left over in the `obj` folder. Anyway, if we want to force clean everything maybe it could be called `purge`?

_Originally posted by @slozier in [#1433 (comment)](https://github.com/IronLanguages/ironpython3/pull/1433#discussion_r863278240)_

BCSharp:
Resolves #1436.

`purge` is a utility target, that is it does not use parameters `frameworks` or `configuration`; it always purges for all configurations and for all frameworks.

The reason to purge for all configurations is that some files being purged are not configuration-specific.

The reason to purge for all frameworks is that it allows to zap leftovers from frameworks no longer supported.

And another reason for both is that it was simpler to implement.

Note that `purge` deletes all produced binaries, and even package dependencies (though not the package cache). It is almost like a fresh checkout but with all source file edits or additional files kept in place. 

Also, the DLR submodule is left intact. I haven't had problems with it in the past.

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -214,6 +214,36 @@ function Test([String] $target, [String] $configuration, [String[]] $frameworks,
     RunTestTasks $tasks
 }
 
+function Purge() {
+    Main "Clean" "Release"
+    Main "Clean" "Debug"
+
+    Write-Verbose "Deleting ""obj"" directories..."
+    Get-ChildItem -Name "obj" -Directory -Path "$_BASEDIR" -Recurse | Remove-Item -Force -Recurse
+
+    Write-Verbose "Deleting ""bin"" directories..."
+    foreach ($dir in @("", (Join-Path "IronPythonAnalyzer" "IronPythonAnalyzer"))) {
+        if (Test-Path (Join-Path $_BASEDIR $dir "bin" -OutVariable targetPath)) {
+            Remove-Item -Path $targetPath -Force -Recurse
+        }
+    }
+
+    Write-Verbose "Deleting "".binlog"" files..."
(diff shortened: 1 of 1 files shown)


slozier (approved):
LGTM! Thanks

PR #1507, opened 2022-07-15
Merge branch 'master' into 3.6 (includes errno and socket updates)

BCSharp:
Below are a few observations from the merge. I haven't attempted to address any of them beyond the need to keep the tests passing; my goal was to have this commit be purely about the merge, and not mix it with anything else.

1. There were some conflicts in `_socket` but all were pretty clearly resolvable. However what got my attention were the changes in 3.6 in `__init__` to handle parameter `fileno`. I wonder if using `Converter.TryConvertToInt64` is the right way. Here is what CPython does:

```pycon
Python 3.10.2 (tags/v3.10.2:a58ebcc, Jan 17 2022, 14:12:15) [MSC v.1929 64 bit (AMD64)] on win32
>>> import socket
>>> socket.socket(socket.AddressFamily.AF_INET, socket.SocketKind.SOCK_STREAM, 0, "12")
  File "C:\Tools\Python310\lib\socket.py", line 232, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
TypeError: 'str' object cannot be interpreted as an integer
>>> socket.socket(socket.AddressFamily.AF_INET, socket.SocketKind.SOCK_STREAM, 0, (1<<63))
  File "C:\Tools\Python310\lib\socket.py", line 232, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
OverflowError: int too big to convert
>>> socket.socket(socket.AddressFamily.AF_INET, socket.SocketKind.SOCK_STREAM, 0, -1)
  File "C:\Tools\Python310\lib\socket.py", line 232, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
ValueError: negative file descriptor
>>> class myindex:
...     def __index__(self): return 42
...
>>> socket.socket(socket.AddressFamily.AF_INET, socket.SocketKind.SOCK_STREAM, 0, myindex())
  File "C:\Tools\Python310\lib\socket.py", line 232, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

This was CPython 3.10 but CPython 3.6 does the same except the last error, which is:

```pycon
>>> socket.socket(socket.AddressFamily.AF_INET, socket.SocketKind.SOCK_STREAM, 0, myindex())
  File "C:\Users\Pawel\.conda\envs\py36\lib\socket.py", line 144, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
TypeError: an integer is required (got type myindex)
```
Perhaps something like `TryToIndex` would make ipy more like CPython.

2. A couple of tests in `test_functools_stdlib` are now causing stack overflow (and were previously simply failing), both calling `test_recursive_repr`. Perhaps due to the new `__repr__` implementation on `master`, which can call itself recursively in some cases. I have commented out both tests.

3. `test_argparse` from StdLib can fail if the console window is wider than 80 columns. This is also true when testing CPython, but still kind of inconvenient. I guess something to keep in mind when testing locally.

4. `test_ssl_stdlib` fails now on `test_get_server_certificate_ipv6 (test.test_ssl.NetworkedTests)`. This is because `_ssl.do_handshake` calls `_sslStream.AuthenticateAsClient(server_hostname ?? _socket._hostName, ...)` but both `server_hostname` and `_socket._hostName` are `null`. I have investigated this a little bit and it seems to me that both being `null` is a legitimate case (though .NET does not support it). This is because `get_server_certificate` in `ssl.py` calls `create_connection(addr)` (in `socket.py`) which creates a socket without a host name for some reason, then `context.wrap_socket(sock)` creates a sslsock without `server_hostname` set. It tried to modify `do_handshake` to use the IP address of of the remote endpoint of the socket if the host name is `null`, and it works, but then it causes other tests to hang indefinitely when the remote IP is `127.0.0.1`. Apparently .NET does not like it either. It still puzzles me how could this test ever have passed on CI; when I rerun the tests on a pristine 3.6 checkout (i.e. before the merge), this test fails too.


slozier:
Thanks for the merge! There are a few failures in CI but I'll just merge as-is and fix later.

>     1. ... I wonder if using `Converter.TryConvertToInt64` is the right way. ... Perhaps something like `TryToIndex` would make ipy more like CPython.

I agree `TryConvertToInt64` is certainly not correct. The change was just to get things back in working order and the call to `HandleToSocket(long)` required a `long`. I'll file an issue based on your observations so we don't forget about it.

>     2. A couple of tests in `test_functools_stdlib` are now causing stack overflow (and were previously simply failing), both calling `test_recursive_repr`. Perhaps due to the new `__repr__` implementation on `master`, which can call itself recursively in some cases. I have commented out both tests.

Thanks will take a look.

>     3. `test_argparse` from StdLib can fail if the console window is wider than 80 columns. This is also true when testing CPython, but still kind of inconvenient. I guess something to keep in mind when testing locally.

I wonder if a later version of CPython has a fix for this - we could backport...

>     4. `test_ssl_stdlib` fails now on `test_get_server_certificate_ipv6 (test.test_ssl.NetworkedTests)`. ... It still puzzles me how could this test ever have passed on CI; when I rerun the tests on a pristine 3.6 checkout (i.e. before the merge), this test fails too.

On my machine this test is being skipped, probably the same thing that occurs during CI:
```
test_get_server_certificate_ipv6 (test.test_ssl.NetworkedTests) ... skipped "Resource 'ipv6.google.com' is not available"
```

slozier (approved):

slozier:
Oops, probably shouldn't have done a squash and merge... ah well.

BCSharp:
> Oops, probably shouldn't have done a squash and merge... ah well.

It think rebase and merge would preserve the history better... Will see if I can repair it.

BCSharp:
Couldn't do rebase and merge because there is already a branch off that commit (slozier/tests). So resorted to a dummy merge commit.

BCSharp:
Dummy merge didn't look good: the history of changes was ambiguous. So I undid the changes and did the rebase-merge.

Unfortunately, this means some inconvenience for you. To get a clean merge from slozier/tests you can best transplant the one commit on that branch, maintaining that same branch name, to the new HEAD of 3.6 and force-push the branch.

If you want, and if I have write access to that branch, I could do the changes for you.

slozier:
Thanks for fixing up my mistake. Looks much better with the proper merge. No worries about breaking my branch, easily fixed.

PR #1511, opened 2022-07-17
Make IPv6 SSL test passing

BCSharp:
Since I do have IPv6 connectivity, I decided to look into `test_get_server_certificate_ipv6` that is being skipped on CI but fails locally.

The good new is that my solution I tried during #1507 turned out to be the right way, despite causing now some other tests (that were previously failing) to hang indefinitely. The change unblocked those tests to advance and hit another, unrelated problem.

The bad news is that I had to comment them out to let the whole test suite pass, rather than have them as _expected failure_. I also think I got to the root cause of the problem: creating a socket using a descriptor of another existing socket resets the timeout of the existing socket to default (infinite by default). This is because creating a socket around an existing descriptor does not actually create a new system socket, it merely shares it. I see no obvious solution to it so I will file a separate issue where it can be discussed it more detail.


slozier (approved):
Looks good to me.

PR #1515, opened 2022-07-20
Re-enable some tests

slozier:


slozier:
@BCSharp any idea about the macOS failures on Mono with `test_ssl_stdlib` and `test_httplib_stdlib`? I'm fine with disabling the failing tests on Mono+macOS if it's not a quick fix (or if you have no time to look at it).

slozier:
Grammar failure occurs when run after `test_function`:
> .\make.ps1 -f net462 test-"IronPython.test_function|test_grammar_stdlib"

Failure appears to be (note that this only works with CPython 3.7+ since older versions don't accept > 255 arguments):
```py
# from test_function
size = 257
d = {}
exec('def f(' + ','.join(f'a{i}' for i in range(size)) + '): pass', d)
exec('f(*(' + ', '.join(map(str, range(size))) + '))', d)

# from test_grammar
def d01(a=1): pass

d01(*(1,))
```

BCSharp:
I can look into Mono/macOS failures in the next couple of days.

BCSharp:
The problem in `test_ssl_stdlib` was easy to fix: when setting timeout on a closed socket, Mono throws `ArgumentException` instead of `SocketException` with `SocketError.InvalidArgument` like .NET. I have pushed the patch onto your test branch.

The problem in `test_httplib_stdlib` is more elusive. I see that the certificate chain passed to verify is empty, therefore the verification of a "good" certificate fails, but don't know why it happens only on Mono. Apparently, CPython 3.6 on macOS  no longer uses system root certificates (see [BPO-29065](https://bugs.python.org/issue29065)) and presumably package [certifi](https://pypi.org/project/certifi/) is the answer, but I could not verify this because I coudn't get pip to work on Mono/macOS. So to investigate it will take more time.

slozier:
Thanks for looking into these.

> The problem in `test_ssl_stdlib` was easy to fix: when setting timeout on a closed socket, Mono throws `ArgumentException` instead of `SocketException` with `SocketError.InvalidArgument` like .NET. I have pushed the patch onto your test branch.

Interestingly it's now causing `test_timeout` failures on .NET Core (raising an `OSError` instead of `ValueError`). At least I'm able to reproduce this new issue on Linux.

> The problem in `test_httplib_stdlib` is more elusive. I see that the certificate chain passed to verify is empty, therefore the verification of a "good" certificate fails, but don't know why it happens only on Mono. Apparently, CPython 3.6 on macOS no longer uses system root certificates (see [BPO-29065](https://bugs.python.org/issue29065)) and presumably package [certifi](https://pypi.org/project/certifi/) is the answer, but I could not verify this because I coudn't get pip to work on Mono/macOS. So to investigate it will take more time.

I guess I should check if pip works on Linux, would be an unfortunate regression if it didn't...

BCSharp:
> The problem in `test_httplib_stdlib` is more elusive. […] presumably package [certifi](https://pypi.org/project/certifi/) is the answer

I copied `certifi` from CPython 3.6 (Anaconda), but by itself, it is not sufficient, it has to be installed. The [install script](https://github.com/python/cpython/blob/560ea272b01acaa6c531cc7d94331b2ef0854be6/Mac/BuildScript/resources/install_certificates.command) uses `_ssl.get_default_verify_paths()`, which is not implemented in IronPython.

BCSharp:
> Interestingly it's now causing `test_timeout` failures on .NET Core (raising an `OSError` instead of `ValueError`). At least I'm able to reproduce this new issue on Linux.

It is also failing on Windows and Mono/macOS, but the test is disabled on Mono so it appears successful.

```ini
[CPython.test_timeout]
RunCondition=NOT $(IS_MONO) # TODO: investigate
```
Perhaps it would be enough to check the range of the argument and throwing `ArgumentException` explicitly, rather than relying on the exception thrown on setting the timeout on the socket.

BCSharp:
> Grammar failure occurs when run after `test_function`:

Funny, function failure occurs when run after `test_grammar`.

Error in the original test script:

```
  File "run.py", line 10, in <module>
TypeError: d01() takes at most 1 argument (1 given)
```

Error when `f` and `d01` are executed in reversed order:
```python
# from test_grammar
def d01(a=1): pass
d01(*(1,))

# from test_function
size = 257
d = {}
exec('def f(' + ','.join(f'a{i}' for i in range(size)) + '): pass', d)
exec('f(*(' + ', '.join(map(str, range(size))) + '))', d)
```

```
  File "run.py", line 9, in <module>
  File "<string>", line 1, in <module>
TypeError: f() takes exactly 257 arguments (513 given)
```

slozier:
> Funny, function failure occurs when run after `test_grammar`.

I did look into this one a bit. Both function calls are single argument list calls and so are treated as "simple" calls and so it only checks the `PythonFunction.FunctionCompatibility` value for equivalence. The failure started to occur due to a change in the `PythonFunction.CalculatedCachedCompat` function. Both of the functions end up with a function compatibility value of 257 so it's trying to use the code from one function to unpack the other. I thought it was a regression but the issue is also present in IronPython 2.7 (just unlikely to occur due to the number of arguments involved). For example:
```py
# this one fails with IronPython 2.7
exec("def f(" + ",".join("a" + str(i) for i in range(0x4001)) + "): pass")
f(*range(0x4001))

def g(a=1): pass

g(*(1,))
```

Anyway, I'm still poking around looking for a good idea to resolve this. I'm wondering if it might be possible to remove the keyword-only arguments from the function compat value. Another option might to to set `FunctionBinderHelper._needCodeTest` if there are > 255 arguments.

Side note, the list argument unpacking seems a bit inefficient, it does:
```c#
if (list.__len__() != 0) {
    return list.pop(0);
}
```
until the list is exhausted...

BCSharp:
> I did look into this one a bit.

Thanks for keeping me posted. This is one of the corners of IronPython I have never visited before, so it is illuminating to first try to figure things out myself, get lost, and then see the real answer. 😃 

A few observations from my side:

1. If I am not mistaken, CPython can handle up to 2 ** 31 -1 arguments. While matching that insanely high number is not really a hard requirement, 255 is rather low, esp. when using code generation. I think ipy2 limit of 2 ** 14 -1 was a reasonable compromise. Regardless the limit however, I think it would be better if the compilation failed the moment the code defines a function with more parameters than the limit, rather than the current spooky behaviour.
2. `PythonFunction.FunctionCompatibility` is defined as `int` and all this call info has to be crammed into 32 bits. Since this property is checked at every call, I understand that the comparison should be very efficient (thus precluding an elaborate data struct), but at least it could be `ulong`, giving twice the capacity. On 64-bit systems (i.e. all contemporary systems), it will be practically the same performance. But frankly, given all that is going on when a function call is being made, even a modest-sized struct would be probably acceptable.
3. `_needCodeTest` for supersized function signatures sounds fine — those need not be efficient, but should still work correctly.
4. I have noticed the parameter unpacking code too and indeed it seems unreasonably inefficient, especially when comparing to the great lengths the code goes on other points to be very efficient with such calls.

BCSharp (approved):
Looks good to me!

PR #1516, opened 2022-07-28
Fix function compatibility

slozier:
Resolves an invalid `TypeError` when defining/calling certain functions. See https://github.com/IronLanguages/ironpython3/issues/1515 for details.

---
Grammar failure occurs when run after `test_function`:
> .\make.ps1 -f net462 test-"IronPython.test_function|test_grammar_stdlib"

Failure appears to be (note that this only works with CPython 3.7+ since older versions don't accept > 255 arguments):
```py
# from test_function
size = 257
d = {}
exec('def f(' + ','.join(f'a{i}' for i in range(size)) + '): pass', d)
exec('f(*(' + ', '.join(map(str, range(size))) + '))', d)

# from test_grammar
def d01(a=1): pass

d01(*(1,))
```

Broke in https://github.com/IronLanguages/ironpython3/pull/447

_Originally posted by @slozier in https://github.com/IronLanguages/ironpython3/issues/1515#issuecomment-1191722994_

### DIFF
--- a/Src/IronPython/Runtime/Binding/MetaPythonFunction.cs
+++ b/Src/IronPython/Runtime/Binding/MetaPythonFunction.cs
@@ -365,11 +365,21 @@ namespace IronPython.Runtime.Binding {
             /// Makes the test for our rule.
             /// </summary>
             private BindingRestrictions/*!*/ GetRestrictions() {
-                if (!Signature.HasKeywordArgument()) {
-                    return GetSimpleRestriction();
+                if (_extractedDefault) {
+                    return BindingRestrictions.GetInstanceRestriction(_func.Expression, _func.Value);
+                } else if (_needCodeTest) {
+                    return GetSimpleRestriction().Merge(
+                        BindingRestrictions.GetInstanceRestriction(
+                            Expression.Property(
+                                GetFunctionParam(),
+                                "__code__"
+                            ),
+                            _func.Value.__code__
+                        )
+                    );
                 }
 
-                return GetComplexRestriction();
+                return GetSimpleRestriction();
             }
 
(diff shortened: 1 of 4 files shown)


BCSharp (commented):
To hopefully help me better understand this part of the codebase, I wanted to ask some questions.

1. If I understand it correctly, there is no specific limit in IronPython on the number of parameters a function can have (besides the inherent limits of .NET, like the maximum array size etc.) The `0xff` or now the `0x3ff` limit is just a soft limit, under which function invocations of compatible functions can share call sites. It that right?

2. Assuming the above point is true, what is the benefit of expanding (for the purpose of optimized handling) the number of normal arguments from `0xff` to `0x3ff`, as well as the other two types of arguments, at the expense of the keyword-only default argument count? I'd think that functions having at least one keyword-only default parameter are more frequent than functions with more than 255 parameters. Or is it so that call sites of functions with keyword-only default parameters are inherently not shareable?

3. I have noticed that you decided not to widen the value of `CalculatedCachedCompat` to 64 bits. Any reason for this? Is it because, as I speculate above, call sites of functions with keyword-only default parameters are not shareable, so one bit reserved for those parameters is sufficient anyway?

4. The test in property `NeedsCodeTest` uses operator `>=` for comparison, e.g. `NormalArgumentCount >= 0x3ff`. Based on my current understanding, the test `NormalArgumentCount > 0x3ff` would suffice. So is value `0x3ff` reserved for some purpose? Or maybe the whole `CalculatedCachedCompat` of value -1 has some meaning?

BCSharp:
Thanks, it makes a lot more sense now.

Issue #20, opened 2014-03-21
PEP 3115 -- Metaclasses in Python 3000


jdhardy:
http://www.python.org/dev/peps/pep-3115/

moto-timo:
Initial CPython implementation: https://hg.python.org/cpython/rev/19f7ff443718

moto-timo:
Very early beginnings of implementation: https://github.com/moto-timo/ironpython3/tree/pep-3115-metaclass-syntax

moto-timo:
Getting closer: https://github.com/moto-timo/ironpython3/tree/pep-3115

slozier:
Revert https://github.com/IronLanguages/ironpython3/pull/169 hack once this is properly implemented.

BCSharp:
[Originally](https://github.com/IronLanguages/ironpython3/issues/1208#issuecomment-834757250) I intended to remove `__classcell__` and replace it with an extra parameter to `PythonOps.MakeClass`, but since Python 3.6 introduces it and IronPython 3.4 already implements a number of Python 3.6 behaviour, it is probably not worth it.  Instead of it, I plan now to implement the Python 3.6 behaviour fully.

There are two significant changes in Python [3.6](https://docs.python.org/3.6/reference/datamodel.html#creating-the-class-object) WRT [3.4](https://docs.python.org/3.4/reference/datamodel.html#creating-the-class-object) related to handling of `__class__`:

> **CPython implementation detail:** In CPython 3.6 and later, the `__class__` cell is passed to the metaclass as a `__classcell__` entry in the class namespace. If present, this must be propagated up to the `type.__new__` call in order for the class to be initialised correctly. Failing to do so will result in a [DeprecationWarning](https://docs.python.org/3.6/library/exceptions.html#DeprecationWarning) in Python 3.6, and a [RuntimeError](https://docs.python.org/3.6/library/exceptions.html#RuntimeError) in Python 3.8.

And:

> When a new class is created by `type.__new__`, the object provided as the namespace parameter is copied to a new ordered mapping and the original object is discarded. The new copy is wrapped in a read-only proxy, which becomes the [__dict__](https://docs.python.org/3.6/library/stdtypes.html#object.__dict__) attribute of the class object.

At the second step `__classcell__` is filtered out rather than deleted from the original namespace dict.

So here is my revised plan:

[...]

PR #1517, opened 2022-07-29
DeprecationWarning on failed __classcell__ propagation

BCSharp:
Part of #20.

`test__ssl` fails but not fixing here since a fix is already in the pipeline of #1516.


slozier (approved):
Seems fine to me.

Don't really know much about this stuff so I tried to come up with something which would trigger the warning. What I came up with doesn't seem to trigger anything. Is it something that would be resolved by some other part of your plan?
```py
class MyDict(dict):
    def __setitem__(self, key, value):
        pass

class MetaClass(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        return MyDict()

class MyClass(metaclass=MetaClass):
    def test(self):
        return __class__
```

BCSharp:
> Is it something that would be resolved by some other part of your plan?

It would have been, but I have reshuffled some code to include your test in the test suite now.

PR #1520, opened 2022-07-30
Improve handling of __classcell__

BCSharp:
Part of #20


Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1526,8 +1539,10 @@ public static string Repr(CodeContext/*!*/ context, object? o)
                 }
             }
 
-            static object? CallPrepare(CodeContext/*!*/ context, object meta, string name, PythonTuple bases, PythonDictionary? keywords, PythonDictionary dict) {
-                object? classdict = dict;
+            static object? CallPrepare(CodeContext/*!*/ context, object? meta, string name, PythonTuple bases, PythonDictionary? keywords, PythonDictionary dict) {
```
slozier, line 1542: Is the return type still nullable?

slozier (approved):
Looks good to me. Not sure if you looked at these, but 3.6 has some `__classcell__` tests under `test_super`.

Src/IronPython/Runtime/Operations/PythonOps.cs
BCSharp: No, it is not indeed. But frankly the whole `CallPrepare` is slated to be reworked in the last step. I'm going to change the nullability now though, because you never know when there will be time to finish it.

BCSharp:
> Not sure if you looked at these, but 3.6 has some `__classcell__` tests under `test_super`.

No, I was not aware of that test, thanks. Will look into it soon.

PR #1531, opened 2022-08-08
Use CommonDictionaryStorage for class variables

BCSharp:
Part of #20.

Because it effectively changes the name lookup mechanism in a class body lambda, I was curious what performance impact this may have. The old mechanism uses `RuntimeVariableDictionaryStorage`, which is backed by a fixed length array and a linear lookup. In the new code, only `__class__` goes there, all other variables are stored in `CommonDictionaryStorage`, which grows dynamically.  Since the new mechanism is only required for classes created by user metaclasses, it is possible to fall back on the old mechanism if no metaclass is specified, thus maintaining the existing performance profile in most cases.

Intuitively, the old mechanism should be better for small (few methods) classes, while the new approach should outperform for very large classes. But the question was where the break-even is and how much gain/loss it entails.

Here are best-case test results for classes of various sizes. _N_ is the number of methods (with empty body). On top of that, each class has a method `g` returning `__self__`, which adds symbols `g`, `__class__`, `__classcell__`. Plus every class has the mandatory `__module__`. (CPython classes also have `__qualname__`).

|  _N_ | CPython* |    New |    Old |
| ---: | ------: | -----: | -----: |
|    0 |  4.8 μs | 3.2 μs | 3.1 μs |
|   10 |  6.4 μs | 5.1 μs | 5.1 μs |
|  100 |   17 μs |  24 μs |  23 μs |
|  500 |   66 μs | 120 μs | 154 μs |

*CPython is for reference only, runs with GC disabled.
Standard deviation was in the order of 0.5 μs at the low end, 5 μs around _N_=100, and approx. 10 μs for _N_=500.

What it tells me that any differences between the two algorithms are negligible for any sensibly-sized class. The bulk of time is probably spend on parsing and namebinding, and the choice of storage is not a significant factor.

### DIFF
--- a/Src/IronPython/Compiler/Ast/ClassDefinition.cs
+++ b/Src/IronPython/Compiler/Ast/ClassDefinition.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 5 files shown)


Src/IronPython/Compiler/Ast/ClassDefinition.cs
```diff
@@ -240,6 +259,9 @@ public class ClassDefinition : ScopeStatement {
             }
         }
 
+        private MSAst.Expression SeLocalName(string name, MSAst.Expression expression)
```
slozier, line 262: Is this a typo? `SetLocalName`?

slozier (approved):
Don't kow much about this part of the codebase, but looks good to me.

Issue #1519, opened 2022-07-30
iter not called on subclasses of bytes


slozier:
For example:
```py
class test(bytes):
    def __init__(self):
        self.it= False
    def __iter__(self):
        self.it = True
        return self    
    def __next__(self):
        raise StopIteration

# via for loop
t = test()
for x in t: pass

assert t.it

# via list
t = test()
list(t)

assert t.it
```

This issue also seems to occur for other iterable base classes:
- `str`: fails on list call
- `tuple`: StackOverflowException
- `list`: StackOverflowException
- `collections.deque`: fails on for loop and list call
- `set`: fails on for loop and list call
- other cases?

Hit this failure while trying to run `ipy -m pip install pyserial`.

slozier:
I guess it's not exclusive to `pyserial`. Unable to pip install anything tonight...

PR #1532, opened 2022-08-09
Get pip working again

slozier:
Workaround for https://github.com/IronLanguages/ironpython3/issues/1519

### DIFF
--- a/Src/IronPython/Runtime/Bytes.cs
+++ b/Src/IronPython/Runtime/Bytes.cs
@@ -1163,6 +1163,10 @@ namespace IronPython.Runtime {
         #region IEnumerable Members
 
         System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator() {
+            // workaround for https://github.com/IronLanguages/ironpython3/issues/1519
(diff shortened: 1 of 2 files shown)


BCSharp:
This is just a workaround? What then would be a proper solution?

slozier:
> This is just a workaround? What then would be a proper solution?

I'm not sure, I guess one option is to patch all our `IEnumerable` classes but it feels more like a core issue. It also seems a bit icky to have `IEnumerable.GetEnumerator` and `IEnumerable<byte>.GetEnumerator` returning different results. Another issue is what happens when the base `IEnumerable` is out of our control? In this case, patching `GetEnumerator` to handle subclasses is no longer an option:

```py
class test(System.Collections.Generic.List[int]):
    def __iter__(self):
        return iter("ABC")
```

To complicate matters, what happens if you're trying to use this in C#? I'm not even sure how normal iterators behave if you try to foreach over them...

It feels like on the Python side of things going through `__iter__` would be the expected behaviour for all user subclasses. I looked at tweaking `PythonProtocol.MakeEnumeratorOperation` to accomplish this which seemed to resolve the issue with for loops, however it didn't help with things like `list(test())`, `max(test())`, etc. Probably need to do something in `PythonOps.GetEnumerable` as well.

Anyway, I didn't want to spend too much time on this and considering this is not a new issue, I decided to opt for the quick workaround to get `pip` back in working order.

Issue #562, opened 2019-03-04
LocalsDictionary not generated in certain cases


slozier:
```Python
a = locals
def test():
    return a()

assert test() == {}
```
It looks like there's an optimization to only create a locals dictionary when needed. However the way it determines if it's needed is by looking at the name of the call expression (see `IronPython.Compiler.Ast.CallExpression.NeedsLocalsDictionary`). `NeedsLocalsDictionary` returns false in the example above.

Similarly this will also fail for `dir`, `vars`, `eval` and `exec`...

BCSharp:
For now, a possible workaround is to run `ipy` with full frames enabled (commadline option `-X FullFrames`).

PR #1535, opened 2022-08-18
Resolve obscure super errors

BCSharp:
It started when I looked into why `super()` generates `NameError`/`UnboundLocalError` while in CPython it is `RuntimeError` (tested for by StdLib). It turns out that the CPython's implementation hence behaviour is different in some corner cases. Specifically, 0-argument `super()` does not have default arguments `__class__` and `self` injected during compilation, but resolved during runtime. One of the consequences is that `__class__` is always resolved to a class cell rather than whatever the namebinder finds in the local scope.

As far as I can see, these are truly obscure cases that would lead to IronPython/CPython differences and almost always bad Python programming, but it was an interesting exercise and you never know what other consequences might be there.

For what I know, most IronPython/CPython differences are resolved now, the only few remaining are in some pathological and totally useless constructs, in which the error message (but not type) might be different. See the new tests for specifics.

### DIFF
--- a/Tests/test_super.py
+++ b/Tests/test_super.py
@@ -321,7 +321,7 @@ class SuperTest(IronPythonTestCase):
         self.assertEqual(C().f().__thisclass__, C)
         self.assertEqual(C().f().__self_class__, C)
 
-        # A class using suped in a parameterles method still has a classcell
+        # A class using super in a parameterles method still has a classcell
         class C(metaclass=AssertHasClasscell):
             def f():
                 return super()
@@ -370,7 +370,7 @@ class SuperTest(IronPythonTestCase):
         # Similarly, super() in a generator expression does trigger classcell generation
         class C(metaclass=AssertHasClasscell):
             test = (super() for _ in range(1))
-        # The class creation succeeds because the generator is nor evaluated yet
+        # The class creation succeeds because the generator is not evaluated yet
         # but the generator fails when it is evaluated
         msg = "super(type, obj): obj must be an instance or subtype of type"
         if is_cli:
@@ -387,7 +387,7 @@ class SuperTest(IronPythonTestCase):
         self.assertEqual(C().f().__thisclass__, C)
         self.assertEqual(C().f().__self_class__, C)
 
-        # Using super though an alias does not trigger classcell generation, although the alias works OK
+        # Using super through an alias does not trigger classcell generation, although the alias works OK
         s_p_r = super
         class C(metaclass=AssertHasNoClasscell):
             def f(self):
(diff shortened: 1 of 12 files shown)


Tests/test_super.py
```diff
@@ -277,4 +277,130 @@ def f():
...
+        # This difference highlights a difference in the way IronPython and CPython handle comprehensions.
+        # In CPython the for-expression is assigned to a parameter of the comprehension context.
+        # IronPython optimizes that parameter away, and the for-expression is more like
+        # an anonymous free variable from the outer context.
+        # See also: https://github.com/IronLanguages/ironpython3/pull/1130
+
+        # Similarly, super() in a generator expression does trigger classcell generation
+        class C(metaclass=AssertHasClasscell):
+            test = (super() for _ in range(1))
+        # The class creation succeeds because the generator is nor evaluated yet
```
slozier, line 360: typo: not

Tests/test_super.py
```diff
@@ -277,4 +277,130 @@ def f():
...
+
+        # A regular super(cls, self) call also triggers classcell generation
+        # though __class__ is not used
+        class C(metaclass=AssertHasClasscell):
+            def f(self):
+                return super(C, self)
+        self.assertEqual(C().f().__thisclass__, C)
+        self.assertEqual(C().f().__self_class__, C)
+
+        # Using super though an alias does not trigger classcell generation, although the alias works OK
```
slozier, line 377: typo: through

Tests/test_super.py
```diff
@@ -277,4 +277,130 @@ def f():
...
+                return __class__
+
+        # A class using super() has a classcell
+        class C(metaclass=AssertHasClasscell):
+            def f(self):
+                return super()
+        self.assertEqual(C().f().__thisclass__, C)
+        self.assertEqual(C().f().__self_class__, C)
+
+        # A class using suped in a parameterles method still has a classcell
```
slozier, line 311: typo: super

slozier (approved):
Thanks for the PR! I'm sure I've come across the super alias issue before but I can't find any issue about it... Looks good to me. My understanding of this part of the code base is somewhat limited so I will defer to you - feel free to merge when you're happy with it.

In case you're interested in another obscure error:
```py
class C:
    def f2(self):
        s_p_r = super
        return s_p_r()

C().f2()
```

BCSharp:
> In case you're interested in another obscure error:
> ```python
> class C:
>     def f2(self):
>         s_p_r = super
>         return s_p_r()
> 
> C().f2()
> ```


Thanks, it is a good one. I think I will adjust the part that triggers the class cell generation, but not the part that lifts `arg[0]` to a cell. Class cell generation is fairly innocuous, being executed only once per class, so very low performance impact, but lifting `self` to a cell in every case `super` is referred to means that the price is being paid for every call of such method and every statement in that method using `self`. It feels like a bad deal, just to support another obscure scenario. Note that all parameters get lifted unconditionally if _FullFrames_ are used, so there will be a workaround.

It a way it then becomes another case like #562. So this will be fully solved if #562 gets solved.

PR #1542, opened 2022-08-31
Execute class body lambda in outer context

BCSharp:
Part of #20.

The "outer context" is basically a local context but created outside of the class lambda. I call it "class context" rather than "local context" because, strictly speaking, the class lambda still creates its own local context inside. The content of both class and local contexts is the same, except that the latter contains the `__class__` variable, if present. I kept the creation of the local context like this since the machinery to handle it is in `ScopeStatement`, which servers all scope-creating constructs: class, function, comprehensions, generators, and AST root, so it was easier to implement it this way. Besides, having a separate context without `__class__` may be useful, I remember in the past having to filter `__class__` out explicitly on some occasions. Perhaps this can be simplified now.

Incidentally, this PR removed some old and obscure incompatibly with CPython (`test_sys_getframe`); I take it as a sign that the changes are going in the right direction.


slozier (approved):
Looks good to me.

Issue #1428, opened 2022-05-01
Handle async/await syntax


slozier:
Add basic async/await syntax support so we don't raise a `SyntaxError` when trying to use the 3.6 standard library.

In order to bootstrap things, https://github.com/IronLanguages/ironpython3/pull/1427 will comment out some async/await related code in the standard library. These changes should be reverted (if possible) as part of this issue.

slozier:
There's also code commented in https://github.com/IronLanguages/ironpython3/pull/1544 that should be enabled once we handle the syntax.

PR #1544, opened 2022-09-02
Re-enable some tests

slozier:
Notes:
- `test_asyncio.test_selector_events.test_sock_connect_resolve_using_socket_params` was blocking and was updated to use the version in 3.7.9.
- `TestOneTrickPonyABCs.test_Collection` in `test_collections` was amended to use the behaviour from 3.7.

Related issues:
- https://github.com/IronLanguages/ironpython3/issues/1428
- https://github.com/IronLanguages/ironpython3/issues/1550


BCSharp:
One small point more: `test_class` from StdLib contains some new tests wrt 3.4, so `test_class_stdlib` needs to be regenerated.

PR #1546, opened 2022-09-02
Support arbitrary PythonStorage as backing for RuntimeVariablesDictionaryStorage

BCSharp:
Part of #20.

### DIFF
--- a/Src/IronPython/Runtime/CustomDictionaryStorage.cs
+++ b/Src/IronPython/Runtime/CustomDictionaryStorage.cs
@@ -11,38 +11,38 @@ using Microsoft.Scripting.Runtime;
 
 namespace IronPython.Runtime {
     internal abstract class CustomDictionaryStorage : DictionaryStorage {
-        private readonly CommonDictionaryStorage/*!*/ _storage;
+        private DictionaryStorage/*!*/ _storage;
 
         public CustomDictionaryStorage() {
-            _storage = new();
+            _storage = new CommonDictionaryStorage();
         }
 
-        protected CustomDictionaryStorage(CommonDictionaryStorage storage) {
+        protected CustomDictionaryStorage(DictionaryStorage storage) {
             _storage = storage;
         }
 
-        public CommonDictionaryStorage Storage => _storage;
+        public DictionaryStorage Storage => _storage;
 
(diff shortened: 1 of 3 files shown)


Src/IronPython/Runtime/CustomDictionaryStorage.cs
```diff
@@ -11,38 +11,38 @@
...
             Add(key, value);
         }
 
         public override void AddNoLock(ref DictionaryStorage storage, object key, object value) {
-            if (key is string && TrySetExtraValue((string)key, value)) {
+            if (key is string skey && TrySetExtraValue(skey, value)) {
                 return;
             }
-            _storage.AddNoLock(ref storage, key, value);
+            _storage.AddNoLock(ref _storage, key, value);
```
slozier, line 34: This class is confusing me - what happens to the `storage` argument? I guess I just have a hard time wrapping my head around the design of `DictionaryStorage`...

slozier (approved):
Seems fine to me if you're confident about the usage of `DictionaryStorage`.

Src/IronPython/Runtime/CustomDictionaryStorage.cs
BCSharp: I believe it was a mistake to use `storage` here. The usage pattern for all methods with a `ref` parameter is that it ought to be the same as the object on which the method is invoked. This is so because if an object of a `DictionaryStorage` subtype is immutable, it has to replace itself with an updated copy.

In this particular instance the mistake was inconsequential because `_storage` was always `CommonDictionaryStorage`, which is mutable so it ignores the `ref` argument. But now that `_storage` can be of any type, the call has to be correct. 

This is not the most elegant API in the world, but it is designed this way for performance reasons, I believe. One of the unfortunate consequences is that `PythonDictionary._storage` must be a non-private, non-readonly field. But since `DictionaryStorage` is internal, oh well, I guess it is OK…

BCSharp: OK, I think I understand why it was `storage` here before. It was not a mistake. `_storage` could not have been used because it was `readonly`, but using `storage` was safe knowing it will be ignored.

PR #1548, opened 2022-09-05
Enable nullable annotations in StringDictionaryStorage

BCSharp:
Also fixes handling of entries with null or empty string keys.

Frankly, I struggle to see any benefit of `StringDictionaryStorage` over e.g. `CommonDictionaryStorage`. The one thing that `StringDictionaryStorage` does better is the fast test `HasNonStringAttributes`, but since the keys have to be enumerated anyway to match keyword arguments to parameter names, any check against non-strings could be done at that moment, with no extra cost. This is how CPython handles it, I believe.

Perhaps `Dictionary<string, object?>` is faster than the buckets of `CommonDictionaryStorage`. If so, I wonder if `ConcurrentDictionary<string, object?>` wouldn't be even better, but I haven't done any performance measurements. All theses implementations do not preserve order, but order preservation is needed for 3.6, so a new design is needed anyway.

### DIFF
--- a/Src/IronPython/Runtime/StringDictionaryStorage.cs
+++ b/Src/IronPython/Runtime/StringDictionaryStorage.cs
@@ -2,9 +2,12 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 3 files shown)


slozier (approved):
Looks good to me! I have no about the benefits over `CommonDictionaryStorage`, but I assume it's perf related. Sometimes I wish we had some commit history going further back to get an idea as to why some things are the way they are... In any case, as you said, the whole thing will need to be revisited for order preservation.

PR #1561, opened 2022-09-14
Enable nullable annotations in Ast.ScopeStatemenet

BCSharp:


Src/IronPython/Compiler/Ast/ScopeStatement.cs
```diff
@@ -318,43 +256,47 @@ public abstract class ScopeStatement : Statement {
...
                     }
                 }
             }
         }
 
         internal virtual void FinishBind(PythonNameBinder binder) {
-            List<ClosureInfo> closureVariables = null;
+            List<ClosureInfo>? closureVariables = null;
+
+            if (FreeVariables is not null && FreeVariables.Count > 0) {
```
slozier, line 290: Should we change this to `IsClosure`?

BCSharp: Ah, I forgot that `MemberNotNullWhenAttribute` exists. Never used it before.

slozier:
Thanks!

Issue #1560, opened 2022-09-14
Dictionary lookup of class namespace should not apply to explicitly declared variables


BCSharp:
A dictionary lookup behaviour is standard for classes. This means that any locally referenced name is first looked up in the class attribute dictionary (local namespace), before looked up in the enclosing lexical scopes. This is true even if the variable to be looked up is late-bound, e.g. introduced at runtime by modifying `locals()` or though metaclass `__prepare__`. This behaviour is already implemented in #1151 in general and in recent #1551 for metaclasses.

For instance:

```python
prepared = "global"

class C:
    executed = prepared  # lexical name binding

assert C.executed == "global"

class Meta(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        return {'prepared' : "meta"}

class C(metaclass=Meta):
    executed = prepared  # namespace dictionary lookup

assert C.executed == "meta"
```

However, when a variable is explicitly declared global, the default namebinding rules are overridden and the name has to resolve to a global variable:

```python
prepared = "global"

class Meta(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        return {'prepared' : "meta"}

class C(metaclass=Meta):
    global prepared
    executed = prepared  # explicit lexical name binding

assert C.executed == "global"
```

This is not how IronPython handles it currently.

-----

An interesting story emerges when the variable is declared `nonlocal` rather than `global`. One (like me) would expect it too to override default namebinding rules and force lexical lookup. However, all CPython versions up to and including 3.10 still do local dictionary lookup:

```python
class Meta(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        return {'prepared' : "meta"}

def f():
    prepared = "f-local"
    class C(metaclass=Meta):
        nonlocal prepared
        executed = prepared  # which prepared?

    assert C.executed == "meta"  # ah, namespace dictionary lookup

f()
```

IronPython currently matches it, simply because none of this explicit declaration is taken into account yet. But as a programmer and IronPython developer I see it as a bug. Perhaps nobody reported it yet to CPython, or a report has been made, but got a very low priority. After all, this is a very unlikely piece of code that would reveal the bug. And arguably bad style.

Another demonstration that this is likely a bug rather a deliberate design is that references are resolved differently depending whether they are read (namespace lookup) or written (lexical lookup). To modify the example above a little:

```python
class Meta(type):
    @classmethod
    def __prepare__(metacls, name, bases):
        return {'prepared' : "meta"}

[...]

Issue #1028, opened 2020-11-12
reloading a builtin module does not restore deleted members


slozier:
This issue comes from `test_namebinding` and was reported at https://github.com/IronLanguages/main/issues/462.

While the running `reload` adds the method back to the `builtins` modules, it doesn't seem to restore `pow` in the global context. Note that this issue doesn't seem to occur when running the code directly in the interpreter.
```Python
def test():
    import builtins
    from importlib import reload
    try:
        del builtins.pow
    finally:
        reload(builtins)
        assert builtins.pow(2,2) == 4
        assert pow(2,2) == 4 # this line fa

test()
```

An equivalent repro (with 3.4) is: `ipy -c "import builtins; del builtins.pow; import _imp; _imp.init_builtin('builtins'); pow"` which results in a `NameError: name 'pow' is not defined`.

slozier:
This issue is specific to CPython 3.4. With subsequent versions, `importlib.reload(builtins)` does not restore `pow` or `builtins.pow` so closing the issue.

PR #1563, opened 2022-09-16
Resolve explicit globals in class lexically

BCSharp:
Handles the `global` part from #1560, making IronPython's behaviour consistent with CPython. Changes to `HasLateBoundVariableSets` are just for consistency; this property is not used when binding names in a class definition.

I intend to handle the `local` part in a separate PR (assuming no other tests fail), for clarity and to allow an easy revert in case it's needed.

`test_namebinding.py` fails when run with `ipy` from the command line, but passes when run with `make test`. This is an unrelated issue, likely #1028, although it is now closed. The issue has all the symptoms of a bug in the expression rewriting code, so that would be my guess.

### DIFF
--- a/Src/IronPython/Compiler/Ast/ClassDefinition.cs
+++ b/Src/IronPython/Compiler/Ast/ClassDefinition.cs
@@ -89,12 +89,8 @@ namespace IronPython.Compiler.Ast {
         private PythonVariable? ClassVariable { get; set; }
 
         internal override bool HasLateBoundVariableSets {
-            get {
-                return base.HasLateBoundVariableSets || NeedsLocalsDictionary;
-            }
-            set {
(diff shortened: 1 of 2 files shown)


slozier:
The repro for the test failure is as follows:
```py
import builtins
from importlib import reload
del builtins.pow
reload(builtins)
assert "pow" in dir(builtins)
pow(2,2)
```
Note that it also does not fail when run in the REPL so to reproduce the failure it should be saved to a file and executed.

I don't think it's quite the same as https://github.com/IronLanguages/ironpython3/issues/1028 since that one was about the change in behaviour between 3.4 and 3.5. In 3.5 the `assert` fails whereas with this bug the `assert` succeeds and the `pow` fails.

PR #1566, opened 2022-09-17
Enable nullable annotations in Ast.Node

BCSharp:
Point of interest: `Node.Parent`.

This property is initialized to `null` upon construction and stays `null` up to the beginning of the first namebinding phase. So technically its type is nullable (may be null). However, it is used a lot during and after namebinding and during namebinding it is guaranteed to be non-null by `PythonNameBinder`, while it should not be accessed at all beforehand. So a non-nullable type is a better fit. This is what I chose then with some guards around it.

One caveat: previously, the was one exception the rule above: the root node (`PythonAst`) had `Parent` set to `null` at all times. I've changed this so that the root node is its own parent. This makes the nullability of the property consistent with other node types and allows safe and well defined meaning of expressions like `Parent.Parent...` (I remember I wished to had that when I was working on comprehensions; this is currently not used anywhere).

As a result, the test `node.Parent == null` is no longer a valid test to detect the root node. The canonical form of this test remains `node.IsGlobal` (although at one place I saw an equivalent of `node is PythonAst`, which works too).

----

Something else: `Debug.Assert(x != null)` has an unpleasant side effect that the nullability analyzer assumes that `x` may be null after such statement, even if `x` was determined as non-null before. A workaround is to use `Assert.NotNull(x)`, which pushes `Debug.Assert` one level deeper thus does not suffer from this phenomenon. Unfortunately, after `Assert.NotNull` the analyser still may assume `x` being nullable if it was nullable before. It would be better if `Assert.NotNull` had `[NotNull]` declared on its parameter, but `Assert` is in the DLR, which does not refer to the DLL containing the nullability attributes. Perhaps it may be worth defining something like `PythonAssert.NotNull` with the proper attributes on the parameters?

### DIFF
--- a/Src/IronPython/Compiler/Ast/ClassDefinition.cs
+++ b/Src/IronPython/Compiler/Ast/ClassDefinition.cs
@@ -138,16 +138,11 @@ namespace IronPython.Compiler.Ast {
             }
 
             // Try to bind in outer scopes, except the global scope
-            for (ScopeStatement? parent = Parent;
-                parent != null && !parent.IsGlobal;
-                parent = !parent.IsGlobal ? parent.Parent : null) {
+            if (TryBindOuterScopes(this, reference, out variable, stopAtGlobal: true)) {
+                // for implicit globals, fall back on dictionary behaviour
+                if (variable.Kind is VariableKind.Global) return null;
 
-                if (parent.TryBindOuter(this, reference, out PythonVariable? outerVariable)) {
-                    // for implicit globals, fall back on dictionary behaviour
-                    if (outerVariable.Kind is VariableKind.Global) return null;
-
-                    return outerVariable;
-                }
+                return variable;
             }
 
             return null;
--- a/Src/IronPython/Compiler/Ast/ClassDefinition.cs
+++ b/Src/IronPython/Compiler/Ast/ClassDefinition.cs
@@ -138,7 +138,10 @@ namespace IronPython.Compiler.Ast {
             }
 
             // Try to bind in outer scopes, except the global scope
-            for (ScopeStatement parent = Parent; parent is not null && !parent.IsGlobal; parent = parent.Parent) {
+            for (ScopeStatement? parent = Parent;
+                parent != null && !parent.IsGlobal;
+                parent = !parent.IsGlobal ? parent.Parent : null) {
+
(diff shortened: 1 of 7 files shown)


slozier:
Regarding the nullability of `Debug.Assert(x != null)` I agree that it's annoying. We could add the nullability package to the DLR, but I'm not a huge fan of using the DLR `*Utils` classes. So having an internal `Assert` class is entirely reasonable (naming it `Assert` instead of `PythonAssert` could be an incentive to stop using `Microsoft.Scripting.Utils` - although we could always start with `PythonAssert`).

Src/IronPython/Compiler/Ast/ClassDefinition.cs
```diff
@@ -138,7 +138,10 @@ public class ClassDefinition : ScopeStatement {
             }
 
             // Try to bind in outer scopes, except the global scope
-            for (ScopeStatement parent = Parent; parent is not null && !parent.IsGlobal; parent = parent.Parent) {
+            for (ScopeStatement? parent = Parent;
```
slozier, line 141: I spent entirely too much time staring at this to wrap my head around the logic. I think in this case it could have been reduced to `for (ScopeStatement parent = Parent; !parent.IsGlobal; parent = parent.Parent)`? But I guess it's to be consistent with the other outer binding loops? Wondering if it's worth having a helper (I think this works?):
```C#
internal bool TryBindOuterScopes(ScopeStatement from, PythonReference reference, [NotNullWhen(true)] out PythonVariable? variable, bool stopAtGlobal = false) {
    for (ScopeStatement? parent = from.Parent; parent != null && !(stopAtGlobal && parent.IsGlobal); parent = !parent.IsGlobal ? parent.Parent : null) {
        if (parent.TryBindOuter(from, reference, out variable)) {
            return true;
        }
    }
    variable = null;
    return false;
}
```

slozier (approved):
Thanks! Looks good to me.

Src/IronPython/Compiler/Ast/ClassDefinition.cs
BCSharp: I like your idea.

Yes, I tried to keep `BindReference` consistent with other versions, to make it easier to compare them. When I started working on this part of code a few years ago I was often puzzled by (and suspicious of) all those subtle differences.

PR #1567, opened 2022-09-18
Resolve explicit nonlocals in class lexically

BCSharp:
Handles the `nonlocal` part from #1560, making IronPython's behaviour again inconsistent with CPython.


slozier (approved):
Maybe add the issue number next to test as well. Otherwise looks good to me.

BCSharp:
> Maybe add the issue number next to test as well.

I don't understand. The issue number (link) is already in the comment at the top of the test. Did you want to move it somewhere else?

slozier:
Ahh, I see it now. It was in the collapsed part of the diff.

PR #1571, opened 2022-09-28
Dotnet tool

### LINKED ISSUE #1570: IronPython as dotnet tool
Currently there is no easy way to install a standalone IronPython/.NET6.0 on Linux or macOS. The official _deb_ and _pkg_ installers install IronPython running on Mono. One convenient way to provide official installation of IronPython/.NET6.0 is to create a NuGet package of type `DotnetTool`, containing the IronPythonConsole app. Such package can be then installed user-wide with:

```powershell
dotnet tool install --global IronPython.Tool
```

If IronPython is used as part of the toolchain of a project, it can be installed project-locally:

```powershell
dotnet new tool-manifest  # if first tool
dotnet tool install IronPython.Tool
```
Such a project-bound tool can be easily restored by any developer cloning the project repository (assuming `.config/dotnet-tools.json` is checked in):

```powershell
dotnet tool restore
```

The third option is to install IronPython in a dedicated directory, effectively providing an alternative mechanism to _venv_.

```powershell
dotnet tool install --tool-path ./venv IronPython.Tool
```

In all cases, the IronPython interpreter (conlole app) will be invokable as `ipy`.

BCSharp:
Resolves #1570.

### DIFF
--- a/Package/dotnettool/IronPython.Console.csproj
+++ b/Package/dotnettool/IronPython.Console.csproj
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.Build.NoTargets/3.3.0">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <AssemblyName>ipy</AssemblyName>
+    <UseAppHost>false</UseAppHost>
+    <IncludeSymbols>false</IncludeSymbols>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <PackageOutputPath>$(PackageDir)</PackageOutputPath>
+
+    <PackAsTool>true</PackAsTool>
+    <ToolCommandName>ipy</ToolCommandName>
+    <PackageType>DotnetTool</PackageType>
+    <RestoreProjectStyle>DotnetToolReference</RestoreProjectStyle>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- nuspec properties -->
+    <PackageId>IronPython.Console</PackageId>
+    <Title>IronPython Command-line Interpreter</Title>
+    <Version>$(PackageVersion)</Version>
+    <Authors>IronPython Contributors, Microsoft</Authors>
+    <PackageProjectUrl>https://ironpython.net</PackageProjectUrl>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/IronLanguages/ironpython3.git</RepositoryUrl>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
+    <Description>IronPython is an open-source implementation of the Python programming language which is tightly integrated with the .NET Framework. IronPython can use the .NET Framework and Python libraries, and other .NET languages can use Python code just as easily.<br/>This package contains a standalone Python interpreter, invokable from the command line as <code>ipy</code>. It also includes the Python Standard Library released by the Python project, but slightly modified to work better with IronPython and .NET.</Description>
+    <PackageTags>ironpython python dynamic dlr</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="DynamicLanguageRuntime" Version="1.3.2" />
+    <PackageReference Include="Mono.Unix" Version="7.1.0-final.1.21458.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ToolFiles Include="$(StageDir)\$(TargetFramework)\ipy.deps.json" />
+    <ToolFiles Include="$(StageDir)\$(TargetFramework)\**\*.dll" Exclude="$(StageDir)\$(TargetFramework)\IronPythonTest.dll" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="$(StageDir)\LICENSE" />
+    <None Include="@(ToolFiles)"  CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+</Project>
--- a/Build.proj
+++ b/Build.proj
@@ -100,6 +100,16 @@
           DestinationFolder="$(StageDir)\%(RecursiveDir)" />
   </Target>
 
+  <Target Name="_CopyRuntimeConfigFiles" DependsOnTargets="_MakeStageDir">
+    <ItemGroup>
+      <RuntimeConfigFiles
+          Include="$(BinDir)\$(Configuration)\*\ipy.deps.json" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(RuntimeConfigFiles)"
+          DestinationFolder="$(StageDir)\%(RecursiveDir)" />
+  </Target>
+
   <Target Name="_CopyMainStageFiles" DependsOnTargets="_MakeStageDir">
     <Copy SourceFiles="@(MainStageFiles)" DestinationFolder="$(StageDir)" />
   </Target>
@@ -115,6 +125,7 @@
       _MakeStageDir;
       _PlatformStage;
       _CopyDlrMainReferences;
+      _CopyRuntimeConfigFiles;
       _CopyMainStageFiles;
       _CopyStdLib;
     </StageDependsOn>
--- a/Package/dotnettool/DotnetTool.Packaging.targets
+++ b/Package/dotnettool/DotnetTool.Packaging.targets
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+  <ItemGroup>
+    <DotnetTool Include="IronPython.Console" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <NuPkgOutputs>@(DotnetTool->'$(PackageDir)/%(Identity).$(PackageVersion).nupkg')</NuPkgOutputs>
+  </PropertyGroup>
+
+  <Target Name="DotnetToolPackage" DependsOnTargets="Stage" AfterTargets="Package" Outputs="$(NuPkgOutputs)">
+    <MSBuild Projects="$(MSBuildThisFileDirectory)\IronPython.Console.csproj" Targets="Restore" />
+    <MSBuild Projects="$(MSBuildThisFileDirectory)\IronPython.Console.csproj" Targets="Pack" Properties="$(BuildProperties)" />
+  </Target>
+</Project>
--- a/Package/dotnettool/runtimeconfig.template.json
+++ b/Package/dotnettool/runtimeconfig.template.json
@@ -1,0 +1,9 @@
+{
+    "framework": {
+        "name": "Microsoft.NETCore.App",
+        "version": "6.0.0"
+    },
+    "configProperties": {
+        "Microsoft.NETCore.DotNetHostPolicy.SetAppPaths": true
+    }
+}


slozier:
Ok this is pretty cool... 👍

BCSharp:
## Notes

### nuspec

I did not use a `nuspec` like the regular packages do because the .NET documentation recommends to use project files to configure `nupkg` metadata for new .NET projects and I was having some problems expressing everything directly in `nuspec` (esp. dependencies). For most of the `nuspec` metadata, there is an [equivalent project property](https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target), which is then used to generate a `nuspec` on the fly, but there are a few differences, described below.

### owners

The existing `nuspec` specifies `owners`, but there is no equivalent project property. `owners` is obsolete and _NuGet_ uses the uploader's ID as "owner". Whatever is specified in `nuspec` is ignored.

### summary

The existing `nuspec` specifies `summary`, but there is no equivalent project property. I assume it is obsolete too, I don't see the value of `summary` displayed anywhere on _NuGet_.

### language

The existing `nuspec` specifies `language`, but I haven't found an equivalent property. Using `Language` or `PackageLanguage` did not lead to `language` in the generated metadata. I don't know what this language is supposed to mean either. It is not the language of the DLLs, since those are `any`. Perhaps the language of the description. If so, _en_US_ is likely the default, so no harm done.

### copyright

The existing `nuspec` does **not** specify `copyright` so I did not include it either as a property. The .NET documentation recommends `copyright` be always specified. Despite no `Copyright` property, `copyright` is still included in the metadata, probably extracted from the first entry in `authors`. In our case it is _© IronPython Contributors_. Perhaps worth setting it explicitly.

### Other properties

The project file is probably over-specified, that is, some properties may be redundant or ineffective. The documentation is a bit terse. At one point it says that is its sufficient just to set `PackAsTool` and `ToolCommandName`, but this applies to ordinary projects that also are used to build and run the app. `IronPython.Console.csproj` is not like that; it builds nothing, just packs stuff from the staging directory. For instance, `PackageType` is ignored by `msbuild` (or set implicitly based on `PackAsTool`), but other packaging tools do use it, so I think it is better to set it explicitly just to a bit more future-proof. Maybe even `PackAsTool` is redundant since all included items have `Pack` and `PackagePath` set properly for a tool.

### RestoreProjectStyle

`DotnetTool` packages may only reference other `DotnetTool` packages, not the ordinary library (dependency) packages. At first I tried to set a reference to _DynamicLanguageRuntime_ and _Mono.Unix_ inside the generated package (rather than embedding those packages directly in the generated `nupkg`), but I hit [NU1212](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1212). Setting `RestoreProjectStyle` to the correct value did not help, but did not hurt either. I guess it is now ineffective because no other tool packages are referenced. I left it in just in case, some tool references are added in the future.

### Description

HTML tags in the description are not interpreted in any way, but the whole description can be superseded by a markdown file (at least on https://nuget.org). I used a tailored down version of the main `README.md`, which is included in the existing library packages. The full `README.md` is not appropriate, since it talks about project github status, engine embedding, current build status, has relative links, which will be broken, etc.

### Package icon

I intended to also pack an icon/logo for the package, but it would be good to do it for all three packages produced. Hence I move it out of this PR.

### RestoreUseStaticGraphEvaluation

This property is set explicitly to `false`, because for some reason static graph evaluation for restore does not restore any packages in `IronPython.Console.csproj`. Perhaps because it is an odd type of a project, I don't know. It is supposed to work, although [admittedly](https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#restoring-with-msbuild-static-graph-evaluation) it doesn't in all cases. Although `RestoreUseStaticGraphEvaluation` is `false` by default, it is slated to be enabled by default in some point in the future. Or possibly we enable it for the main IronPython build, to speed things up. Just in such case, setting it to `false` explicitly for packaging projects will prevent mysterious breakdown of packaging, I hope.

### Changes in `NuGet.Packaging.targets`

The package target in `DotnetTool.Packaging.targets` was mysteriously running out of memory on Linux CI, but not on other systems or my local Linux machine. It took me while to figure it out, but it seems to be fixed by properly passing all relevant build properties to the sub-build task doing restore. It was working OK for `NuGet.Packaging.targets` because `NuGet.Packaging.csproj` does not have any packages to restore. Nevertheless, it is a bug, which I backported to `NuGet.Packaging.targets` just in case there is something to restore there in the future.

### Package upload

I assumed (@slozier, correct me if I'm wrong) that you want to be the official owner and responsible of uploading this package to _NuGet_. So I have not uploaded anything to the official _NuGet_ repository. Besides, I did not know whether it would make sense to upload a current snapshot as 3.4.0-beta1, since it would not match the code of the other uploaded packages with the same versions. If the "final" release is coming up soon, then it would be moot anyway.

I have uploaded it to [_NuGet QA_](https://int.nugettest.org/packages/IronPython.Console/3.4.0-beta1.4) for testing. Hence I became the "owner" of it. If you want to use _NuGet QA_ for testing too, I will gladly transfer the ownership to you.

BTW, have you considered [reserving](https://learn.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation) the _IronPython_ prefix?

### Package versioning

Following the lead of the two existing packages, the package version is `3.4.-beta1`. This means, it can be uploaded only once to _NuGet_ (until a next release milestone). Also, any new build (from a new commit) still produces the same package version. This makes it not so convenient to upgrade the installation locally when a new build is made that is not an official release. For a dotnet tool, this requires uninstalling the tool completely and then reinstalling it again. Can be done, but all additional Python packages installed in the meantime are then gone. I guess `pip freeze` is my friend. But in general, I would prefer to have something like _SemVer_ compatible versioning. Perhaps this is a discussion worth having under a separate issue.

slozier:
> ### nuspec

Project file works for me.

> ### owners

Works for me, `nuspec` documentation says it's deprecated and recommends using authors instead, so we should clean up the other packages (outside the scope of this PR).

> ### summary

[`nuspec` documentation](https://learn.microsoft.com/en-us/nuget/reference/nuspec#summary) also says this is deprecated and to use description instead so fine by me.

> ### language

Indeed, there doesn't appear to be an equivalent property. Seems to be related to "localized packages" which this isn't really so no big deal.

> ### copyright

Agreed, we should probably set it explicitly everywhere (outside the scope of this PR).

> ### RestoreProjectStyle

Couldn't find useful doc relating to this so if it doesn't hurt I guess it's fine.

> ### Description

I guess we should consider a trimmed down version for the other packages as well (and including it in the nuspec)...

> ### Package icon

Would be interesting!

> ### RestoreUseStaticGraphEvaluation

I was wondering about this, if you think it may prevent issues in the future I guess this is fine (although one would hope they wouldn't turn it on by default if it's still buggy).

> ### Changes in `NuGet.Packaging.targets`

Sounds good.

> ### Package upload
> 
> I assumed (@slozier, correct me if I'm wrong) that you want to be the official owner and responsible of uploading this package to _NuGet_. So I have not uploaded anything to the official _NuGet_ repository. Besides, I did not know whether it would make sense to upload a current snapshot as 3.4.0-beta1, since it would not match the code of the other uploaded packages with the same versions. If the "final" release is coming up soon, then it would be moot anyway.

Yeah, don't think there's a point in having a beta1 available on NuGet with the final release coming up (probably going to start looking into it this week if you have anything else you want to squeeze in before then).

> I have uploaded it to [_NuGet QA_](https://int.nugettest.org/packages/IronPython.Console/3.4.0-beta1.4) for testing. Hence I became the "owner" of it. If you want to use _NuGet QA_ for testing too, I will gladly transfer the ownership to you.

Interesting, I have no heard of that before! Guess I should start using it (I always worry that I'm messing something up when publishing to nuget). As for the owner thing I think you can have multiple owners so feel free to add me.

> BTW, have you considered [reserving](https://learn.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation) the _IronPython_ prefix?

The though crossed my mind when I first heard about it, but I did not look into what was involved. Maybe worth revisiting if we're adding a 3rd package.

> ### Package versioning
> 
> But in general, I would prefer to have something like _SemVer_ compatible versioning. Perhaps this is a discussion worth having under a separate issue.

I guess the constraint of having the major/minor locked down to match Python doesn't leave a lot of room for manoeuvring. In my opinion having a different versioning scheme would probably just lead to confusion but we can certainly have the discussion.

slozier:
One small point, there are extra testing assemblies (e.g. `System.Drawing.Common.dll`) ending up in the package (also in the zip package). Not sure if it's related to this PR or something that I changed since the beta release.

BCSharp:
> One small point, there are extra testing assemblies (e.g. `System.Drawing.Common.dll`) ending up in the package (also in the zip package). Not sure if it's related to this PR or something that I changed since the beta release.

The following assemblies are being pulled in by `<PackageReference Include="DynamicLanguageRuntime" Version="1.3.2" />`: 

```
Microsoft.Dynamic.dll
Microsoft.Scripting.dll
Microsoft.Scripting.Metadata.dll
Microsoft.Win32.SystemEvents.dll
System.CodeDom.dll
System.Configuration.ConfigurationManager.dll
System.Drawing.Common.dll
System.Security.AccessControl.dll
System.Security.Cryptography.ProtectedData.dll
System.Security.Permissions.dll
System.Security.Principal.Windows.dll
System.Windows.Extensions.dll
```

The DLR does not target .NET 6 so I suppose the .NET Core 3.1 dependencies are being pulled in. Maybe worth targeting .NET 6 explicitly in the DLR package? Anyway, it does not seem to be caused by anything from this PR.

I haven't looked into the zip package story.

slozier:
> The DLR does not target .NET 6 so I suppose the .NET Core 3.1 dependencies are being pulled in. Maybe worth targeting .NET 6 explicitly in the DLR package? Anyway, it does not seem to be caused by anything from this PR.

I actually started looking at targeting .NET 6 explicitly in the DLR last night since since it's something I want to do for the release anyway (.NET Core 3.1 is nearing end-of-life), but I got sidetracked by all the COM stuff wanting `SupportedOSPlatform` annotations. Looking at the dependency graph I guess it's `System.Configuration.ConfigurationManager` bringing in all the those extra DLLs. I guess I should just remove the dependency from the nuspec since most people probably don't care about it (and those that might can opt in by manually adding the reference).

BCSharp:
> As for the owner thing I think you can have multiple owners so feel free to add me.

Sure, but you need to create an account at https://int.nugettest.org/. _NuGet QA_ accounts are not linked to _NuGet_ accounts.

BCSharp:
> > BTW, have you considered [reserving](https://learn.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation) the _IronPython_ prefix?
> 
> The though crossed my mind when I first heard about it, but I did not look into what was involved. Maybe worth revisiting if we're adding a 3rd package.

Maybe at some point a 4th package too, like _IronPython.WinApp_ (for an `ipyw` tool)…

slozier:
> > As for the owner thing I think you can have multiple owners so feel free to add me.
> 
> Sure, but you need to create an account at https://int.nugettest.org/. _NuGet QA_ accounts are not linked to _NuGet_ accounts.

Done

BCSharp:
> > ### RestoreUseStaticGraphEvaluation
> 
> I was wondering about this, if you think it may prevent issues in the future I guess this is fine (although one would hope they wouldn't turn it on by default if it's still buggy).

I hope that too but I wouldn't bet on it, if not needed. Because the dependencies in the packaging projects are so small. practically there is no significant performance gain from using the static graph evaluation algorithm. I expect it to be much more advantageous for the main build project.

BCSharp:
> > ### Package versioning
> > But in general, I would prefer to have something like _SemVer_ compatible versioning. Perhaps this is a discussion worth having under a separate issue.
> 
> I guess the constraint of having the major/minor locked down to match Python doesn't leave a lot of room for manoeuvring. In my opinion having a different versioning scheme would probably just lead to confusion but we can certainly have the discussion.

Yeah, _SemVer_ sensu stricto may not work for IronPython. I was more thinking about using some _SemVer_-supporting tools like _GitVersion_) to automaticaly manage a versioning scheme in which each commit produces a unique version. The actual versioning scheme is up for discussion, as well as the branching model. I can whip up a proposal to kick off a discussion. This need not be concluded before the "final" release.

BCSharp:
> the final release coming up (probably going to start looking into it this week if you have anything else you want to squeeze in before then).

Actually, I'd like to prepare a zip package suitable for Posix systems. .NET tool is cool, but it requires the .NET SDK installed. The zip file would be usable on systems with the .NET Runtime only.

Also I'd like to prepare a Homebrew cask that would be a fully standalone installation. This need not hold up the upcoming release, since Homebrew has its own syndication channels (like Chocolatey), so it can catch up later.

Package/dotnettool/IronPython.Console.csproj
```diff
@@ -0,0 +1,56 @@
...
+    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
+    <Description>IronPython is an open-source implementation of the Python programming language that is tightly integrated with the .NET Framework. IronPython can use the .NET Framework and Python libraries, and other .NET languages can use Python code just as easily.
+
+This package contains a standalone Python interpreter, invokable from the command line as ipy. It also includes the Python Standard Library released by the Python project, but slightly modified to work better with IronPython and .NET.</Description>
+    <PackageTags>ironpython python dynamic dlr</PackageTags>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="DynamicLanguageRuntime" Version="1.3.2" />
```
slozier, line 39: Wondering if we should grab the DLR from staging, the advantage being that we don't need a release in order to do a build with DLR updates.

slozier (approved):
Feel free to merge when you're happy with it.

Package/dotnettool/IronPython.Console.csproj
BCSharp: I'm going to park this option for now; depending on the outcome of the versioning discussion, it might become a useful option.

PR #1573, opened 2022-10-04
Remove int/long section from "Differences frm CPython"

### LINKED ISSUE #1572: Upgrade section about integers in differences from CPython
https://github.com/IronLanguages/ironpython3/blob/master/Documentation/differences-from-c-python.md#integers

The document still talks about `long` which is no longer applicable. I think the whole section about integers could be dropped since for all intents and purposes we have a single type within the Python realm. @BCSharp Any thoughts?

BCSharp:
Resolves #1572


slozier:
Thanks!

PR #1575, opened 2022-10-06
Improve ipy.sh

BCSharp:


slozier (approved):
Seems fine to me. I'm assuming this cases all the launchers to end up in the zip package?

BCSharp:
> I'm assuming this cases all the launchers to end up in the zip package?

Yes, that was the idea. In one of the next PRs I am going to include `Mono.Unix.dll` and friends. In this way, the zip package will be usable across all supported platforms and systems.

Issue #1472, opened 2022-05-24
Incorrect filemode for regular files installed on Posix systems


BCSharp:
Installing NuGet packages `IronPython` or `IronPython.StdLib` on Linux or macOS creates all regular files with user-executable bit set (4). This is not the convention used by Unix-like systems and confuses system tools (like auto-completion, which considers executable files as programs).

Note that all directories created should have the executable bit(s) set (as it is now) to make them searchable.

slozier:
I don't know how much control we have over the permissions during NuGet package creation. Not finding much documentation on the matter.

BCSharp:
I have never created a NuGet package, but if they have something like a post-install script, then it can do `chmod -x` on Posix systems.

mwchase:
I poked around a little, and it looks like this is an open bug in NuGet: https://github.com/NuGet/Home/issues/6778. So, I assume that until that bug is addressed, some form of workaround is necessary.

BCSharp:
Indeed, as @mvchase writes, this is a _NuGet_ problem and should be addressed there. As for a workaround here: since we have now a .NET tool package and a complete zip package, this issue is not so important anymore. The zip package install script sets permissions correctly, and the .NET tool installs the executable launcher in a separate directory than the rest of the files so no confusion for Unix tools occurs.

Therefore I am closing this issue as "won't fix without prejudice".

PR #1576, opened 2022-10-06
Set Unix file permissions in zip files

BCSharp:
This is somewhat related to #1472.

When the packaging target created a zip archive, the entries in the zip file had no file permissions set. As a result, unzipping the archive leaves behind a bunch of files that are not readable on Posix systems (until `chmod -R +r` at least).

Note that directories are created with proper permissions because they are not stored in the zip archive but are being created by the unzipping process.

In this PR, zip archive entries are created with the expected `rw-r--r--` permissions, except for `*.sh` and `*.ps1` files, which are supposed to be executable. The permissions are not read from the staged files because they don't exist on Windows (except for the _Archive_ flag, which I haven't bothered dealing with).


slozier (approved):
Seems fine to me. Unfortunate that we have to go through reflection for this...

PR #1579, opened 2022-10-07
Stage Mono.Unix

BCSharp:
This makes the zip archive usable on all supported platforms and systems.

The size of the archive is increased by 5 MB. Apparently no file-deduplication takes place. If we want deduplication, I suppose this would need to be manually handled by hacking the "Zip" task snippet. It would save ca. 2.5 MB. Frankly, I don't think it is worth the effort.

### DIFF
--- a/Build.proj
+++ b/Build.proj
@@ -100,6 +100,18 @@
           DestinationFolder="$(StageDir)\%(RecursiveDir)" />
   </Target>
 
+  <Target Name="_CopyMonoUnixReferences" DependsOnTargets="_MakeStageDir">
+    <ItemGroup>
+      <MonoUnixReferenceFiles
+          Include="$(BinDir)\$(Configuration)\*\Mono.Unix.dll" />
+      <MonoUnixReferenceFiles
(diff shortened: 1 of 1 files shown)


slozier (approved):
Looks good to me. Agreed that reduplication is not worth it.

Wondering if we should have some documentation about the different packages. For example how to install the tool, debian pkg, overview of the contents of the zip package (which files are required or not depending on platform), etc. Obviously not for this PR.

BCSharp:
> Wondering if we should have some documentation about the different packages. For example how to install the tool, debian pkg, overview of the contents of the zip package (which files are required or not depending on platform), etc.

I was thinking of writing a couple of project wiki pages explaining all that, and updating the main `README.md` with a link for details. Putting it in a wiki page instead of in `Documentation` would make it a more light-weight and easier to keep up to date, as the packages are released, or scripts adapted.

By the same token, I thought that `package-compatibility.md` would be better in a wiki page, except that there is so little update in this file recently that this idea did not feel urgent to be raised. Perhaps with a wiki page I would be quicker to report any successful package.

PR #1580, opened 2022-10-08
Do not stage IronPythonTest products

BCSharp:
This is staging directory cleanup. There is no reason to stage anything from IronPythonTest and then pollute the packaging targets with explicit exclusions.


slozier (approved):
Good idea.

PR #1582, opened 2022-10-11
Provide README for zip package

BCSharp:
Since the content is zip-file specific, it is being sourced from `Package/zip`, rather than the staging directory (similarly to the _NuGet_ packages). `Package/$(Configuration)/Stage/README.md`, which is a copy of the `README.md` in the project root directory, is, AFACS, now only used by the _MSI_ package.


slozier (approved):
Looks like there are merge conflicts after I merged the other PR. Otherwise looks good to me.

PR #1585, opened 2022-10-13
Update Install-IronPython and pack in zipfile

BCSharp:
`Install-IronPython` now can facilitate installation from within the zip file.

`Enter-IronPythonEnvironment` is also being distributed in the zip file.


slozier (approved):
Looks good to me. Don't mind the failing CI, I must have broken something with my commits - probably the DLR upgrade...

PR #1586, opened 2022-10-13
Update DLR

slozier:


slozier:
I bumped up the System.CodeDom package for .NET 6 in the DLR but when `make.ps1 package` is run it's restoring the old version (because of the DLR reference in the IronPython.Console project) to the bin folder and causing the tests to fail on github CI (which does packaging before testing). @BCSharp any good ideas?

BCSharp:
Strange that the package is being pulled in to the main bin directory. Would it help to replace the DLR package reference with regular references to the individual DLR DLLs in the stage directory?

Issue #1590, opened 2022-10-26
MissingMemberException in V3.4.0 beta 


nalsandori:
**# Previously, I had installed IronPython(Ver 2.7.9) in the  c# code and used it**
**# Now, I installed IronPython-3.4.0b1.msi**
**# But, there is an exception when I run c# code below**
`IronPython.Hosting.Python.CreateEngine()`
**Exception Message - System.MissingMemberException(IronPython.dll): 'type' object has no attribute '_ORIGIN'**
**# There was a problem like the above, so I rolled back to version 2.7.9 or 2.7.12**   
**→ It was good working!**
**# Please check the problem.**

slozier:
Where are you seeing this exception? It is from importlib and is caught so it should not cause any problems...

nalsandori:
> Where are you seeing this exception? It is from importlib and is caught so it should not cause any problems...

MissingMemberException is pointing ti IronPython.dll
The exception point is when Python.CreateEngin() is called
I used it well in previous versions(2.7.9)
My program implemented by WPF uses .NET Framework 4.7.2

slozier:
Yes, there is a `MissingMemberException` that occurs inside IronPython.dll when `Python.CreateEngine()` is called, but it is also caught inside IronPython.dll so you can ignore it. I do not understand the issue.

nalsandori:
> Yes, there is a `MissingMemberException` that occurs inside IronPython.dll when `Python.CreateEngine()` is called, but it is also caught inside IronPython.dll so you can ignore it. I do not understand the issue.

All right.
I will ignore those exceptions.
The reason why i asked this question is that the functikn associated with the eventhandler is not called when calling print.
`memorystream = new MemoriStream();
outputWr = new EventRaisingStreamWriter(memorystream)
outputWr.StringWritten += new EventHandler<MyEvtArgs<stribg>>(pythonOutput_Written);
engine = Python.CreateEngin()
engine.Runtime.IO.SetOutput(memoryStream, outputWr);
engine.Runtime.IO.SetErrorOutput(memoryStream, outputWr);`
I'll look further into thrle issue of events not being invoked.
If you have any advice on this issue, plz
Ah, when I import logging library, there are so many exception.
Can i ignore this message, too?
I wish you all the best

slozier:
[...]

PR #1594, opened 2022-11-16
Avoid exceptions when loading importlib

slozier:
Avoid exceptions when loading importlib. While the exceptions are normally caught, not having them occur in the first place makes for a more pleasant debugging experience. Related to https://github.com/IronLanguages/ironpython3/issues/1590

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
     - name: Install tools
-      if: matrix.os == 'ubuntu-latest'
+      if: matrix.os == 'ubuntu-20.04'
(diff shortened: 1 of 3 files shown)


BCSharp:
This is fantastic! It was long on my wish list, but I never got time to look into it.

slozier:
> This is fantastic! It was long on my wish list, but I never got time to look into it.

Wish I could somehow avoid more of them. Maybe something could be done to make zipimport.zipimporter light throwing, but unfortunately it's a class and not a function... I guess I diverge from Python and put a light throwing function in sys.path_hooks, but I'm not sure it it helps long term (zipimport got moved to the stdlib in 3.8).

Issue #1562, opened 2022-09-16
exe made by ipyc says "Error occurred: Object reference not set to an instance of an object"


retsyo:
### Description

The generated exe from simple code
```python
print("hello world")
```
can't run but says

```
Error occurred: Object reference not set to an instance of an object
```

### Steps to Reproduce

1. the official binary build, i.e. IronPython.3.4.0-beta1 can work on my windows 10 64 bits

D:\tmp>e:\prg\py\IronPython.3.4.0-beta1\net46\ipy.exe helloworld.py
hello world

2. but the generated exe by `ipyc` does not work
```
D:\tmp>e:\prg\py\IronPython.3.4.0-beta1\net46\ipyc.exe helloworld.py  /platform:x64 /standalone /target:exe
IronPython Compiler for IronPython 3.4.0b1 (3.4.0.0)
Input Files:
        helloworld.py
Output:
        helloworld
OutputPath:

Target:
        ConsoleApplication
Platform:
        AMD64

compiling...
Generating stand alone executable
        Embedded Microsoft.Scripting 1.3.2.0
        Embedded IronPython 3.4.0.0
        Embedded Microsoft.Dynamic 1.3.2.0
        Embedded IronPython.Modules 3.4.0.0
Saved to helloworld.exe

D:\tmp>helloworld.exe
Error occurred: Error occurred: Object reference not set to an instance of an object
```

3. old ironpython can produce proper exe
```
D:\tmp>e:\prg\py\IronPython.2.7.12\net45\ipyc.exe  helloworld.py  /platform:x64 /standalone /target:exe
IronPython Compiler for IronPython 2.7.12 (2.7.12.0)
Input Files:
        helloworld.py
Output:
        helloworld
OutputPath:

Target:
        ConsoleApplication
Platform:
        AMD64

compiling...
Generating stand alone executable
        Embedded Microsoft.Scripting 1.3.1.0
        Embedded IronPython 2.7.12.0
        Embedded Microsoft.Dynamic 1.3.1.0
        Embedded IronPython.Modules 2.7.12.0
Saved to helloworld.exe

D:\tmp>helloworld.exe
hello world
```

### Versions

```
D:\tmp>e:\prg\py\IronPython.3.4.0-beta1\net46\ipy.exe -V
IronPython 3.4.0b1
```

slozier:
Thanks for the report. My guess is that the standalone build is not embedding the `System.Memory` related dependencies.

[...]

PR #1599, opened 2022-11-25
Update IronPython.nuspec

scott-xu:
`System.Memory` was updated from 4.5.4 to 4.5.5 because of issue https://github.com/IronLanguages/ironpython3/issues/1562 and PR https://github.com/IronLanguages/ironpython3/pull/1565/
This change is to align the version in nuspec


slozier:
Thanks!

PR #1600, opened 2022-11-29
Fix unzipping 'lib' directory

BCSharp:
This used to work when tested on Windows, but now was failing on macOS. Differences in Powershell versions?...


slozier:
Thanks! We could probably set up a CI step that runs the script to test...

Issue #1523, opened 2022-08-01
`test_httplib_stdlib` failure on Mono+macOS


slozier:
The problem in `test_httplib_stdlib` is more elusive. I see that the certificate chain passed to verify is empty, therefore the verification of a "good" certificate fails, but don't know why it happens only on Mono. Apparently, CPython 3.6 on macOS  no longer uses system root certificates (see [BPO-29065](https://bugs.python.org/issue29065)) and presumably package [certifi](https://pypi.org/project/certifi/) is the answer, but I could not verify this because I coudn't get pip to work on Mono/macOS. So to investigate it will take more time.

_Originally posted by @BCSharp in https://github.com/IronLanguages/ironpython3/issues/1515#issuecomment-1192882018_

PR #1601, opened 2022-11-30
Version 3.4.0

slozier:


slozier:
Ran into a blocker, `clr.CompileModules` doesn't like this anymore (was working with beta1):
```py
class MyClass:
    """description"""
```
> SystemError: CompileToMethod cannot compile constant 'description' because it is a non-trivial value, such as a live object. Instead, create an expression tree that can construct this value.

@BCSharp if you have any ideas what change might have caused this (off the top of your head). If not no worries, I'll do a git bisect when I have a bit of time.

slozier:
Ran a git bisect and it looks like it stopped working in https://github.com/IronLanguages/ironpython3/pull/1531

BCSharp:
Will look into it tomorrow.

slozier:
Trying to run ensurepip from a root folder (e.g. `C:\` or `R:\`) results in a failure:
> R:\> ipy -m ensurepip --user
> Could not import runpy module

This is a regression from the beta. Seems to have broken in https://github.com/IronLanguages/ironpython3/pull/1484. Need to investigate...

slozier:
`os.listdir(".")` seems to be failing when run on `C:\` with the error:
> Traceback (most recent call last):
ValueError: Path cannot be the empty string or all whitespace.
Parameter name: path2

slozier:
> The same for `Directory.GetFiles` and `Directory.GetDirectories`. A bug in .NET Framework?

I noticed the same, it has all the symptoms of a bug in .NET Framework. Presumably one that was introduced with https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/retargeting/4.6-4.6.2#changes-in-path-normalization 

I was jumping back and forth between fixing in the PAL and directly in `listdir`. On the one hand, fixing in the PAL fixes it in more places, but on the other hand I assume the whole PAL concept is probably busted so I'm half tempted to abandon it altogether. I guess we'd need to try to target a platform that could maybe make use of it (e.g. browser-wasm) to see if there's any merit...

Anyway, since I haven't published the DLR yet I might as well fix it there and leave the PAL investigation for another time.

BCSharp:
> Anyway, since I haven't published the DLR yet I might as well fix it there and leave the PAL investigation for another time.

I suppose the workaround could have been wrapped in `#if NETFRAMEWORK` / `#endif` but it seems to have no detrimental side effects in .NET (Core) anyway.

BCSharp:
One more thing for 3.4.0 from my side: since some time, that is since one of the recent Windows 11 updates, `test_ssl_stdlib` fails deterministically on my machine. I was hoping somebody else would report it to make sure it is not some specific configuration issue at my side. Or that the issue would go away 😄 But it is still here. Anyway, it looks exactly like what was fixed in #1511 on the 3.6 branch. Maybe worth backporting?

slozier:
> > Anyway, since I haven't published the DLR yet I might as well fix it there and leave the PAL investigation for another time.
> 
> I suppose the workaround could have been wrapped in `#if NETFRAMEWORK` / `#endif` but it seems to have no detrimental side effects in .NET (Core) anyway.

Thought about the conditional but then I would have to add netstandard as well (just in case) and then it has to work with .NET (Core) anyway... Sometimes I wish they come up with .NET Standard. :smile:

> Anyway, it looks exactly like what was fixed in #1511 on the 3.6 branch. Maybe worth backporting?

I'm fine with the backport. Will add it shortly.

slozier:
Looks like the backport is causing new failures in `test_ssl_stdlib` on macOS+Mono. I'm assuming they're caused by the `ValidateCertificate` changes? I'm assuming it's the same issue as noted in https://github.com/IronLanguages/ironpython3/issues/1523. @BCSharp Do you need both parts of the PR to resolve your failure? Maybe we can only backport part of it? Otherwise I guess we'll need to skip the failing tests on macOS+Mono.

BCSharp:
Yes, both parts are needed, as far as I remember. I will look into the test situation later today.

slozier:
Will do some more testing and hopefully find time to prepare the release over the weekend.

PR #1602, opened 2022-12-05
Fix class __doc__ type

BCSharp:
Resolves issue raised in https://github.com/IronLanguages/ironpython3/pull/1601#issuecomment-1336308322.


slozier (approved):
Thanks for the quick fix! Seems to do the trick.

PR #1603, opened 2022-12-06
Update IronPythonTest.csproj

scott-xu:
Fix "System.ArgumentException: Unknown framework version 7.0" in step Test (net462)
image

Reference: https://github.com/nunit/nunit3-vs-adapter/issues/987


slozier (approved):
Oh nice, had not even noticed the failures. Thanks!

PR #1604, opened 2022-12-07
Update IronPython.wxs

scott-xu:
Add "VSNet45ReferenceAssemblyKey" back, so that VS can find the IronPython related dll files.
Reference: https://github.com/dotnet/msbuild/blob/main/src/Tasks/Microsoft.Common.CurrentVersion.targets#L641

### DIFF
--- a/Package/msi/IronPython.wxs
+++ b/Package/msi/IronPython.wxs
@@ -74,6 +74,11 @@
           Root="HKLM" Key="SOFTWARE\IronPython\$(var.ReleaseSeries)\InstallPath" Type="string" Value="[INSTALLDIR]" />
       </Component>
 
+      <Component>
+        <RegistryValue Id="VSNet45ReferenceAssemblyKey"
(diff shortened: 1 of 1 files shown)


scott-xu:
Not sure if it relates to https://github.com/IronLanguages/ironpython3/pull/1601#issuecomment-1339587979

slozier:
Why would you not use a PackageReference instead?

scott-xu:
I have a project which only refers (and only want to refer) IronPython.dll.
With PackageReference, the project will refer IronPython.Modules.dll, IronPython.SQLite.dll and IronPython.Wpf.dll

slozier:
PackageReference shouldn't add an assembly reference if you don't use anything from the assemblies. For example, if you only use `Python.CreateEngine` then your assembly will only reference IronPython.dll. You can also srt it up so the assemblies are only available at compile time (e.g. with `IncludeAssets="compile"`).

PR #1612, opened 2022-12-10
Use libc iso libdl on newer glibc versions

### LINKED ISSUE #1595: libdl assembly failing to load on Ubuntu 22.04 during CI
The failure occurs on `import ctypes` which on initialization invokes `_ctypes.dlopen`. This fails with a SystemError.

On Mono:
> SystemError: libdl assembly:<unknown assembly> type:<unknown type> member:(null)

On .NET:
> SystemError: Unable to load shared library 'libdl' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable:

BCSharp:
Resolves #1595. If the tests pass, maybe it can sneak into the 3.4.0 release?

### DIFF
--- a/Src/IronPython.Modules/_ctypes/NativeFunctions.cs
+++ b/Src/IronPython.Modules/_ctypes/NativeFunctions.cs
@@ -58,32 +58,72 @@ namespace IronPython.Modules {
         }
 
         // unix entry points, VM needs to map the filenames.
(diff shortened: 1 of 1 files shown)


slozier (approved):
Thanks!

Issue #751, opened 2020-02-28
test_strformat failing on .NET Core 2.1 on Linux


slozier:
Causes a failure in CI:
```Python
import _locale
_locale.setlocale(_locale.LC_ALL, 'en_US')
assert (1000.0).__format__('n') == '1,000'
```

Test disabled in https://github.com/IronLanguages/ironpython3/commit/2e4ac312174bee06a4fcd326138e6e2058858c01

PR #1622, opened 2022-12-18
Fix some localization issues

slozier:
Fixes some test failures which occur when changing my regional settings to french (which uses a comma as the decimal separator).


slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/751

PR #1624, opened 2022-12-19
Create Installing IronPython Binaries document

BCSharp:


Documentation/installing.md
```diff
@@ -0,0 +1,185 @@
+# Installing IronPython Released Binaries
+
+IronPython can be used as a standalone interpreter, or embedded within another .NET application. For embedding, the recommended approach is to install IronPython within the target project using NuGet. This document describes various ways of installing the standalone interpreter on the target system. Once installed, IronPython is invocable from the command line as `ipy`. Use `ipy -h` for a list of available commadline options. Use `ipy -VV` to produce version information for reporting issues.
+
+## IronPython on .NET (Core)
+
+Since .NET is a cross-platform framework, the instructions in this section apply to all supported operating systems (Windows, Linux, macOS), unless explicitly indicated.
+
+Currently IronPython is supported on .NET 6.0 and .NET Core 3.1.
```
slozier, line 9: .NET Core 3.1 just went out of support, wondering if we should be mentioning it.

Documentation/installing.md
```diff
@@ -0,0 +1,185 @@
...
+
+### .NET SDK
+
+If the target system has already a full .NET SDK installed, the most straightforward method to install a standalone IronPython interpreter is by using `dotnet tool`.
```
slozier, line 13: Maybe we can link to some Microsoft documentation for additional info, for example https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools

Documentation/installing.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Windows
+
+IronPython for .NET Framework requires .NET Framework version 4.6.2 or higher, which comes preinstalled on modern versions of Windows. To install IronPython, download the `.msi` file from the project's [release page](https://github.com/IronLanguages/ironpython3/releases/latest) and execute it. Alternatively, use _Chocolatey_ package manager:
+
+```
+choco install ironpython
+```
+
+The installation registers IronPython in the System Registry in compliance with [PEP 514](https://peps.python.org/pep-0514/) so that other tools may detect and use it.
```
slozier, line 130: I think the chocolatey installer doesn't do this...

slozier (commented):
Nice!

Documentation/installing.md
BCSharp: OK, I'll remove this line; one place less to keep track of when supported frameworks change.

Documentation/installing.md
BCSharp: Oh, that's too bad. Can it be done?

For now I'll reorder the lines so that this note comes before Chocolatey.

BCSharp: I also wonder how this would interact/conflict with a parallel installation though a `.msi`…

BCSharp:
> Did you plan on keeping the Standalone (TODO) section?

Yes, to invite comments, like about superiority or inferiority of Flatpak versus Snap, or "oh no, don't brew a whole cask, just brew a bottle" 😄 

Anyway, I am looking into Flatpak and Homebrew now, but have no plans for a "standalone" Windows support, Chocolatey or not. So if you too do not intend to develop a standalone installer for .NET on Windows, I'd rather remove this particular section.

slozier:
I have no experience with Linux and macOS packaging so whatever you prefer works for me!

> So if you too do not intend to develop a standalone installer for .NET on Windows, I'd rather remove this particular section.

No plans for the moment. I'm not even sure how I would proceed since for a proper cli experience you'd need to have the full runtime available. I guess I could do like PowerShell and package the whole(?) thing which is not very clean. Otherwise I'd have to enforce having a runtime installed on the system...

PR #1638, opened 2022-12-30
Fix fileno parameter handling for socket class

### LINKED ISSUE #1509: Use index instead of TryConvertToInt64 in socket.__init__
In the 3.6 branch, `socket.__init__` uses `Converter.TryConvertToInt64` on the `fileno`. This is incorrect and probably should be something like `TryToIndex`.

See comment by @BCSharp https://github.com/IronLanguages/ironpython3/pull/1507#issue-1306616504 for details.

Grub4K:
This adds the expected `fileno` parameter handling for the `socket.socket` class (See #1507). It now throws the expected errors before creating the socket instance.

It also backports tests for the `fileno` parameter from Python 3.11. `os_helper` and `socket_helper` get shimmed by classes with the used attributes from the respective module file if they cannot be imported.

Closes #1509

### DIFF
--- a/Src/StdLib/Lib/test/test_socket.py
+++ b/Src/StdLib/Lib/test/test_socket.py
@@ -30,6 +30,45 @@ try:
 except ImportError:
     fcntl = None
 
+# ironpython: backported shims for older python
+try:
+    from test.support import os_helper
+except ImportError:
+    class os_helper:
+        make_bad_fd = support.make_bad_fd
+        unlink = support.unlink
+
+try:
+    from test.support import socket_helper
+except ImportError:
+    class socket_helper:
+        HOST = "localhost"
+        HOSTv6 = "::1"
+
+        @staticmethod
+        def create_unix_domain_name():
+            """
+            Create a UNIX domain name: socket.bind() argument of a AF_UNIX socket.
+            Return a path relative to the current directory to get a short path
+            (around 27 ASCII characters).
+            """
+            return tempfile.mktemp(prefix="test_python_", suffix='.sock',
+                                   dir=os.path.curdir)
+
+        IPV6_ENABLED = False
+
+    if socket.has_ipv6:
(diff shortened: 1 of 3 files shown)


Src/StdLib/Lib/test/test_socket.py
```diff
@@ -1531,6 +1570,94 @@ def fileno(self):
...
+        with tempfile.NamedTemporaryFile() as afile:
+            with self.assertRaises(OSError):
+                socket.socket(fileno=afile.fileno())
+
+            with self.assertRaises(OSError) as cm:
+                socket.socket(
+                    socket.AF_INET,
+                    socket.SOCK_STREAM,
+                    fileno=afile.fileno())
+            self.assertEqual(cm.exception.errno, errno.ENOTSOCK)
```
slozier, line 1659: Looks like this is failing on Linux/macOS which have distinct error codes for non-socket descriptors and bad descriptors. Don't think our code really has a good way to distinguish between the two. We should either disable or tweak the test to something like:
```py
self.assertIn(cm.exception.errno, (errno.EBADF, errno.ENOTSOCK))
```

slozier (commented):
Thanks!

Src/StdLib/Lib/test/test_socket.py
Grub4K: After thinking about it I couldn't come up with something clever either. I would add it to the failing tests in `test_socket_stdlib` for now.
Later on throwing the proper error should happen automatically in the `HandleToSocket` function.
```py
        if is_posix:
            failing_tests += [
                test.test_socket.GeneralModuleTests('test_socket_fileno_requires_socket_fd'), # https://github.com/IronLanguages/ironpython3/pull/1638#discussion_r1059535867
            ]
```

slozier (approved):
Thanks!

PR #1706, opened 2023-07-04
Use Linq to compile getters

arontsang:
Use Linq for faster Getter accessors

### DIFF
--- a/Src/IronPython/Runtime/Binding/PythonGetMemberBinder.cs
+++ b/Src/IronPython/Runtime/Binding/PythonGetMemberBinder.cs
@@ -11,6 +11,7 @@ using System.Dynamic;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text;
+using System.Threading;
 
 using Microsoft.Scripting;
 using Microsoft.Scripting.Actions;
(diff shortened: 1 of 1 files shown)


slozier:
Thanks for the PR! Was there a specific property access you were trying to optimize? I'm curious as to what sort of speed improvements this can provide. Could probably further optimize the int/bool cases by avoiding the boxing unboxing.

arontsang:
Hi @slozier 

I'm starting to use IronPython at my company as a generic (user specified) rule engine.

During evaluation, I noticed that for a very simple rule, I was only able to get approximately 10million invocations per second.

Pulling up the profiler, I found that most of the time was spent in `Scope.GetVariable` (I have since added a PR for that) and `FastPropertyGet.GetPropertyInt`. 

Given that I am simply doing a Property.Get, I would have expected it should take less than 400ns per op and closer to the 3ns per op that native C# should provide.

Since running a noop func in IronPython was clocking in at about 12ns per op. I am wondering where the rest of the perf was going...

arontsang:
@slozier 

Also, I've reworked my PR with the following:

- Moved the cached Boxing into the Linq expression, thus removing an unnecessary box/unbox before entering the cached boxing method.
- Moved the type check into the Linq expression (thus removing an unnecessary casting check)
- Added a run counter for tiered compilation (compiles on the 8th access, so we don't slow down on once off property calls)

arontsang:
FYI the code I was optimising for was:

    def foo():
        bar.X + bar.Y

With the C#

    var scope = engine.CreateScope();
    
    engine.Execute(PythonCode, scope);
    var calc = scope.GetVariable<Func<int>>("foo");    

    foreach(var item in items)
    {
        scope.SetVariable("bar", item);
        var value = calc();
    }

We are using global variables because we intend to use Blockly to generate the Python code that is run in IronPython

arontsang:
Actually, come to think about it. The MethodInfo.Invoke path would always box the result.

There is no reason to call the DLR.Box method, as we don't save on allocation even with a cache hit.

slozier (approved):
Thanks for the PR! Looks good to me.

My only though is that some tests for the different branches would be useful. For example I haven't been able to come up with code uses `TryGetByJitWithoutCast` (I think TSelfType is pretty much always object so makes sense that it doesn't take that path). If you do have some example please let me know and I'll add them to the tests.

Anyway, will merge it as-is since I'm trying to get a release out the door.

PR #1709, opened 2023-07-13
Use OS-specific values for file open flags

BCSharp:
This is the second step towards `fcntl` and hopefully _IPython_/_Jupyter_ support one day. It defines the _O_-flags in `PythonNT` to have values consistent with the system headers on all OS-es IronPython targets. This change only affects the `posix` module, not the `nt` module. It also adds flags that are mandatory in Python for `posix`. (`fcntl` depends on flags from `posix` and `termios`).

The actual value of the flags is not relevant as long as they are only used with IronPython API, but once they are used with `fcntl` or `ioctl`, or if passed to 3-rd party extensions, they must be correct for the operating system used.

The added flags are not supported by `posix.open` (yet), nevertheless, they need to be present to support `fcntl`. I deliberately omitted optional flags for this reason; packages seem to test for the existence of the optional flags and enable some workarounds if the flags are not present. I'd rather see that than an attempt to use such flag that our `posix.open` simply ignores.

Also, in `posix`, Windows-specific flags are not available anymore. This is technically a breaking change but those flags should not have been available in `posix` anyway. The most notable is `O_TEMPORARY`, which actually was emulated correctly on Posix by `PythonNT` + `System.IO`. This does not fit well, however, with making `posix` more CPython-compatible.

The constants are generated by a _generate_* script, like the system error codes are. I have considered using constants from _Mono.Unix_, but it would have been considerably more clunky, slower and error-prone, since to avoid loading `Mono.Unix.dll` on Windows, the accessors would all have to be delegated to separate methods. Luckily, the constants are unlikely to change since any such change by an OS would break binary compatibility with many programs for that system.


slozier (approved):
Thanks! Looks good to me!

PR #1711, opened 2023-07-15
Fix file descriptor leak in FileManager

BCSharp:
Something that has caught my attention while reading the code trying to understand how `PythonFileManager` works.

### DIFF
--- a/Src/IronPython.Modules/nt.cs
+++ b/Src/IronPython.Modules/nt.cs
@@ -339,7 +339,7 @@ namespace IronPython.Modules {
             } else {
                 Stream stream = fileManager.GetStreamFromId(fd);
                 fileManager.RemoveObjectOnId(fd);
-                fileManager.DerefAndCloseLast(stream);
+                fileManager.DerefAndCloseIfLast(stream);
             }
         }
 
--- a/Src/IronPython.Modules/nt.cs
+++ b/Src/IronPython.Modules/nt.cs
@@ -332,16 +332,14 @@ namespace IronPython.Modules {
 #endif
 
         public static void close(CodeContext/*!*/ context, int fd) {
-            PythonContext pythonContext = context.LanguageContext;
-            PythonFileManager fileManager = pythonContext.FileManager;
-            if (fileManager.TryGetFileFromId(pythonContext, fd, out PythonIOModule.FileIO file)) {
-                fileManager.CloseIfLast(context, fd, file);
+            PythonFileManager fileManager = context.LanguageContext.FileManager;
+            if (fileManager.TryGetFileFromId(fd, out PythonIOModule.FileIO? file)) {
+                file.closefd = true;
+                file.close(context);
             } else {
-                Stream? stream = fileManager.GetObjectFromId(fd) as Stream;
-                if (stream == null) {
-                    throw PythonOps.OSError(9, "Bad file descriptor");
-                }
-                fileManager.CloseIfLast(fd, stream);
+                Stream stream = fileManager.GetStreamFromId(fd);
+                fileManager.RemoveObjectOnId(fd);
+                fileManager.DerefAndCloseLast(stream);
             }
         }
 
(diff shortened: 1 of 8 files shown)


slozier:
Hmm, this one is tricky. Been a while since I looked at it but there are probably some cases that we want to "leak"... For example but this would break:
```
>>> f = open(0)
>>> f.close()
>>> f = open(0)
OSError: [Errno 9] Bad file descriptor
```

BCSharp:
When I run your example as a script, CPython crashes at line 3. The REPL seems to behave differently. Perhaps the first three descriptors are close-protected on REPL; maybe we can implement something similar. When running a script, the first three descriptors are "closable", and opening a file after closing any of them will reuse the closed descriptor, making them behave like any other descriptor. (tested with CPython 3.11.4/Windows 11N).

I am not aware of any case where we would want to "leak"… If you have ideas for tests of some corner cases, feel free to commit them here. I am still working on this PR, since there are more cases when descriptors leak.

BCSharp:
The last change removes the leak AFAICS, but 
in order to keep the PR as small as possible, I kept an number of aspects unaddressed.

Here is a list of the outstanding issues:

* There is a race condition in `dup2`, when one thread does `dup2` and another one opens a file. I did not address it since it is a pre-existing issue and don't want to lump a solution to it with the proposed here changes to `PythonFileManager`. What I've seen in practice, `dup2` was always used with the standard descriptors, so I've maintained the workaround to never automatically assign those numbers to open files. This creates a small difference with CPython, but unlikely a significant one.

* The ability/inability to close standard descriptors (as demonstrated in the above example by @slozier).

* The limit of the open files in `PythonFileManager`: I have raised it from 0x1000 to 0x100000, which is the hard limit on Linux. The soft limit is 0x1000 but actually I hardly see a reason to have a limit here at all, since the OS will impose its own limit and `PythonFileManager` should not stand in its way. The only thing that comes to my mind is that if there is still a sort of "leak" here, there is a stopgap instead a runaway process. Anyway, this can be revisited when genuine file descriptors are implemented for Posix.

* Although there is no "leak" of `FileIO` anymore, those objects may not necessarily be always disposed, lingering around until GC calls their finalizer. This is not dramatic but not great either. Three possible approaches:
    * Add extra logic to dispose them when dereferenced (without closing the underlying streams, but just enough to suppress their finalizer).
    * Do not use `FileIO` objects to encapsulate `Streams` opened through `os.open` or `os.dup` (streams do not suffer from not-disposed object problem). I see two reasons why maybe not to do it:
        * Currently, `os.open` always opens one stream, that serves for both reading and writing; perhaps there is a combination of flags that will require opening two streams, in this case, `FileIO` is needed. Though, if such need arises, we can still fall back on `FileIO` but use `Stream` for the most cases. This will still be better than what we have now.
        * Using `FileIO` adds one extra bit of information that `Stream` does not have: `name`, which is the relative file name used to open the descriptor. I do not see any drawback of dropping it, since the name is relative, a stream already carries the absolute name, and maintain the name is not what CPython does. But maybe there is a reason for keeping it that I do not see.
    * Do nothing, since it is a matter of just a possible optimization.

* If the choice above is not to replace `FileIO` with `Stream`, we could adopt the same approach in `msvcrt` and drop the whole support of `Stream` in `PythonFileManager`. This will significantly simplify the code in a number of places, since now the code has to deal with two types of objects possibly associated with a descriptor.

* The calls on `PythonFileManager` to do a reverse lookup (from object to its descriptor) are pretty common, and not very efficient, since the whole hash table is being enumerated in a loop. This can be optimized by maintaining a reverse lookup table.

Since the usage and behaviour of `PythonFileManager` is rather tricky, I'd appreciate a thorough review.

Src/IronPython/Modules/_fileio.cs
```diff
@@ -273,17 +277,27 @@ public FileIO(CodeContext/*!*/ context, string name, string mode = "r", bool clo
...
+                            }
+                        } else {
+                            _readStream.Close();
+                        }
                     }
-                    if (_writeStream != null && !ReferenceEquals(_readStream, _writeStream)) {
+                    if (_writeStream is not null && !ReferenceEquals(_readStream, _writeStream) && readStreamClosed) {
                         _writeStream.Close();
-                        _writeStream.Dispose();
+                        myManager?.Remove(_readStream);
```
slozier, line 299: Should it be `_writeStream`?

Src/IronPython/Modules/_fileio.cs
```diff
@@ -294,10 +308,9 @@ public FileIO(CodeContext/*!*/ context, string name, string mode = "r", bool clo
                 }
             }
 
-            public bool closefd {
-                get {
-                    return _closefd;
-                }
+            internal bool closefd {
```
slozier, line 311: Should this be internal? Or only the setter?

Src/IronPython/Runtime/PythonFileManager.cs
```diff
@@ -2,92 +2,204 @@
...
+    /// <summary>
+    /// Emulates a file descriptor table for <see cref="Modules.PythonIOModule.FileIO"/> and <see cref="System.IO.Stream"/> objects.
+    /// </summary>
+    /// <remarks>
+    /// PythonFileManager emulates a file descriptor table. On Windows, .NET uses Win32 API which uses file handles
+    /// rather than file descriptors. The emulation is necesary to support Python API, which in some places uses file descriptors.
+    ///
+    /// The manager maintains a mapping between open files (or system file-like objects) and a "descriptor", being a small non-negative integer.
+    /// Unlike in CPython, the descriptors are allocated lazily, meaning they are allocated only when they become exposed (requested)
+    /// though relevant API calls. Therefore the ordering of assigned descriptors in IronPython may be different than in CPython.
```
slozier, line 28: typo: through
and on the next line: good

Src/IronPython/Runtime/PythonFileManager.cs
```diff
@@ -2,92 +2,204 @@
...
+    /// This should not be a problem since assumptions on descriptor ordering are not goot programming practice.
+    ///
+    /// Open filesystem files are represented in the manager as lower level Stream objects, and higher level FileIO objects.
+    /// FileIO basically encapsulates one ot two Streams (one for reading and one for writing, which may be the same).
```
slozier, line 32: typo: one of

Src/IronPython/Runtime/PythonFileManager.cs
```diff
@@ -2,92 +2,204 @@
...
+    /// There are two levels of sharing that are supported. On the higher level, several different FileIO objects may share the same
+    /// file descriptor. In such case, all of those FileIO objects maintain references to the same underlying Stream object(s)
+    /// and the manager maintains a mapping between the descriptor and only one of those FileIO objects (the first one opened).
+    /// In such situations, only one of the FileIO may be opened with flag `closefd` (CPython rule).
+    ///
+    /// If a filesystem file is opened without a corresponding FileIO (i.e. a bare file descriptor), the manager maintains a hidden FileIO
+    /// for itself to manage the mapping. This object is not exposed to Python and only used for opening/closing/dup etc. purposes.
+    ///
+    /// The second lever of sharing of open files is below the file descriptor level. A file descriptor can be duplicated using dup/dup2,
+    /// but the dupplicated descriptor is still refering to the same open file in the filesystem. In such case, the manager maintains
```
slozier, line 43: typo: duplicated

Src/IronPython/Runtime/PythonFileManager.cs
```diff
@@ -2,92 +2,204 @@
...
-            mapping.RemoveOnObject(o);
+        public int AddStream(Stream stream) => Add(stream);
+
+        // Public API controls which types are allowed to be added
+        private int Add(object obj) {
+            ContractUtils.RequiresNotNull(obj, nameof(obj));
+            lock (_synchObject) {
+                while (_objects.ContainsKey(_current)) {
+                    _current++;
+                    if (_current == LIMIT_OFILE)
```
slozier, line 95: `>=`?

Src/IronPython/Runtime/PythonFileManager.cs
```diff
@@ -2,92 +2,204 @@
...
 
-        public void CloseIfLast(int fd, Stream stream) {
-            mapping.RemoveOnId(fd);
-            if (-1 == mapping.GetIdFromObject(stream)) {
-                stream.Close();
+        public void AddRef(Stream stream) {
+            _refs.AddOrUpdate(stream, 1, (_,  v) => v + 1);
+        }
+
+        public bool DerefAndCloseLast(Stream stream) {
```
slozier, line 170: Maybe `DerefAndCloseIfLast`?

Src/IronPython.Modules/nt.cs
```diff
@@ -332,16 +332,14 @@ public static void chmod(CodeContext context, object? path, int mode, [ParamDict
...
 
         public static void close(CodeContext/*!*/ context, int fd) {
-            PythonContext pythonContext = context.LanguageContext;
-            PythonFileManager fileManager = pythonContext.FileManager;
-            if (fileManager.TryGetFileFromId(pythonContext, fd, out PythonIOModule.FileIO file)) {
-                fileManager.CloseIfLast(context, fd, file);
+            PythonFileManager fileManager = context.LanguageContext.FileManager;
+            if (fileManager.TryGetFileFromId(fd, out PythonIOModule.FileIO? file)) {
+                file.closefd = true;
+                file.close(context);
```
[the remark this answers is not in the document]
BCSharp: Indeed, it is not. AFAICS this is not a problem except perhaps in two cases:
1. Closing one of the standard descriptors — handling the standard descriptors is an issue in itself so I will look into it then.
2. Closing the target in `dup2` — there are a number of issues with `dup2`, which currently only handles basic cases. Best to look into it separately.

In the end, after those issues are addressed, this behaviour may change. We'll see.

Src/IronPython/Modules/_fileio.cs
BCSharp: Yes! Thanks for the catch.

Src/IronPython/Modules/_fileio.cs
BCSharp: This was an intentional change (by me). I see no reason why `closefd` should be public. It is not documented anywhere in Python API, and no versions of CPython I tried (ranging from 3.4 to 3.10) have this property visible.

slozier: Hmm, weird, I see it here. Am I looking at the wrong object?
```
>>> f = open("temp.txt")
>>> f.buffer.raw.closefd
True
```

Src/IronPython.Modules/nt.cs
```diff
@@ -355,65 +353,55 @@ public static void chmod(CodeContext context, object? path, int mode, [ParamDict
...
+                fileManager.EnsureRef(file._readStream);
+                fileManager.AddRef(file2._readStream);
+                return fd2;
             } else {
-                Stream? stream = pythonContext.FileManager.GetObjectFromId(fd) as Stream;
-                if (stream == null) {
-                    throw PythonOps.OSError(9, "Bad file descriptor");
-                }
-                return pythonContext.FileManager.AddToStrongMapping(stream);
+                var stream = (Stream)obj;
```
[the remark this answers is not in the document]
BCSharp: It would throw, but currently it is impossible to add to `FileManager` anything else than `FileIO` or `Stream` because only those types are allowed by public API on `PythonFileManager`. If someone tried to add support for any third type, there are a number of places that will need update.

I am not there yet, but with genuine file descriptors, I expect `PythonFileManager` to work quite differently.

Src/IronPython/Modules/_fileio.cs
BCSharp: Oh, it seems I was looking at the wrong object. Will make it public read-only again.

PR #1712, opened 2023-07-18
Improve handling of NUL file on Windows

BCSharp:
`NUL` file is now `stat`-able. The special handling of `/dev/null` is removed since this system file is supported by the OS and can be treated like any other file.


BCSharp:
> `StatStream` is kind of an odd pattern (or I'm just not used to seeing it in C#) but it's private so I guess it's fine.

With a public API I would have been more careful (I hope) and probably opted for something like
`bool TryStatStream(Stream stream, [NotNullWhen(true)] out stat_result? res)` but with local functions the habit is to write as few lines of code as possible 😃

PR #1713, opened 2023-07-18
Use absolute file name for stat

BCSharp:
Resolves problems when `stat` would fail if the current directory was changed after file opening.


slozier (approved):
Thanks! Looks good to me!

PR #1717, opened 2023-07-26
Enable PythonOps.PrintException on basic console builds

BCSharp:
This change makes IronPython 3 buildable without `FEATURE_FULL_CONSOLE` (except for `IronPythonConsole` and `IronPythonWindow`). Such a build behaves properly in LinqPad without a need for stdio redirecting (i.e. like IronPython 2).


slozier:
Thanks!

PR #1721, opened 2023-08-08
Use actual code context for enumerating

### LINKED ISSUE #1718: Memory leak when calling list(...) on OrderedDict
### Description

There is a memory leak when calling the list function with an OrderedDict. Even after the runtime has been shut down, objects of type PythonGetMemberBinder, PythonType, FunctionDefinition, FunctionCode etc. remain referenced and will not be garbage collected.

### Steps to Reproduce
(Repro project available here: https://github.com/spkl/IronPythonMemoryRepro)
1. In a .NET 6 program, create a ScriptEngine and ScriptScope.
2. Create a ScriptSource from a file containing the following code:
```python
import collections

def MyMethod(dummy):
    od = collections.OrderedDict()
    for i in range(100):
        od[i] = []
    
    for v in list(od.values()):
        continue

    return dummy
```
3. Execute the ScriptSource.
4. Call the `MyMethod` function.
5. Shut down the engine.
6. Repeat this 50 times.

**Expected behavior:** The memory footprint is stable. (For reference: After the first execution, it is ~50MB).

**Actual behavior:** The memory footprint exceeds 300 MB after 50 repetitions. Forcing a garbage collection does not change it significantly. (With a larger, more complex example, we have observed ~600 MB of additional, uncollectable memory after 30 script executions.)

**Workaround:** Instead of `for v in list(od.values()):`, use `for v in [x for x in od.values()]` or `for v in od.values():`.

### Versions

The issue is reproducible with both 3.4.0 and 3.4.1.

BCSharp:
Partially addresses #1718 (when the wrapping type is `list`).

### DIFF
--- a/Src/IronPython.Modules/IterTools.cs
+++ b/Src/IronPython.Modules/IterTools.cs
@@ -743,11 +743,11 @@ namespace IronPython.Modules {
 
         [PythonType]
         public class product : IterBase {
-            public product(params object[] iterables) {
-                InnerEnumerator = Yielder(ArrayUtils.ConvertAll(iterables, x => new PythonList(PythonOps.GetEnumerator(x))));
+            public product(CodeContext context, params object[] iterables) {
+                InnerEnumerator = Yielder(ArrayUtils.ConvertAll(iterables, x => new PythonList(context, PythonOps.GetEnumerator(x))));
             }
 
-            public product([ParamDictionary]IDictionary<object, object> paramDict, params object[] iterables) {
+            public product(CodeContext context, [ParamDictionary]IDictionary<object, object> paramDict, params object[] iterables) {
                 object repeat;
                 int iRepeat = 1;
                 if (paramDict.TryGetValue("repeat", out repeat)) {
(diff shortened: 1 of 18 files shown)


slozier:
Interesting... any ideas about the underlying issue? Seems like the leak comes back when we replace `list` with `tuple`. Do we need to try to eliminate all calls that use the default context?

BCSharp:
> Do we need to try to eliminate all calls that use the default context?

I think so. The underlying issue is that by using the default context with objects of an arbitrary Python type (like `OrderedDict`, not any builtin types or .NET types) pollutes the conversion call sites (and probably any call sites) with types that cease to exist after the runtime instance is shut down.

I think it may be OK to use the default context for operations on objects of guaranteed .NET types or builtins, but not when an arbitrary object may appear as argument.

There should be more places still when the leak comes back. E.g. I expect a leak when an `OderedDict` is splatted in a call with keyword arguments (untested, just my thought based on reading the code).

slozier (approved):
Sorry it took a while to review, just noticed it was no longer marked as draft...

Looks good to me.

PR #1725, opened 2023-08-16
Prevent closing of standard descriptors through file objects

BCSharp:
Now that I know where to look for `closefd`… This addresses https://github.com/IronLanguages/ironpython3/pull/1711#issuecomment-1636935329.

It is still not exactly like CPython, because it will prevent closing of the descriptors even when redirected but perhaps this is good enough. I cannot imagine when this would be desirable, but there isn't a valid use case. Anyway, note that a determined programmer may still force close a standard descriptor by `os.close(fd)`. This is likely not a good idea anyway since it will crash the CPython process on any attempt to use stdio, and those descriptors cannot be reopened.


slozier (approved):
Looks good to me. Thanks!

PR #1729, opened 2023-08-19
Redesign PythonFileManager

BCSharp:
This is the second pass of redesign of `PythonFileManager` and `FileIO`. There should be no functional changes (except for some small increase in compatibility with CPython). Basically, a part of `FileIO` has been factored out into `StreamBox`. This has several benefits:

* It simplifies the API of `PythonFileManager`, which now contains only one type of objects and the API becomes strongly typed. The callers do not need to wrestle with downcasting and having different code paths for different type object, which was not done consistently. Indeed, in some places, only one type of objects was handled.
* `StreamBox`, unlike `FileIO`, is not disposable and does not have a finalizer. This should reduce the pressure on the GC, esp. when files are opened using the `PyhonNT` API. 
* It sets the stage to finally implement `dup2` correctly, which is currently half-broken.

The handling and detection of TTY/Console streams remains tricky, not to a small extent because it is not always clear to me what the expected/desired/CPython behaviour is in all cases across all OSes, various console support levels, IO piping/redirecting, combined with reassigned (or not) IO in embedding scenarios. All I can say is that the current behaviour should not have changed much, and where it did, it's more in line with CPython's.

### DIFF
--- a/Src/IronPython/Modules/_fileio.cs
+++ b/Src/IronPython/Modules/_fileio.cs
@@ -38,33 +38,30 @@ namespace IronPython.Modules {
 
             private static readonly int DEFAULT_BUF_SIZE = 32;
 
-            internal Stream _readStream;
-            private Stream _writeStream;
+            private StreamBox _streams;
             private bool _closed, _closefd;
-            internal int _fd = -1;
             private WeakRefTracker _tracker;
             private PythonContext _context;
+
             public object name;
-            private ConsoleStreamType _consoleStreamType;
 
             internal FileIO(CodeContext/*!*/ context, Stream stream)
+                : this(context, new StreamBox(stream)) {
+            }
+
+            internal FileIO(CodeContext/*!*/ context, StreamBox streams)
                 : base(context) {
                 _context = context.LanguageContext;
+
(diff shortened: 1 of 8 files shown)


Src/IronPython/Modules/_fileio.cs
```diff
@@ -442,7 +353,7 @@ public FileIO(CodeContext/*!*/ context, string name, string mode = "r", bool clo
 
                 var span = pythonBuffer.AsSpan();
                 for (int i = 0; i < span.Length; i++) {
-                    int b = _readStream.ReadByte();
+                    int b = _streams.ReadStream.ReadByte();
```
slozier, line 356: Unrelated to the changes, but I sure hope nothing calls this code path. 😅

slozier (approved):
Thanks. Looks good to me!

Src/IronPython/Modules/_fileio.cs
BCSharp: I sure hope too! 😃 

I have some ideas for optimizing reading/writing, which I will submit in a separate PR.

PR #1730, opened 2023-08-19
Add option ConsoleSupportLevel

BCSharp:
This option is only available for embedding scenarios. There is no corresponding command switch since `ipy` always uses `Full` console support level.

I wonder how to make this option more discoverable.

### DIFF
--- a/Documentation/upgrading-from-ipy2.md
+++ b/Documentation/upgrading-from-ipy2.md
@@ -31,6 +31,55 @@ System.StringSplitOptions["None"]
 
 Similarly, `True` and `False` are also keywords in Python 3.
 
+## Redirecting output
+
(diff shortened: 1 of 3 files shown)


slozier (approved):
Looks good to me.

As for making it more discoverable, I don't have any great ideas. I wonder if a strongly typed overload to `Python.CreateEngine` with an options class would be helpful...

BCSharp:
> I wonder if a strongly typed overload to `Python.CreateEngine` with an options class would be helpful...

Perhaps it would be more helpful, but I don't see an easy way of implementing it. The current `Dictionary<string, object>` is being dragged through an number of layers in IronPython and the DLR, so the change would mean an awful lot of new overloads.

slozier:
I was thinking more along the lines of a single overload which we translate to a dictionary with the relevant options. Anyway, was just my first idea to make options more discoverable. I guess documentation would help as well. 🙂

Issue #1564, opened 2022-09-17
test_namebinding fails when run using ipy


slozier:
The repro for the test failure is as follows:
```py
import builtins
from importlib import reload
del builtins.pow
reload(builtins)
assert "pow" in dir(builtins)
pow(2,2)
```
Note that it also does not fail when run in the REPL so to reproduce the failure it should be saved to a file and executed.

I don't think it's quite the same as https://github.com/IronLanguages/ironpython3/issues/1028 since that one was about the change in behaviour between 3.4 and 3.5. In 3.5 the `assert` fails whereas with this bug the `assert` succeeds and the `pow` fails.

_Originally posted by @slozier in https://github.com/IronLanguages/ironpython3/issues/1563#issuecomment-1249955415_

PR #1732, opened 2023-08-22
Performance improvements of reading and writing through StreamBox

BCSharp:
This PR is intended to improve performance on .NET Framework and Mono. .NET (Core), with its `Span`-aware API does not have those performance issues. The main use case is `bytearray` as the buffer. AFAICS, this addition should cover all built-in types plus .NET types, except for `array.array`.

The unsafe `IPythoBuffer` helpers will do their best to extract the underlying CLI-array, to avoid a data copy or the wicked loop spin (see https://github.com/IronLanguages/ironpython3/pull/1729#discussion_r1299501437). This code path may perhaps be not the most popular one, but I thought I saw "opportunities" like this in other places too. At least this will serve as a sample solution.

### DIFF
--- a/Src/IronPython/Runtime/PythonFileManager.cs
+++ b/Src/IronPython/Runtime/PythonFileManager.cs
@@ -109,17 +109,41 @@ namespace IronPython.Runtime {
             return buffer;
         }
 
-        public int Write(ReadOnlySpan<byte> bytes) {
+        public int ReadInto(IPythonBuffer buffer) {
 #if NETCOREAPP
+            return _readStream.Read(buffer.AsSpan());
+#else
+            byte[]? bytes = buffer.AsUnsafeWritableArray();
+            if (bytes is not null) {
+                return _readStream.Read(bytes, 0, buffer.NumBytes());
+            }
+
+            var span = buffer.AsSpan();
+            for (int i = 0; i < span.Length; i++) {
+                int b = _readStream.ReadByte();
+                if (b == -1) return i;
+                span[i] = (byte)b;
+            }
+            return span.Length;
+#endif
+        }
+
+        public int Write(IPythonBuffer buffer) {
+            int count;
+#if NETCOREAPP
+            ReadOnlySpan<byte> bytes = buffer.AsReadOnlySpan();
+            count = bytes.Length;
             _writeStream.Write(bytes);
 #else
-            _writeStream.Write(bytes.ToArray(), 0, bytes.Length);
+            byte[] bytes = buffer.AsUnsafeArray() ?? buffer.AsUnsafeWritableArray() ?? buffer.ToArray();
+            count = buffer.NumBytes();
+            _writeStream.Write(bytes, 0, count);
 #endif
             _writeStream.Flush(); // IO at this level is not supposed to buffer so we need to call Flush.
             if (!IsSingleStream) {
                 _readStream.Seek(_writeStream.Position, SeekOrigin.Begin);
             }
-            return bytes.Length;
+            return count;
         }
 
         public void Flush() {
--- a/Src/IronPython/Runtime/PythonFileManager.cs
+++ b/Src/IronPython/Runtime/PythonFileManager.cs
@@ -118,11 +118,14 @@ namespace IronPython.Runtime {
                 return _readStream.Read(bytes, 0, buffer.NumBytes());
             }
 
+            const int chunkSize = 0x400; // 1 KiB
+            bytes = new byte[chunkSize];
             var span = buffer.AsSpan();
-            for (int i = 0; i < span.Length; i++) {
-                int b = _readStream.ReadByte();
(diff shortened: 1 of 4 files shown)


BCSharp:
I have no idea why the Mono test on macOS on GitHub CI fails. I have noticed though that Azure uses "Test Execution Command Line Tool Version 17.3.3 (x64)", while GitHub uses "Test Execution Command Line Tool Version 17.7.0-preview-23364-03+bc17bb9693cfc4778ded51aa0ab7f1065433f989 (x64)". Is it intentional?

Also, running Mono tests locally on 6.2.0.200 (CI is using 6.2.0.188) in **debug** mode fails on three tests: `test_ctypes`, `test_memoryio_stdlib`, `test_namebinding` (the last one is different than #1564).  They all pass in release mode and they all pass on .NET in both debug and release modes.

@slozier, do those tests pass on your local machine?

Src/IronPython/Runtime/PythonFileManager.cs
```diff
@@ -109,17 +109,41 @@ internal sealed class StreamBox {
...
+        public int ReadInto(IPythonBuffer buffer) {
 #if NETCOREAPP
+            return _readStream.Read(buffer.AsSpan());
+#else
+            byte[]? bytes = buffer.AsUnsafeWritableArray();
+            if (bytes is not null) {
+                return _readStream.Read(bytes, 0, buffer.NumBytes());
+            }
+
+            var span = buffer.AsSpan();
```
slozier, line 121: If we're moving this code into a helper that might end up being called elsewhere it's probably worth optimizing it a bit.

slozier (commented):
Also not sure about the macOS failure during CI. Have not had the time to investigate. Different execution commands are not intentional, could be the different agents that have different stuff installed.

Looks like they're failing for me in debug mode as well. I'm on Mono 6.12.0.182.

slozier:
@BCSharp Looks like all the debug failures are GC related issues. If you do something like set the object to `None` before the delete then the tests start passing. For example in `test_memoryio_stdlib`:
```py
        buf = None # add this line to make the test pass in Mono debug
        del buf
        support.gc_collect()
        memio.truncate()
```

BCSharp:
Strange that it only happens in debug builds. Perhaps debug GC hangs on to some objects past their lifetime for some performance tracking.

Src/IronPython/Runtime/PythonFileManager.cs
slozier: Thanks, works for me! Another option would have been to go with the base `Stream.Read(Span<byte>)` implementation:
```cs
        private static int Read(this Stream stream, Span<byte> buffer) {
            byte[] sharedBuffer = ArrayPool<byte>.Shared.Rent(buffer.Length);
            try {
                int numRead = stream.Read(sharedBuffer, 0, buffer.Length);
                if ((uint)numRead > (uint)buffer.Length) {
                    throw new IOException();
                }

                new ReadOnlySpan<byte>(sharedBuffer, 0, numRead).CopyTo(buffer);
                return numRead;
            } finally {
                ArrayPool<byte>.Shared.Return(sharedBuffer);
            }
        }
```

slozier (approved):

BCSharp:
@slozier I have managed to get one of the tests to fail in the release mode on Mono, running on a battery-powered laptop, so the system was slowed down. I think now that it may be a race condition between `gc.collect()` and some cleanup `del` does in the background and is not finished when `gc.collect()` is called. If this is true, it would mean it is not Mono-specific, it's just that other frameworks are fast enough on the hardware we are using.

For instance, adding a delay after `del` fixes all problem cases. To use your example:

```python
        del buf
        timer.sleep(1) # add this line to make the test pass in Mono
        support.gc_collect()
        memio.truncate()
```

slozier:
I ran a few experiments and I'm not sure about the race condition, I came up with a "minimal" example and my suggestion of using `buf = None` doesn't seem to work. However just calling a random builtin with one argument (in your case `time.sleep(1)`) seems to make the issue go away? Perhaps the reference it being held on to by something on the interpreter stack...
```py
import gc
import _pyio
import time

def test_getbuffer():
    memio = _pyio.BytesIO(b"1234567890")
    buf = memio.getbuffer()
    del buf
    bin(0) # call any builtin with 1 argument?
    gc_collect()
    memio.truncate()
    memio.close()

def gc_collect():
    gc.collect()
    gc.collect()
    gc.collect()

test_getbuffer()
```

BCSharp:
OK, some more results from my side. It doesn't have to be a builtin call. The following is enough to get the test passing in the release mode:
```python
    del buf
    i = 1
    i -= 1
    gc_collect()
```

To get it passing in debug mode it is insufficient, but something slightly more complicated will do:
```python
    del buf
    i = 1
    if True:
        i -= 1
    gc_collect()
```

PR #1757, opened 2023-12-11
Resolve some warnings about string comparisons

BCSharp:


Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -2079,7 +2079,7 @@ internal static partial class CodecsInfo {
 #if DEBUG
                 foreach (KeyValuePair<string, Lazy<Encoding?>> kvp in d) {
                     // all codecs should be stored in lowercase because we only look up from lowercase strings
-                    Debug.Assert(kvp.Key.ToLower(CultureInfo.InvariantCulture) == kvp.Key);
+                    Debug.Assert(kvp.Key.Equals(kvp.Key, StringComparison.OrdinalIgnoreCase));
```
slozier, line 2082: Oh hmm, guess I was a bit quick on the merge here. We want the `ToLower` conversion since we're asserting that all the keys are lowercase, no?

BCSharp: Yes, you are right. I guess I was a bit quick on the commit here.

Issue #1785, opened 2024-03-27
The ipyc /out: option dose not change the output path of exe host


wherewhere:
### Prerequisites

ipyc /out: option dose not change the output path of exe host.

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description

When I set the output to `bin\Program` with `ipyc /out:bin/Program /main:Program.py`, the `Program.exe` still output at the path which `Program.py` at. And the host get the `Program.dll` at `bin/Program.dll`, as I can not just copy the exe to `bin`. If I set the output path to an absolute path, I even can not move the program to another place...
```cs
public class PythonMain
{
    public static int Main()
    {
        try
        {
            string currentDirectory = Environment.CurrentDirectory;
            Environment.CurrentDirectory = new FileInfo(Assembly.GetEntryAssembly().Location).Directory.FullName;
            string fullPath = Path.GetFullPath("bin/Program.dll");
            Environment.CurrentDirectory = currentDirectory;
            return PythonOps.InitializeModuleEx(Assembly.LoadFile(fullPath), "__main__", null, ignoreEnvVars: false, null);
        }
        catch (Exception ex)
        {
            string currentDirectory = ex.Message;
            Console.WriteLine("Error occurred: {0}", currentDirectory);
            return -1;
        }
    }
}
```

So why not combined the host and dll just like JScript?
```cs
public class JScript Main
{
    [STAThread(/*Could not decode attribute arguments.*/)]
    public static void Main(string[] P_0)
    {
        GlobalScope globalScope = VsaEngine.CreateEngineAndGetGlobalScope(fast: true, new string[5] { "mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089", "System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089", "System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a", "System.Runtime.InteropServices, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a", "System.Reflection, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" });
        JScript 0 obj = new JScript 0(globalScope);
        globalScope.engine.PushScriptObject(obj);
        obj.Global Code();
        new JScript 1(globalScope).Global Code();
[...]

PR #1786, opened 2024-03-27
Let IronPython Compiler only create one exe file when target to exe or winexe

wherewhere:
The issue is #1785
I add a new method `CreateAssemblyGen` in `ClrModule` to get `AssemblyGen`, so that I can add exe host code into ScriptCode assembly. Then I change the init code from
```cs
string currentDirectory = Environment.CurrentDirectory;
Environment.CurrentDirectory = new FileInfo(Assembly.GetEntryAssembly().Location).Directory.FullName;
string fullPath = Path.GetFullPath("bin/Program.dll");
Environment.CurrentDirectory = currentDirectory;
return PythonOps.InitializeModuleEx(Assembly.LoadFile(fullPath), "__main__", null, ignoreEnvVars: false, null);
```
to
```cs
return PythonOps.InitializeModuleEx(Assembly.GetExecutingAssembly(), "__main__", null, ignoreEnvVars: false, null);
```
since the ScriptCode assembly is itself.
But now it have some new problem. Since the `AssemblyBuilder` dose not have `DefineIconResource` method, I have no ideal to add win32 icon. And the Windows Defender said it is viruses when `/standalone` option is on...

### DIFF
--- a/Src/IronPython/Runtime/ClrModule.cs
+++ b/Src/IronPython/Runtime/ClrModule.cs
@@ -15,6 +15,10 @@ using ComTypeLibInfo = Microsoft.Scripting.ComInterop.ComTypeLibInfo;
 using ComTypeLibDesc = Microsoft.Scripting.ComInterop.ComTypeLibDesc;
 #endif
 
+#if FEATURE_FILESYSTEM && FEATURE_REFEMIT
+using System.Reflection.Emit;
+#endif
+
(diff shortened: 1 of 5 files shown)


wherewhere:
Looks fine with .NET Core App
image

wherewhere:
The Win32 icon setting is working now.

PR #1810, opened 2024-11-11
Update Parser.cs

m1090c:
In the DebugInfoRewriter class within Microsoft.Dynamic.dll, the VisitParameter method adds debugging information starting from the second Expression. This causes the first line of code to lack debugging information, making it undebuggable.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -1,0 +1,59 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest-large]
+
+    steps:
+    - name: Install tools
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt-get -yq install mono-vbnc dos2unix
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+    - name: Setup .NET Core 3.1
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: '3.1.x'
+    - name: Setup .NET 6.0
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: '6.0.x'
+    - name: Setup .NET 8.0
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: '8.0.x'
+    - name: Version Information
+      run: |
+        dotnet --info
+        try { msbuild -version } catch { }
+        try { mono --version } catch { }
+      shell: pwsh
+    - name: Build
+      run: pwsh make.ps1
+    - name: Package
+      run: pwsh make.ps1 package
+    - uses: actions/upload-artifact@v4
+      with:
+        name: packages-${{ matrix.os }}
+        path: Package/Release/Packages
+    - name: Test (net462)
+      run: ./make.ps1 -frameworks net462 test-all
+      shell: pwsh
+    - name: Test (netcoreapp3.1)
+      run: ./make.ps1 -frameworks netcoreapp3.1 test-all
+      shell: pwsh
+    - name: Test (net6.0)
+      run: ./make.ps1 -frameworks net6.0 test-all
+      shell: pwsh
+    - name: Test (net8.0)
+      run: ./make.ps1 -frameworks net8.0 test-all
+      shell: pwsh
--- a/Src/IronPython/Compiler/Parser.cs
+++ b/Src/IronPython/Compiler/Parser.cs
@@ -3095,6 +3095,7 @@ namespace IronPython.Compiler {
                 l.Add(s);
             }
 
+            l.Insert(0, new EmptyStatement());
             Statement[] stmts = l.ToArray();
 
             if (returnValue && stmts.Length > 0) {


slozier:
Thanks the PR. Unfortunately this appears to break all tests. Could you maybe provide a code example of what you're trying to fix?

m1090c:
Using the SetTrace method in IronPython.Hosting.Python to set a traceback.
Class:IronPython.Hosting.Python
public delegate TracebackDelegate TracebackDelegate(TraceBackFrame frame, string result, object payload);
/// <summary>
/// Enables call tracing for the current thread in this ScriptEngine.  
/// 
/// TracebackDelegate will be called back for each function entry, exit, exception, and line change.
/// </summary>
public static void SetTrace(this ScriptEngine/*!*/ engine, TracebackDelegate traceFunc) {
    SysModule.settrace(GetPythonContext(engine).SharedContext, traceFunc);
}

When executing simple assignment statements, such as:
x = 1  
y = 2  
in the result output, only the y = 2 statement is marked with TraceEventKind.TracePoint. The debug marker for x = 1 is skipped.

Class:IronPython.Runtime.PythonTracebackListener
private void DispatchTrace(List<FunctionStack> thread, Debugging.TraceEventKind kind, object payload, TracebackDelegate traceDispatch, object traceDispatchObject, TraceBackFrame pyFrame) {
    object args = null;

    // Prepare the event
    string traceEvent = String.Empty;
    switch (kind) {
        case Debugging.TraceEventKind.FrameEnter: traceEvent = "call"; break;
        case Debugging.TraceEventKind.TracePoint: traceEvent = "line"; break;
        case Debugging.TraceEventKind.Exception:
            traceEvent = "exception";
            object pyException = PythonExceptions.ToPython((Exception)payload);
            object pyType = ((IPythonObject)pyException).PythonType;
            args = PythonTuple.MakeTuple(pyType, pyException, new TraceBack(null, pyFrame));
            break;
        case Debugging.TraceEventKind.FrameExit:
            traceEvent = "return";
            args = payload;
            break;
    }

In reviewing the debug information generation method in DebugInfoRewriter from Microsoft.Dynamic, the int locationCookie = this._locationCookie; starts from 0. To address this, I added an EmptyStatement to handle the issue, but this might only be a temporary solution. Perhaps you have a better approach to resolve it.

protected override Expression VisitDebugInfo(DebugInfoExpression node)
		{
			if (node.IsClear)
			{
				return Expression.Empty();
			}
			if (node.Document == null)
			{
				throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, ErrorStrings.DebugInfoWithoutSymbolDocumentInfo, new object[] { this._locationCookie }));
			}
			DebugSourceFile debugSourceFile = this._debugContext.GetDebugSourceFile(string.IsNullOrEmpty(node.Document.FileName) ? "<compile>" : node.Document.FileName);
			int locationCookie = this._locationCookie;
			this._locationCookie = locationCookie + 1;
			int num = locationCookie;
			Expression expression2;
			if (!this._transformToGenerator)
			{
				Expression expression;
				if (num == 0)
				{
					expression = Expression.Empty();
				}
				else
				{
					expression = Expression.Call(typeof(RuntimeOps).GetMethod("OnTraceEvent"), this._thread, Utils.Constant(num), Expression.Convert(Expression.Constant(null), typeof(Exception)));
				}
				expression2 = Expression.Block(Expression.Assign(this._debugMarker, Utils.Constant(num)), Expression.IfThen(Expression.GreaterThan(Expression.Property(this._sourceFilesToVariablesMap[debugSourceFile], "Mode"), Expression.Constant(1)), Expression.IfThen(Expression.OrElse(Expression.Equal(Expression.Property(this._sourceFilesToVariablesMap[debugSourceFile], "Mode"), Expression.Constant(3)), Expression.ArrayIndex(this._traceLocations, Utils.Constant(num))), Expression.Block(this._pushFrame ?? Expression.Empty(), expression))));
			}
			else
			{
				expression2 = Expression.Block(new Expression[] { Utils.YieldReturn(this._generatorLabelTarget, this._debugYieldValue, num) });
				if (this._currentLocals.Count > 0)
				{
					BlockExpression blockExpression = this._currentLocals.Peek();
					IList<VariableInfo> list;
					if (!this._variableScopeMapCache.TryGetValue(blockExpression, out list))
					{
						list = new List<VariableInfo>();
						BlockExpression[] array = this._currentLocals.ToArray();
						for (int i = array.Length - 1; i >= 0; i--)
						{
							foreach (ParameterExpression parameterExpression in array[i].Variables)
							{
								list.Add(this._localsToVarInfos[parameterExpression]);
							}
						}
						this._variableScopeMapCache.Add(blockExpression, list);
					}
					this._variableScopeMap.Add(num, list);
				}
				DebugSourceSpan debugSourceSpan = new DebugSourceSpan(debugSourceFile, node.StartLine, node.StartColumn, node.EndLine, node.EndColumn);
				this._markerLocationMap.Add(num, debugSourceSpan);
			}
			return expression2;
		}

m1090c:
@slozier slozier

slozier:
Can reproduce the issue with the following:
```py
import sys

calls = []

def trace_func(frame, event, arg):
    calls.append(f"{event} {frame.f_lineno}")
    return trace_func

sys.settrace(trace_func)
exec('x=1\ny=2')
sys.settrace(None)

assert calls == ['call 1', 'line 1', 'line 2', 'return 2']
```

I wonder if we can just drop the branch in `VisitDebugInfo`...
```cs
if (num == 0)
{
	expression = Expression.Empty();
}
```

m1090c:
The issue doesn't occur if the first line of the code contains an importstatement. The issue only arises when the first line is an AssignmentStatement.This issue doesn't occur in Python either.

PR #1812, opened 2024-11-17
GitHub: Simplify root directory and migrate documentation

Lamparter:
I'm an enthusiastic user of IronPython and have been for years... it's sad to see the last commit to main being nearly six months ago!
In this PR I've taken the effort to simplify the root directory and make some other few contributions in an effort to hopefully revive the work on this project to that of a big, ambitious project that IronPython is.


Lamparter:
@dotnet-policy-service agree

Lamparter:
@slozier would you mind reviewing this PR when you get a moment?

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,39 +0,0 @@
...
-
-**Expected behavior:**
-
-[What you expected to happen]
-
-**Actual behavior:**
-
-[What actually happened]
-
-### Version Information
```
slozier, line 30: Hmm, not sure what the issue templates changes will look like in practice, but we want the information from the Version Information section.

.github/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+More examples and documentation on how to use IronPython outside the interactive console can be found [here](https://github.com/IronLanguages/ironpython3/wiki/Using-IronPython).
+
+## 🐍 State of the Project
+
+The current target is Python 3.4, although features and behaviors from later versions may be included.
+
+See the following lists for features from each version of CPython that have been implemented:
+
+- [`Python3.0`](https://github.com/IronLanguages/ironpython3/wiki/WhatsNew%E2%80%90Python3.0)
```
slozier, line 73: Bit weird that these are using the unicode hyphen instead of the good old ASCII dash. :smile:

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,72 @@
+name: Bug Report
```
slozier, line 1: Although I like the idea of better structured bug reports I'm not sure this set of templates is well suited for the project. I assume this lets you pick a style before proceeding to filing an issue? Is there an escape hatch that allows not using a template?

slozier (commented):
Thanks for the PR! Seeing community PRs definitely helps motivate me to work on the project.

I'm not sure how I feel about using the wiki for documentation. The biggest drawbacks for me would be:
- The content is not visible in the cloned repo so I tend to neglect it. For example if I want to drop support for .NET 6 (which is no longer supported by Microsoft) I usually search though files to find relevant things to change.
- Can't keep different branches of documentation, for example `main` and `3.6` could have different `Package Compatibility` or `What's New` pages. Although it looks like in practice I haven't done that.

@BCSharp I know you played with the documentation a bit. Any thoughts or concerns?

.github/ISSUE_TEMPLATE/bug_report.yml
Lamparter: There is a way to still use blank issues, it can be enabled from the `config.yml` file I created. I'll enable it now.

.github/README.md
Lamparter: @slozier the ASCII dash symbol doesn't apply the bullet point in Markdown like the Unicode one does:

– ASCII dash
- Unicode dash

If you insist I'll add it though 🙂

slozier: Sorry I wasn't clear. I meant in the filename `WhatsNew%E2%80%90Python3.0`.

.github/ISSUE_TEMPLATE.md
Lamparter: I have similar issue templates in a few of my repos, [here](https://github.com/RiversideValley/Graphite/issues/new/choose)'s a repo that you can see what they look like in practice. Feel free to make any issues on that repo to try them out, I don't mind.

Lamparter: As for the version information, I'll had that soon.

.github/README.md
Lamparter: Oh! Right, I'll change that.

slozier (commented):
Seems generally fine. Will probably need to adjust the issue templates some more but it's a good start.

Looks like renaming LICENSE might have broken the packaging step.

Lamparter (commented):
The NuGet package built-in README will need to be changed as well, since NuGet isn't using the latest of MarkDig and doesn't support more recent GitHub Markdown like info bars.

slozier:
> The NuGet package built-in README will need to be changed as well, since NuGet isn't using the latest of MarkDig and doesn't support more recent GitHub Markdown like info bars.

What's wrong with the NuGet package README? I guess some of the URLs will have to change, but I don't think it needs substantial changes.

BCSharp:
@Lamparter, I am enthusiastic to see an enthusiastic contributor!

> I'm not sure how I feel about using the wiki for documentation. The biggest drawbacks for me would be:
> 
> * The content is not visible in the cloned repo so I tend to neglect it. For example if I want to drop support for .NET 6 (which is no longer supported by Microsoft) I usually search though files to find relevant things to change.
> 
> * Can't keep different branches of documentation, for example `main` and `3.6` could have different `Package Compatibility` or `What's New` pages. Although it looks like in practice I haven't done that.
> 
> 
> @BCSharp I know you played with the documentation a bit. Any thoughts or concerns?

I am very positive about moving the documentation to the GitHub wiki. Actually, I was planning to do it myself at some point, but never got to finish it. Over the years, I would clean up the old wiki (moving any useful info to the checked in docs), hoping for one day to put it to good use. Writing `.md` docs was an interim solution. It was easy for me since I have write access to the repo, but a regular would-be contributor had to go through the whole PR-submit process, just to fix a typo. Yes, some did it anyway, but I feel we would have gotten bigger engagement with the wiki. I think that even I would have been more inclined to make small fixes more often.

However, @slozier, I share your concerns about the move; here is how I see them:

* I, too, use search through files to make sure that some changes are properly accounted for. I usually use VSCode for this purpose, its search/replace seems to offer much more pleasant experience than the heavyweight VS (I prefer the latter for debugging though). Luckily, the wiki is a git repo too, so in VSCode it is simple to create a workspace spanning two cloned git projects, checked out side by side. Any global search scans then both repos, as if they were one. Also, VS is not available on Linux, and (since recently) not on macOS, so expect to use VSCode more and more often. I don't know your preferred toolset, but if you do not use VSCode, you may want to give it a try. I also think (though never tried so might be wrong) that it is possible to configure VS to have a solution that included the wiki pages next to the regular IronPython code.
* About "Package Compatibility": I'd expect that in general, the wiki documentation pertains to the current main, which currently is 3.4, and when there is some discrepancy between 3.4 and 3.6, it is explicitly tagged on the page. For package compatibility, it could even be separate wiki pages. For maintainability, version-specific comments could be tagged in some standard way so again, the search can pick them up. In code I use string `Python 3.6` (or any other version that the code implements), but I search on `3.6`. Another point: having "Package Compatibility" in the wiki will invite reports from the field about packages that work that we may have never tried ourselves.
* That said, some documents do not tend themselves to be included in the wiki. Of the top of my head it is the feature sets and the "What's New" series. Especially in the case of "What's New", changes are version-controlled together with the corresponding code changes, so it is (relatively) easy to find out how a specific piece of functionality was implemented and when. Even if there is some discrepancy in commits, that is, a feature had been completed and only later the checkmark added to "What's New" (we are not perfect, are we 😃), the PR with the commit with the change to "What's New" would often refer to the other PR that implemented the feature (at least, that's how I have done that), so there is an audit trail. Similarly, with an earlier checkout one can find the corresponding state on feature implementation by simply looking at the "What's New" documents. Further, changes to "What's New" documentation do not benefit from the lightweight approach of the wiki; they should be updated only after careful consideration.

Lamparter:
> What's wrong with the NuGet package README? I guess some of the URLs will have to change, but I don't think it needs substantial changes.

It will break on NuGet

Lamparter:
I was also working on a little secret something, perhaps that might be useful here 😁 

the IronPython website

That is, of course, a new website!

slozier:
Alright, what's the plan with this one? I'm fine with moving docs to the wiki as well as the github issue templates. I'd like to keep the "what's new" files in the repo as I consider them more like developer checklists than documentation.

Lamparter:
Where do you want them to go? The Documentation folder?

Lamparter:
I think `.github` might need to be the place to put this if you don't want them in the wiki.
Though perhaps it might be good to add it as `src\spec` or `src\todo`?

slozier:
> @slozier , are you OK with _Features_ out of the repo?

The feature symbols you mean? They're more of a dev reference and not something a user would care about. Would fit in with `src\spec` I guess.

Lamparter:
They might be easier to edit and reference if they were in the wiki.

BCSharp:
@Lamparter Are you planning on finishing this PR?

Lamparter:
Yes

ferdnyc:
@slozier 

> I'm not sure how I feel about using the wiki for documentation. The biggest drawbacks for me would be:
> 
> * The content is not visible in the cloned repo so I tend to neglect it. For example if I want to drop support for .NET 6 (which is no longer supported by Microsoft) I usually search though files to find relevant things to change.

Just FYI, it's possible to check out the Wiki content using Git (there's a URL at the bottom of the Wiki sidebar; it's https://github.com/IronLanguages/ironpython3.wiki.git for this repo's wiki).

It's even theoretically possible to add that as a Git Submodule in the current repo, so that a `git clone --recursive` will pull down the Wiki content along with the main repo. However, all of the usual caveats and considerations for dealing with submodules apply, so there may be problems keeping up to date with wiki changes and ensuring they're incorporated into the cloned submodule.

But there's certainly nothing to stop you from doing this inside your cloned  `ironpython3`, to  get a purely-local copy of the Wiki source:

```console
$ cd \path\to\cloned\ironpython3
$ git clone https://github.com/IronLanguages/ironpython3.wiki.git wiki
$ # [add /wiki to .git/info/exclude to prevent accidental commits]
```

Then you just have to remember to `git pull` inside the `wiki` directory to update its contents before running any recursive searches.

Other issues, like not being able to keep separate branches of the content, still apply.

PR #1820, opened 2024-11-29
Get rid of ArrayUtils.EmptyObjects

slozier:
Gets rid of uses of `ArrayUtils.EmptyObjects` and `ArrayUtils.EmptyStrings`. Uses the C# 12 empty collection expression `[]` as a replacement. Ultimate goal is to try and remove dependencies on DLR utils (like `ArrayUtils`) so they can be obsoleted (less code to worry about).

### DIFF
--- a/Src/IronPython.Modules/_thread.cs
+++ b/Src/IronPython.Modules/_thread.cs
@@ -228,9 +228,9 @@ namespace IronPython.Modules {
                 try {
 #pragma warning disable 618 // TODO: obsolete
                     if (_kwargs != null) {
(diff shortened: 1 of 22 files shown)


BCSharp (approved):
Is this all or is there more to come?

slozier:
> Is this all or is there more to come?

It's all for now. I looked at more of the `ArrayUtils` usage an there didn't seem to be clear cut replacements (unless there are some array manipulation APIs I'm missing in the core libraries). Thinking that making an internal copy and flagging the DLR class as obsolete might be a good way to proceed...

Lamparter:
Glad to see more activity now 😄
When the "infrastructure tsunami" as @BCSharp calls it gets pushed through there'll be far more activity from my side 🚀

PR #1821, opened 2024-11-29
Implement equality comparisons on System.Array

BCSharp:
Fixes `IndexOutOfRangeException` when equating arrays with different index bases (lower bounds).

Example:

```pycon
>>> from System import *
>>> a = Array.CreateInstance(Int32, (5,), (1,))
>>> b = Array.CreateInstance(Int32, (5,), (1,))
>>> a == b
IndexError: Index was outside the bounds of the array.
>>> a != b
IndexError: Index was outside the bounds of the array.
```

This is due to a bug in .NET:

```c#
var a = Array.CreateInstance(typeof(int), new[] { 5 }, new[] { 1 });
var b = Array.CreateInstance(typeof(int), new[] { 5 }, new[] { 1 });
((IStructuralEquatable)a).Equals(b, EqualityComparer<int>.Default);  // IndexOutOfRangeException
```

I'll see if this is already reported. It just shows that nonzero bases are not very popular.

In this PR I'll also cover equating multidimensional arrays, since there is just a little bit to add to address that as well.

### DIFF
--- a/Src/IronPython/Runtime/Operations/ArrayOps.cs
+++ b/Src/IronPython/Runtime/Operations/ArrayOps.cs
@@ -93,10 +93,38 @@ namespace IronPython.Runtime.Operations {
             if (other is null) throw PythonOps.TypeError("expected Array, got None");
 
             if (self.GetType() != other.GetType()) return ScriptingRuntimeHelpers.False;
+            // same type implies: same rank, same element type
+            for (int d = 0; d < self.Rank; d++) {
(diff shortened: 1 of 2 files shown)


BCSharp:
Ordering comparisons are also not working well, but I leave it as that — I don't have old code to submit.

slozier (approved):
Thanks, seems fine to me. I'll file an issue regarding array comparisons not working since this seems to be a regression from 2.7.

PR #1827, opened 2024-12-04
Correct error raised when slicing CLI arrays

BCSharp:
This aligns the error raised (exception thrown) with `memoryview`.


slozier (approved):
Looks good to me!

PR #1828, opened 2024-12-04
Implement System.Array constructor with base arg

BCSharp:
Generally, the `repr` string in Python can be evaluated by `eval`, providing a round-trip-complete string representation of an object. There are exceptions, like deeply nested or circular structures, or nondescript "object type at memory address".

For CLI arrays, representations of 1D arrays are descriptive, while higher dim are not. This PR adds a constructor that is able to handle the representation for 1D arrays with a non-zero base. I'd make the extra parameter keyword-only, but the DLR doesn't support it yet.

### DIFF
--- a/Tests/test_array.py
+++ b/Tests/test_array.py
@@ -198,6 +198,26 @@ class ArrayTest(IronPythonTestCase):
             for y in range(array3.GetLength(1)):
                 self.assertEqual(array3[x, y], 0)
 
+    def test_constructor_nonzero_lowerbound(self):
+        # 1-based
+        arr = System.Array[int]((1, 2), base=1)
+        self.assertEqual(arr.Rank, 1)
+        self.assertEqual(arr.Length, 2)
+        self.assertEqual(arr.GetLowerBound(0), 1)
+        self.assertEqual(arr.GetUpperBound(0), 2)
+        self.assertEqual(arr[1], 1)
+        self.assertEqual(arr[2], 2)
+        for i in range(1, 3):
+            self.assertEqual(arr[i], i)
+
+    def test_repr(self):
+        from System import Array
+        arr = Array[int]((5, 1), base=1)
+        s = repr(arr)
+        self.assertEqual(s, "Array[int]((5, 1), base=1)")
+        array4eval = eval(s, globals(), locals())
+        self.assertEqual(arr, array4eval)
+
     def test_nonzero_lowerbound(self):
         a = System.Array.CreateInstance(int, (5,), (5,))
         for i in range(5, 5 + a.Length): a[i] = i
@@ -208,7 +228,7 @@ class ArrayTest(IronPythonTestCase):
         self.assertEqual(a[-1:-3:-1], System.Array[int]((9,8)))
         self.assertEqual(a[-1], 9)
 
-        self.assertEqual(repr(a), 'Array[int]((5, 6, 7, 8, 9), base: 5)')
+        self.assertEqual(repr(a), 'Array[int]((5, 6, 7, 8, 9), base=5)')
 
         a = System.Array.CreateInstance(int, (5,), (15,))
         b = System.Array.CreateInstance(int, (5,), (20,))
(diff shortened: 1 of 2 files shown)


Tests/test_array.py
```diff
@@ -320,9 +340,7 @@ def test_base_negative(self):
 
         # test slice indexing
         # 1-dim array [-1, 0, 1]
-        arr1 = System.Array.CreateInstance(int, (3,), (-1,))
-        for i in range(-1, 2):
-            arr1[i] = i
+        arr1 = System.Array[int]((-1, 0, 1), base=-1)
```
slozier, line 343: Negative bases! Should have paid more attention in the previous PR. I assume negative indices are not "from the end" in this case?

slozier (commented):
Unrelated to this PR, but noticed a slight asymmetry in `FixSlice` where it does `ostart < lb` vs `ostop < 0`. Don't think it hurts anything but seemed odd.

BCSharp:
> Unrelated to this PR, but noticed a slight asymmetry in `FixSlice` where it does `ostart < lb` vs `ostop < 0`. Don't think it hurts anything but seemed odd.

It is odd and it does hurt... 

```pycon
>>> from System import *
>>> arr = Array.CreateInstance(Int32, (3,), (2,))       
>>> arr[-1:-5:-1]
IndexError: Index was outside the bounds of the array.
```

Thanks for the catch. This one fell through the cracks. I will issue a separate fix PR.

Tests/test_array.py
BCSharp: > I assume negative indices are not "from the end" in this case?

Yes. Negative bases caught me by surprise too, I discovered them just a few days ago. I always somehow assumed that bases are zero or positive (to support 1-based arrays). Just to be sure, I wrote a test for negative bases and voilà, they work! This got me thinking about how integer indices are treated in .NET and Python. The overview is in [the summary table](https://github.com/IronLanguages/ironpython3/pull/1824#issue-2713347768), but here are some background thoughts.

* Python indices are always relative. Non-negative indices are relative from the beginning, negative indices are relative from the end. The fact that indices are always relative is a direct consequence of lack of the concept of a _base_. Without a _base_, everything is relative.
* .NET integer indices are always absolute. This is a direct consequence of having a _base_ in the first place. The base determines at which (absolute) index the array actually begins.
* Since `Int32` is an absolute index in .NET, there is a special type to represent a relative index: `System.Index`. It is only supported on 1D arrays. IronPython does not support it yet (at least not on arrays).
* In IronPython, we want to support both the Python philosophy and the .NET philosophy seamlessly, so the question is: how should integer indices be treated that is consistent with both .NET and Python? Relative or absolute?
* Luckilly, for 0-based arrays the answer is straightforward:
  * For non-negative indices, relative == absolute (because absolute == relative + base, and  base == 0)
  * For negative indices, well, .NET does not support them, since absolute negative indices are always outside of array bounds for base ==0. So IronPython can adopt the Python convention without creating inconsistencies with .NET.
* Now, in case of base != 0, relative != absolute. So which convention to adopt? Since Python does not know of bases, and non-zero bases imply absolute indexing, the indexing in IronPython in such case has to be absolute too, and it will not conflict with Python. However:
* If base > 0, negative indices, if treated as absolute, will always be out-of-bounds. So, for convenience, IronPython treats them as relative, just as in the case of base == 0. In both cases (base == 0 and base > 0), it is an extension to .NET (i.e. Python convention).
* If base < 0, negative indices cannot be interpreted as relative, because they are used to address some elements in the address space between the base and 0. There is no way around it. So there is a conceptual inconsistency between negative indices with a negative base (absolute, dictated by .NET) and negative indices and 0 base (relative, dictated by Python).
* The only freedom of choice IronPython has, is how to treat negative indices for base > 0? The .NET convention would always throw, the Python convention would make those arrays behave as 0-based arrays. I chose the latter as more practical, but whatever the choice, it will not remove the inconsistency from the previous point. In practice (if there is any), I expect to see 1-based arrays more than any other base (e.g. for a 3-element array, indices 1 ,2 ,3 are then the same as -3, -2, -1).

slozier: Thanks for the explanation. Makes sense.

slozier:
I'm assuming `Multiply` and `Add` are also on your radar?

BCSharp:
> I'm assuming `Multiply` and `Add` are also on your radar?

Yes, but for a different reason than support of non-0-bases. I'll open a separate issue about it, for a dedicated discussion.

PR #1829, opened 2024-12-04
Define UnsupportedOSPlatformAttribute

BCSharp:
I had added `UnsupportedOSPlatformAttribute` in the past already several times, only to remove it again when the final version submitted as a pull request didn't use it. Now again, I have code that uses it, although it is not the final version so in the end it may again not be using it. Nevertheless, I'd like to add this attribute and get it over with for the future. It doesn't harm if it's unused, but it is good to have it available, if needed.


slozier (approved):
Looks good to me.

PR #1832, opened 2024-12-07
Use genuine file descriptors on Posix

BCSharp:
Up until now, file descriptors provided by the Python API (e.g `f.fileno()`) were emulated, that is, they were "fake". This was necessary on Windows, since .NET does not run on top of the C library, but on POSIX systems, at the low level, the genuine file descriptors are used. This PR exposes those genuine file descriptors on the Python API level, making path for a set of POSIX-specific modules, like `fcntl` and `termios`.

I thought that Mono would give me the most problems, because of the [limitation of the `FileStream` constructor](https://github.com/mono/mono/issues/12783), but in the end, it was .NET: it does not share the read/write pointer with duplicated descriptors. I have implemented a workaround (details in the new developer's documentation file), but I am not happy with it. The important thing is that there are no regressions that I know of, and the usage scenarios I have seen in Python various modules should be working.

Here is the overview of the remaining open points:
1. On .NET, `FileStream` is shared between duplicated descriptors, like on Windows. This is a workaround. There are some scenarios for which it doesn't work correctly (see documentation).
2. On Mono, after file descriptor duplication, `UnixStream` is used iso `FileStream`. I do not know if there are any unexpected side effect (I expect none).
3. Handling cases with separate read and write streams is tricky. Currently, it is being used for file open mode 'ab+'. The code includes a workaround for copying over standard descriptors, which is the most common case. Luckily, I think that the separate streams mechanism can be abolished altogether. I plan it for a separate PR.
4. There is still the race condition left in `dup2`. It was there since forever, so it is not a regression. Luckily, it happens only under a very uncommon condition.

### DIFF
--- a/Documentation/file-descriptors.md
+++ b/Documentation/file-descriptors.md
@@ -1,0 +1,144 @@
+# File Descriptors in IronPython
+
+## Windows
+
+The conceptual picture of file descriptors (FDs) usage on Windows, for the most interesting case of `FileStream`:
+
+```mermaid
+graph LR;
+
+FileIO --> StreamBox --> FileStream --> Handle(Handle) --> OSFile[OS File];
+FD(FD) <--> StreamBox;
+```
+
+Conceptually, the relationship between `FD` (a number) and `StreamBox` (a class) is bidirectional because `PythonFileManager` (global singleton) maintains the association between the two so it is cost-free to obtaining the one having the other. FD is not the same as the handle, which is created by the OS. FD is an emulated (fake) file descriptor, assigned by the `PythonFileManager`, for the purpose of supporting the Python API. The descriptors are allocated lazily, i.e. only if the user code makes an API call that accesses it. Once assigned, the descriptor does not change. The FD number is released once the FD is closed (or the associated `FileIO` is closed and had `closefd` set to true.)
+
+It is possible to have the structure above without `FileIO`; for instance when an OS file is opened with one of the low-level functions in `os`, or when an existing FD is duplicated. It is also possible to associate an FD with several `FileIO`. In such cases it is the responsibility of the user code to take care that the FD is closed at the right time.
+
+When FD is duplicated (using `dup` or `dup2`), the associated `StreamBox` is duplicated (there is always a 1-to1 relationship between FD and `StreamBox`), but the underlying `FileStream` objects remain the same, and so are the underlying OS handles. The new FD may be used to create a `FileIO` (or several, just as the original FD). All read/seek/write operations on both descriptors go though the same `FileStream` object and the same OS handle.
+
+```mermaid
+graph LR;
+
+FD1(FD1) <--> StreamBox --> FileStream --> Handle(Handle) --> OSFile[OS File];
+FD2(FD2) <--> StreamBox2[StreamBox] --> FileStream;
+```
+
+The descriptors can be closed independently, and the underlying `FileStream` is closed when the last `StreamBox` using it is closed.
+
+## Posix
+
+On Unix-like systems (Linux, maxOS), `FileStream` uses the actual file descriptor as the handle. In the past. IronPython was ignoring this and still issuing its own fake file descriptors as it is in the case of Windows. Now, however, the genuine FD is extracted from the handle and used as FD at the `PythonFileManager` level, ensuring that clients of Python API obtain the genuine FD.
+
+```mermaid
+graph LR;
+
+FileIO --> StreamBox --> FileStream --> FDH(FD) --> OSFile[OS File];
+FD(FD) <--> StreamBox;
+```
+
+When descriptor FD is duplicated, the actual OS call is made to create the duplicate FD2. In order to use FD2 directly, a new `Stream` object has to be created around it.
+
+### Optimal Mechanism
+
+The optimal solution is to create another `FileStream` using the constructor that accepts an already opened file descriptor.
+
+```mermaid
+graph LR;
+
+FD1(FD1) <--> StreamBox --> FileStream --> FDH1(FD1) --> OSFile[OS File];
+FD2(FD2) <--> StreamBox2[StreamBox] --> FileStream2[FileStream] --> FDH2(FD2) --> OSFile;
+```
+
+In this way, the file descriptor on the `PythonFileManager` level is the same as the file descriptor used by `FileStream`.
+
+Unfortunately, on .NET, somehow, the two independent read/write positions. This is not how duplicated file descriptors should work: both descriptors should point to the same file description structure and share the read/seek/write position. In practice, on .NET, writing through the second file object will overwrite data already written through the first file object. In regular Unix applications (incl. CPython), the subsequent writes append data, regardless which file object is used. The same principle should apply to reads.
+
+Also unfortunately, on Mono, the `FileStream` constructor accepts only descriptors opened by another call to a `FileStream` constructor[[1]]. So descriptors obtained from direct OS calls, like `open`, `creat`, `dup`, `dup2` are being rejected.
+
+### Mono Workaround
+
+To use system-opened file descriptors on Mono `UnixStream` can be used instead of `FileStream`.
+
+```mermaid
+graph LR;
+
+FD1(FD1) <--> StreamBox --> FileStream --> FDH1(FD1) --> OSFile[OS File];
+FD2(FD2) <--> StreamBox2[StreamBox] --> UnixStream --> FDH2(FD2) --> OSFile;
+```
+
+Since FileIO works with various types of the underlying `Stream`, using `UnixStream` should be OK.
+
+Although `UnixStream` is available in .NET through package `Mono.Posix`, this solution still does not work around desynchronized read/write position, which `FileStream` using the original FD1 must somehow maintain independently.
+
+### .NET Workaround
+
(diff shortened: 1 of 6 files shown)


Documentation/file-descriptors.md
```diff
@@ -0,0 +1,144 @@
...
+```mermaid
+graph LR;
+
+FD1(FD1) <--> StreamBox --> FileStream --> FDH1(FD1) --> OSFile[OS File];
+FD2(FD2) <--> StreamBox2[StreamBox] --> FileStream2[FileStream] --> FDH2(FD2) --> OSFile;
+```
+
+In this way, the file descriptor on the `PythonFileManager` level is the same as the file descriptor used by `FileStream`.
+
+Unfortunately, on .NET, somehow, the two independent read/write positions. This is not how duplicated file descriptors should work: both descriptors should point to the same file description structure and share the read/seek/write position. In practice, on .NET, writing through the second file object will overwrite data already written through the first file object. In regular Unix applications (incl. CPython), the subsequent writes append data, regardless which file object is used. The same principle should apply to reads.
```
slozier, line 55: Missing a word in first sentence.

Src/IronPython.Modules/nt.cs
```diff
@@ -351,9 +351,25 @@ public static int dup(CodeContext/*!*/ context, int fd) {
             PythonFileManager fileManager = context.LanguageContext.FileManager;
 
             StreamBox streams = fileManager.GetStreams(fd); // OSError if fd not valid
-            fileManager.EnsureRefStreams(streams);
-            fileManager.AddRefStreams(streams);
-            return fileManager.Add(new(streams));
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
```
slozier, line 354: I'm thinking the platform agnostic version should be the else clause. Though it's possible we haven't been very consistent in doing this. For example, something running on Blazor would fall into the else clause (although the `nt` module might not be super relevant in that case).

Src/IronPython.Modules/nt.cs
```diff
@@ -351,9 +351,25 @@ public static int dup(CodeContext/*!*/ context, int fd) {
...
+                fileManager.EnsureRefStreams(streams);
+                fileManager.AddRefStreams(streams);
+                return fileManager.Add(new(streams));
+            } else {
+                if (!streams.IsSingleStream && fd is 1 or 2) {
+                    // If there is a separate write stream, dupping over stout or sderr uses write stream's file descriptor
+                    fd = streams.WriteStream is FileStream fs ? fs.SafeFileHandle.DangerousGetHandle().ToInt32() : fd;
```
slozier, line 361: Is it documented somewhere that this is the file descriptor? Or would we be relying on an implementation detail? Not that I'm against it, just curious...

Src/IronPython.Modules/nt.cs
```diff
@@ -373,11 +389,45 @@ public static int dup2(CodeContext/*!*/ context, int fd, int fd2) {
                 close(context, fd2);
             }
 
-            // TODO: race condition: `open` or `dup` on another thread may occupy fd2 
+            // TODO: race condition: `open` or `dup` on another thread may occupy fd2 (simulated desctiptors only)
```
slozier, line 392: Typo: desctiptors

Src/IronPython.Modules/nt.cs
```diff
@@ -373,11 +389,45 @@ public static int dup2(CodeContext/*!*/ context, int fd, int fd2) {
...
+                }
+            }
+        }
+
+
+        private static int UnixDup(int fd, int fd2, out Stream? stream) {
+            int res = fd2 < 0 ? Mono.Unix.Native.Syscall.dup(fd) : Mono.Unix.Native.Syscall.dup2(fd, fd2);
+            if (res < 0) throw GetLastUnixError();
+            if (ClrModule.IsMono) {
+                // This does not work on .NET, probably because .NET FileStream is not aware of Mono.Unix.UnxiStream
```
slozier, line 421: typo: UnxiStream

Tests/test_file.py
```diff
@@ -787,6 +787,10 @@ def test_open_abplus(self):
         with open(self.temp_file, "ab+") as f:
             f.write(b"abc")
             f.seek(0)
-            self.assertEqual(f.read(), b"abc")
+            self.assertEqual(f.read(2), b"ab")
```
slozier, line 790: I guess my brain doesn't understand `ab+` mode. :smile: Played around trying to make sense of it, but just ended up finding a behaviour that diverged from CPython...
```py
with open(self.temp_file, "ab+") as f:
    f.write(b"abcdef")
    f.seek(0)
    f.read(2)
    f.write(b"z")
    f.seek(0)
    self.assertEqual(f.read(), b"abzdef")
````

File: Documentation/file-descriptors.md [hunk 7]
```diff
@@ -0,0 +1,144 @@
...
+
+It is possible to have the structure above without `FileIO`; for instance when an OS file is opened with one of the low-level functions in `os`, or when an existing FD is duplicated. It is also possible to associate an FD with several `FileIO`. In such cases it is the responsibility of the user code to take care that the FD is closed at the right time.
+
+When FD is duplicated (using `dup` or `dup2`), the associated `StreamBox` is duplicated (there is always a 1-to1 relationship between FD and `StreamBox`), but the underlying `FileStream` objects remain the same, and so are the underlying OS handles. The new FD may be used to create a `FileIO` (or several, just as the original FD). All read/seek/write operations on both descriptors go though the same `FileStream` object and the same OS handle.
```
slozier, line 18: typos: 1-to1 and though

slozier (commented):
I know it's still a draft, just a few quick comments.

Src/IronPython.Modules/nt.cs
BCSharp: By "the platform agnostic version" do you mean the emulated descriptors version? And that the genuine file descriptors should be limited to Linux and maxOS only? In the past, the platform specific conditions and branches were dictated by the available guard `SupportedOSPlatform`, but now that we have also `UnsupportedOSPlatform` anything is possible. I thought that `UnsupportedOSPlatform` was merely a convenience, but if we were to be considerate of any other platforms than the ones we officially support, it becomes a necessity. OTOH, I see that `IsOSPlatform` accepts only 4 enum values (the 4th being FreeBSD) so how it is supposed to work on Blazor?

Src/IronPython.Modules/nt.cs
BCSharp: This is an implementation detail. The documentation only reads that it is "platform-specific", as far as I could find.

BCSharp: BTW, this piece of code is a hack around the dual-stream case (used for mode `ab+`), which I intend to abolish, in probably the next PR, so this piece will be deleted as well.

Tests/test_file.py
BCSharp: I also do not fully understand the `ab+` mode, so I was playing around with some samples and tests to get a better grip on it; this test being one of them. So this change actually does not belong to this PR, kind of sneaked through from those experiments into the commit. It doesn't hurt though, the more test coverage the better, so I intend to leave it in, but also intend add more `ab+` testing later on.

Another thing I don't understand is why `ab+` is implemented using separate read and write streams. The dual-stream case does not play well with having one genuine file descriptor per file, but more worrisome is that I do not see any actual need for dual-streams at all. This worries me because maybe I am overlooking something essential that led the original programmer to use dual streams, hence the extra tests to try to catch something that doesn't fit in my current mental model of `ab+`. If I don't find anything, the next PR will probably get rid of dual streams.

BCSharp:
Thanks for the review; I put all those typos in there to see is anyone is reading my essays… 😃 …no, I'm just joking.

The PR is draft because I only tested it on Linux and wanted to get the branch over onto my macOS machine. I had no additional work planned, otherwise it was "ready for review". The net465 test on macOS fails, but I don't see any connection with the changes I made. Unfortunately, it also fails locally, for unrelated reasons.

Src/IronPython.Modules/nt.cs
slozier: Yes, I meant using the emulated version when the platform is unknown.

`OSPlatform` can be constructed using arbitrary strings. On .NET it just dispatches to `System.OperatingSystem` which contains all sorts of new goodies. For Blazor the platform string is `BROWSER`. But there are a bunch of others beyond what's in `RuntimeInformation`.

Tests/test_file.py
BCSharp: > I guess my brain doesn't understand `ab+` mode. 😄 Played around trying to make sense of it, but just ended up finding a behaviour that diverged from CPython...

It looks to me that you have found a bug in CPython. If you write four "z" rather than one, it is being correctly appended at the end. I was able to reproduce the strange behaviour you have found on macOS with various CPython versions, up to 3.12.0. Unrelated to this, but I've found [open issues around `ab+` in CPython](https://github.com/python/cpython/issues/86772), so it does not surprise me if there are more.

I have read the CPython code dealing with `ab+` and it simply adds `O_APPEND | O_CREAT` for file opening (in addition to the other usual flags) plus an initial seek to the end in case the first operation is read iso write. Further it relays on the OS to support it, which, according to POSIX documentation, should perform seek to end before each write atomically (apparently, non-atomically on Windows). However, using `pwrite`, it is still possible to write inside a file that is `O_APPEND`, so maybe this is what is happening for writes less than 4 bytes. Or maybe no OS call is done but CPython just updates some internal buffer as an optimization.

In any case, I propose to disregard all discrepancies and simply always write to the end, using one read/write stream.

PR #1834, opened 2024-12-08
Bump dotnet tool to .NET 8

slozier:
Since .NET 6 is no longer supported...

### DIFF
--- a/Package/dotnettool/IronPython.Console.csproj
+++ b/Package/dotnettool/IronPython.Console.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.Build.NoTargets/3.3.0">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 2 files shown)


slozier:
There's a bunch of errors on the github actions net8.0 tests. I think I saw these locally at some point, but now they're gone and I haven't been able to reproduce:
>Could not load file or assembly 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'

BCSharp:
I can't figure it out. I am getting
```
System.IO.FileLoadException: Could not load file or assembly 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The located assembly's manifest definition does not match the assembly reference.
```
every single time and can't get it to work. Maybe something sits in the cache that throws off the process. Or maybe something is missing in the cache.

slozier:
Ok, I guess it was the runtimeconfig file. Not sure why it wasn't an issue on Azure. 😕

BCSharp:
Yes! it work now locally for me. runtimeconfig makes total sense.

PR #1837, opened 2024-12-10
Autodetect processor architecture used for tests

BCSharp:
Useful on macOS.

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -5,12 +5,13 @@ Param(
     [String] $target = "build",
     [String] $configuration = "Release",
     [String[]] $frameworks=@('net462','netcoreapp3.1','net6.0','net8.0'),
-    [String] $platform = "x64",
+    [String] $platform = [Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture,
     [switch] $runIgnored,
     [int] $jobs = [System.Environment]::ProcessorCount
 )
 
 $ErrorActionPreference="Continue"
+if (-not $platform) { $platform = "x64"; }
 
 [int] $global:Result = 0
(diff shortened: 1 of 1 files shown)


slozier (commented):
`ProcessArchitecture` doesn't look like an exact mapping for `TargetPlatform`, but I guess it sort of works.

Based on [docs](https://learn.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2022), if we omit the `TargetPlatform` tag it'll auto-detect.  So might be slightly cleaner to not default platform to anything and omit `RunConfiguration` if unset?

slozier (approved):
Thanks!

BCSharp:
Hmm, it seems not to work for `netcoreapp3.1`. I see it is EOL since 2 years. Do you intend to keep it around for longer?

slozier:
I'm fine with ditching `netcoreapp3.1`. Even `net6.0` is already out of support. MS is moving a bit too fast for my taste. I was also considering dropping the `netstandard2.0` target but maybe too early for that...

BCSharp:
If you are planning to do a release soon, maybe it's worth waiting with the removal till right after the release.

slozier:
Yes, I do want to release soon. .NET 9 issue has come up a number of times. Anyway, I just need to figure out the .NET tool issue and I can start testing the packages.

BCSharp:
Will you have to prepare a release of the DLR too?

BCSharp:
Yes, I have some changes to the DLR in the pipeline; none of which need to be squeezed in before the release. However if you didn't have to do a DLR release I would hold off on those changes not to force an additional release. Since the release is happening anyway, I'll just let them run their course and submit them as they become ready.

Issue #1518, opened 2022-07-30
Argument unpacking is inefficient


slozier:
Argument unpacking is done by calling
```py
if (list.__len__() != 0) {
    return list.pop(0);
}
```
until the list is empty. This is quite inefficient since `pop(0)` shifts the contents of the underlying array to index 0 on every call.

The performance impact can be seen by the test added in https://github.com/IronLanguages/ironpython3/pull/1516. On my machine it takes about 10 seconds to run whereas on CPython it takes a fraction of a second.

PR #1838, opened 2024-12-11
Improve performance in argument unpacking

slozier:
Resolves https://github.com/IronLanguages/ironpython3/issues/1518


slozier:
Technically the public APIs that changed are meant for the consumption by generated expression code. So I'm not too worried about changing them. Maybe they would end up in the DLLs of compiled modules? But I'm not too worried about it. I think we already broke some of these "public" APIs since the 3.4.1 release by adding `CodeContext` arguments.

I don't think thread-safety is an issue? Based on my (limited) understanding of these meta objects it's just generating an expression tree that transforms the given arguments to match the function signature. The whole thing seems overly complicated to me, but not something I really want to take on right now.

PR #1839, opened 2024-12-12
Support mode `ab+` equivalent through os.open

BCSharp:
I think I understand now why IronPython uses sometimes two streams for accessing one file. I may be wrong, but this is what I think is the case:
1. CPython was build on libc, which supports various file opening modes according to POSIX. `O_APPEND | O_CREAT | O_RDWR` is one of them, tagged as mode `ab+`.
2. .NET Framework was build on Win32, which does **not** support file mode `Append` with read access.
3. When CPython is ported to Win32, it uses MSVCRT, which emulates `Append` with R/W by opening with a regular `Open` and then seeking to the end before every write. This is functionally the same as on POSIX but non-atomic, so it may lead to data loss in some circumstances.
4. When .NET was re-developed on POSIX, it maintained the limitation from Win32 of not supporting R/W with `Append`, perhaps for consistency of API.
5. IronPython emulates `ab+` by opening two streams, one for writing with proper file mode `Append`, and a secondary one for reading. Unlike MSVCRT, it does not suffer with problems with data overwrite. 
6. Two streams per file work OK-ish on POSIX, but cause complications when genuine file descriptors come into play (since there are now two open descriptors per one file).

Given above, I decided to keep the two-stream solution, which seems to me is optimal for Win32. In this PR, I extend the two-stream solution to `os.open` (see added test, which is now passing). The usual workarounds for POSIX are still in place, but I still hope that for POSIX, `ab+` can be implemented with one stream — something for another PR.

Other changes in this PR is an explicit set of file position after accessing `SafeFileHandle`, which, according to the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream.safefilehandle?view=net-9.0#remarks) should reset the position to 0. I have not observed this behaviour in the few tests I have run (on macOS/linux the position was unchanged), but I'd rather not argue with the documentation. Perhaps there are some situations when the position is moved. Therefore, `SafeFileHandle` is now only accessed right after opening, before the descriptor/file is passed to the caller (with one exception of `dup2`, which may not even survive the winter).

Finally, wherever I see it, I change the code to put the "platform agnostic" branch in the `else` clause.

### DIFF
--- a/Src/IronPython.Modules/nt.cs
+++ b/Src/IronPython.Modules/nt.cs
@@ -352,10 +352,6 @@ namespace IronPython.Modules {
 
             StreamBox streams = fileManager.GetStreams(fd); // OSError if fd not valid
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) {
-                if (!streams.IsSingleStream && fd is 1 or 2) {
-                    // If there is a separate write stream, dupping over stout or sderr uses write stream's file descriptor
-                    fd = streams.WriteStream is FileStream fs ? fs.SafeFileHandle.DangerousGetHandle().ToInt32() : fd;
-                }
                 int fd2 = UnixDup(fd, -1, out Stream? dupstream);
                 if (dupstream is not null) {
                     return fileManager.Add(fd2, new(dupstream));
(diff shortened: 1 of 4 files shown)


slozier (approved):
Seems fine to me. Only spotted one typo this time (accoding). :wink:

PR #1842, opened 2024-12-16
Fix file seek errors

BCSharp:
Mainly with Mono, which does not accept negative seek positions relative to current position, throwing an overflow exception instead.


slozier (approved):
Looks good to me!

PR #1847, opened 2024-12-22
Disable tests in Mono depending on gc.collect()

BCSharp:
Since I changed to a faster macOS machine, these tests are failing on Mono.

There are still two more tests failing (`test__ssl` and, yup, `test__socket`…), but I haven't investigated them yet. They seem to have a different type of issues than `gc.collect()`.


slozier (approved):
Works for me!

PR #1849, opened 2024-12-23
Fix test__socket, again…

BCSharp:
What an ugly test 😉


slozier (approved):
Thanks! Looks like our mac tests are running again.

PR #1851, opened 2024-12-24
Update test_open_abplus

BCSharp:
Using `raw` better represents what this test is supposed to test — unbuffered access.


slozier (approved):
Thanks!

Issue #1225, opened 2021-05-16
PEP 446 -- Make newly created file descriptors non-inheritable


slozier:
https://docs.python.org/3/whatsnew/3.4.html#pep-446-newly-created-file-descriptors-are-non-inheritable

https://www.python.org/dev/peps/pep-0446/

PR #1000, opened 2020-10-28
Enable some test_memoryio tests

slozier:

PR #1853, opened 2024-12-26
Complete O_flags used for file open

BCSharp:
Includes part of the solution for #1225.

### DIFF
--- a/Src/IronPython/Modules/_io.cs
+++ b/Src/IronPython/Modules/_io.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
@@ -12,6 +12,7 @@ using System.Numerics;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 using System.Text;
 
 using Microsoft.Scripting.Runtime;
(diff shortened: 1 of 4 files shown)


Src/IronPython/Modules/_io.cs
```diff
@@ -54,6 +55,19 @@ public static partial class PythonIOModule {
 
         private static int O_EXCL => RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 0x400 : RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? 0x800 : 0x80;
 
+        [PythonHidden(PlatformsAttribute.PlatformFamily.Windows)]
```
slozier, line 58: Having `PythonHidden` on a `private` seems kind of odd but I guess it doesn't hurt...

Tests/test_file.py
```diff
@@ -701,7 +701,7 @@ def test_errors(self):
 
         with self.assertRaises(OSError) as cm:
             open('path_too_long' * 100)
-        self.assertEqual(cm.exception.errno, (36 if is_posix else 22) if is_netcoreapp and not is_posix or sys.version_info >= (3,6) else 2)
+        self.assertEqual(cm.exception.errno, (63 if is_osx else 36 if is_linux else 22) if is_netcoreapp and not is_posix or sys.version_info >= (3,6) else 2)
```
slozier, line 704: Not that I mind the explicit number check, but wondering if we should be using the `errno` module values instead to make things easier to decipher.

slozier (approved):
Looks good to me!

Tests/test_file.py
BCSharp: Good point, assuming that `errno` is not buggy 😄

Src/IronPython/Modules/_io.cs
BCSharp: Yes, it doesn't hurt, but the less verbosity the better. I assume the same applies to `internal`, though I see there are a few `PythonHidden` that are `internal` already in the codebase.

slozier: Weird, I wonder why we'd have them on `internal` members. Maybe something that was `public` at some point?

BCSharp: Look for example at `ByteArray.UnsafeByteList`. It was established 5 years ago as `internal`, but 4 years ago you added `PythonHidden` in #1000. Do you remember why?

slozier: Can't say I remember, but looking at https://github.com/IronLanguages/ironpython3/pull/693 I changed `Bytes.GetUnsafeByteArray` (which was public) to a `public` property and then in a follow up commit made it `internal`. I guess I forgot to remove the `PythonHidden` when I did this. I'm guessing `ByteArray.UnsafeByteList` is a copy/paste...

PR #1860, opened 2025-01-04
Enable test_mmap on net8.0/POSIX

BCSharp:
Establishing a mmap test from StdLib on POSIX to prevent unnoticed regressions from #1855.

The test does not pass on .NET 6.0, which has problems with mmaps not backed by an existing file. A workaround is possible, but likely quite tricky to get it right (i.e. w/o data leaking). I don't think it is worth the effort given that .NET 6.0 is now past its EOL.

The case of Mono seems hopeless; it dumps core when running this test.


slozier (approved):
Thanks!

PR #1864, opened 2025-01-05
Support additional memory protection flag combinations

BCSharp:
One test is failing on my macOS (using Sequoia on ARM64). I have traced it to python/cpython#107888. I wonder what will happen on CI.

It turns out that `test_mmap` passes on .NET 6.0 on Linux, but not on macOS. Go figure. Anyway, I've relaxed the test run condition accordingly.

I have enabled some more tests that were marked TODO. It required some small and self-evident modifications to the test in StdLib. I've left comments at those places, in case there are some merge conflicts with later Python branches.


slozier (approved):
It's a bit unfortunate about not being able to use `sys.maxsize`. Checking the 32/64-bitness seems to be one of its main uses. Perhaps we should change it...

PR #1865, opened 2025-01-06
Backport fixes to python/cpython#107888

BCSharp:
I kept them as a separate PR, in case they need to be reverted when IronPython migrates to later StdLib versions where these patches are already applied. AFAICS, 3.6 does not contain them.

Fixes local test failure in #1864.


slozier (approved):
Looks good to me. I added a backport comment to make sure I don't zap it by accident.

PR #1867, opened 2025-01-09
Downgrade Ubuntu to Jammy Jellyfish

BCSharp:


slozier (approved):
LGTM. Test failure on macOS was clearly unrelated...

PR #1869, opened 2025-01-09
Rename PythonErrorNumber to PythonErrno

BCSharp:
It turned to be more popular than I first thought and more usage is coming. `PythonErrno` is shorter hence more convenient. I would prefer simply `Errno`, but that could be confused with `Mono.Unix.Native.Errno`.


slozier (approved):
LGTM. Thanks!

PR #1870, opened 2025-01-10
Isolate POSIX-specific native calls in PythonNT into separate file

BCSharp:
This adds platform guards to relevant functions in `PythonNT`. By keeping them in a separate file I hope it will be easier to prevent omissions and inadvertent loads of `Mono.Unix` on Windows. Also, it makes `nt.cs` just a little bit smaller, which is always a good thing.

### DIFF
--- a/Src/IronPython.Modules/_winapi.cs
+++ b/Src/IronPython.Modules/_winapi.cs
@@ -9,6 +9,7 @@
 using System;
 using System.Numerics;
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 using System.Text;
 
 using IronPython.Runtime;
@@ -16,9 +17,11 @@ using IronPython.Runtime.Operations;
 
 [assembly: PythonModule("_winapi", typeof(IronPython.Modules.PythonWinApi), PlatformsAttribute.PlatformFamily.Windows)]
 namespace IronPython.Modules {
+    [SupportedOSPlatform("windows")]
     public static class PythonWinApi {
         #region Public API
 
+        [SupportedOSPlatform("windows")]
         public static object? ConnectNamedPipe(BigInteger handle, bool overlapped = false) {
             if (overlapped) throw new NotImplementedException();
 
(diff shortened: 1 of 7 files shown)


Src/IronPython.Modules/mmap.cs
```diff
@@ -417,7 +417,7 @@ void CheckFileAccessAndSize(Stream stream) {
                             throw PythonOps.OSError(PythonExceptions._OSError.ERROR_ACCESS_DENIED, "Invalid access mode");
                         }
 
-                        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
+                        if (!isWindows) {
```
slozier, line 420: Not really an issue since we're not calling native APIs here, but any idea if the platform analyzer would pick up on an `isWindows` variable and guard against Windows calls?

BCSharp: As it is written now, no platform-specific calls are allowed in `CheckFileAccessAndSize`, regardless whether they are conditioned by `isWindows` in any way. .NET has `SupportedOSPlatformGuardAttribute`, but it cannot be applied to a parameter.

The reason why I replaced the platform check with a parameter is because having the platform test in this function was invalidating the platform context after the first call site, and I needed to make a call to `PythonNT.dupUnix`.

slozier: Cool, didn't know (or forgot) that `SupportedOSPlatformGuardAttribute` existed.  Weird that is breaks the platform check. Alternatively the parameter could be called `allowGrow` or something.

slozier (approved):
LGTM

Src/IronPython.Modules/_winapi.cs
```diff
@@ -9,16 +9,19 @@
...
 using IronPython.Runtime;
 using IronPython.Runtime.Operations;
 
 [assembly: PythonModule("_winapi", typeof(IronPython.Modules.PythonWinApi), PlatformsAttribute.PlatformFamily.Windows)]
 namespace IronPython.Modules {
+    [SupportedOSPlatform("windows")]
     public static class PythonWinApi {
         #region Public API
 
+        [SupportedOSPlatform("windows")]
```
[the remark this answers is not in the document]
BCSharp: No, it's redundant.

PR #1871, opened 2025-01-13
Correct and expand errno mapping to Python exeptions

BCSharp:
Fixes a few error mappings and finds a better place for errno codes in the main IronPython assembly.


slozier (approved):
LGTM!

Issue #1308, opened 2022-01-10
import pty module causes ImportException: No module named 'termios'


ScottKane:
### Prerequisites

* [ Y ] Are you running the latest version?
* [ Y ] Are you reporting to the correct repository?
* [ Y ] Did you perform a cursory search?

### Description

```c#
var engine = Python.CreateEngine();
engine.SetSearchPaths(new []{ "/app/.lib" }); 
var scope = engine.CreateScope();
scope.ImportModule("pty");
engine.Execute(@"# Some python", scope);
```

**Expected behavior:** Imports work and python executes

**Actual behavior:** Unhandled exception. IronPython.Runtime.Exceptions.ImportException: No module named 'termios'

### Versions

Nuget 3.4.0-alpha1

slozier:
Thanks for the report. It looks like `termios` module hasn't been implemented.

jstacoder:
@slozier then why is it documented with the std library as though it is? https://ironpython-test.readthedocs.io/en/latest/library/termios.html

slozier:
@jstacoder no idea. I assume the docs were just forked from the Python docs (with minor changes).

BCSharp:
It looks to me that those documents were a test run / starting point / work in progress that never got fully adapted to IronPython.

BTW, `termios` will be needed for IPython support on Posix systems, so it is on my radar.

PR #1879, opened 2025-01-20
Implement rudimentary version of module `termios`

BCSharp:
`termios` — a module for low-level terminal I/O system control. It doesn't do much beside defining a bunch of system-specific constants and a few direct POSIX calls.[[1]][[2]]  The API is old, so old actually that I think it may be obsolete/unused. Anyone needing to set the baud rate to 300 bytes/second? 6 bits per character? On macOS, for instance, some mandatory but kind of obsolete POSIX flags are defined in the header files, but the syscalls don't support them (or not anymore). However, there are a few things that are still quite useful nowadays and perhaps impossible to perform in any other way in a portable fashion. The two common I have encountered:
1. Setting the terminal input mode to _raw_ (and back to _control-break_)
2. Getting (sometimes also setting) the size of the terminal window (rows, columns)

The implementation of `termios` in this PR is quite rudimentary, meant primarily to be enough to support module `tty` from StdLib (and some other 3rd party ones). The constants defined here are the subset what `termios` from CPython defines. See also #1308

### The Constants

I tried to be thorough but did not try to be complete. The constants defined are the ones in use from other modules (that I have noticed), but once a constant or flag was used, I tried to define the remaining ones from the same group, just for completeness (with the exception of the `TIOC*` codes). This PR comes from my stash some 4 years ago, before we had `generate_os_codes.py`, so all constants are hand-crafted, not generated. To generate them would be messy anyway, because of a lack of clear naming. Perhaps in the future the `TIOC*` codes can be generated.

### The Functions

This is a mock implementation of `termios`; it doesn't actually do any syscalls. Instead it emulates the minimal functionality needed by `tty` by using .NET/IronPython internals, and only for `stdin`. Now that we have genuine file descriptors implemented, this mock code can be replaced by actual syscalls. This will have to be through P/Invoke, as Mono.Unix does not support the termios API (I saw it marked as TODO in Mono). The calls are simple, but the data structure marshalled not so.

[1]: https://www.man7.org/linux/man-pages/man3/termios.3.html
[2]: https://manpages.debian.org/termios(3)

### DIFF
--- a/src/core/IronPython.Modules/termios.cs
+++ b/src/core/IronPython.Modules/termios.cs
@@ -1,0 +1,323 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+
+using System;
+using System.Collections;
+using System.Numerics;
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+
+using Microsoft.Scripting.Runtime;
+
+using IronPython.Runtime;
+using IronPython.Runtime.Operations;
+
+using static IronPython.Modules.PythonIOModule;
+
+[assembly: PythonModule("termios", typeof(IronPython.Modules.PythonTermios), PlatformsAttribute.PlatformFamily.Unix)]
+namespace IronPython.Modules;
+
+[SupportedOSPlatform("linux")]
+[SupportedOSPlatform("macos")]
+public static class PythonTermios {
+
+    public const string __doc__ = "Stub of termios, just enough to support module tty.";
+    // and also prompt_toolkit.terminal.vt100_input
+
+    #region termios IO Control Codes (TIOC*)
+
+    public static int TIOCGWINSZ => RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? 1074295912 : 21523;
+
+
+    #endregion
+
+    #region Other Public Constants
+    // Linux: glibc/bits/termios.h (/usr/include/{x86_64,aarch64}-linux-gnu/)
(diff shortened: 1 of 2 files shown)


src/core/IronPython.Modules/termios.cs
```diff
@@ -0,0 +1,383 @@
...
+    public static int HUPCL  => RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? 0x4000 : 0x0400;  // hang up on last close
+
+    [PythonHidden(PlatformID.MacOSX)]
+    public static int CBAUD  => 0x100f;  // mask for baud rate
+    [PythonHidden(PlatformID.MacOSX)]
+    public static int CBAUDEX => 0x1000;  // extra baud speed mask
+
+
+    // lflag
+    public static uint ECHOKE  => RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? 0x0001u : 0x0800u;       // visual erase for line kill
```
slozier, line 93: I guess these will show up as `System.UInt32` instead of `int` in the Python world. Not necessarily a big deal since this module is somewhat obscure, but something to keep in mind...

slozier (approved):
Thanks!

src/core/IronPython.Modules/termios.cs
BCSharp: Yes, I was debating which type to use here. The issue with `BigInteger` is that it is not very efficient at storing values that don't fit in 32 bits. Here, all those constants are flags and are supposed be used as opaque values, evt. to be OR-ed together, not for some complex integer arithmetics (although it would still be OK, IronPython's support for mixed type integer arithmetics and automatic type conversions is quite extensive). The API that accepts them is/will be using `long` so that it will efficiently capture `UIn32` and negative values of `In32`. Yes, something to keep in mind, though I don't expect it to be a problem.

PR #1883, opened 2025-01-26
Replace test Task with Thread with increased stack size

BCSharp:
Increases the stack size used by the thread running tests to 2 MiB. [Reportedly](https://github.com/dotnet/runtime/issues/96347) the stack size of background treads on .NET 8.0 on macOS is extremely small (0.5 MiB), so it is easy to hit the limit.

This is essentially the last commit from #1879, which does not belong there but I pushed it there to test it with CI. I could not reproduce the stack overflow locally, probably because I develop on ARM which extensively uses registers for parameter passing so its stack usage is lower than on x64. Once this PR is merged, I will rebase #1879 on top of the current main branch (_sans_ the last commit) and presumably the tests will pass.

### DIFF
--- a/tests/IronPython.Tests/Cases/CaseExecuter.cs
+++ b/tests/IronPython.Tests/Cases/CaseExecuter.cs
@@ -11,7 +11,7 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
-using System.Threading.Tasks;
+using System.Threading;
 
 using IronPython;
 using IronPython.Hosting;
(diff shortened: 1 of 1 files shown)


slozier (approved):
Thanks! I guess this has the added advantage that it'll try to abort on .NET Framework?

BCSharp:
> I guess this has the added advantage that it'll try to abort on .NET Framework?

I suppose so. Although we had misconfigured long tasks in the past that were timing out prematurely, my understanding is that the timeout is here to prevent the whole test run to hang indefinitely because of one misbehaving test. If so, the thread will never complete and basically hang on to its resources until the whole process terminates. `ThreadAbort` may leak resources, but it is still better than not releasing them at all. At least we get the stack space back.

The whole solution based on #411 courtesy of @slide.

PR #1884, opened 2025-01-26
Fix `.gitattributes` after restructuring

BCSharp:


Lamparter:
FYI https://github.com/BCSharp/ironpython3/pull/1

PR #1815, opened 2024-11-17
Code Quality: Refactor code and make it more modern

Lamparter:
Hello (again again)!
In this PR I've done the following:
- [x] ~~Run Code Cleanup on the project - remove unnecessary using directives, format documents, etc.~~
	- [ ] ~~Run special **advanced** code cleanup~~
	- [ ] ~~Sort out `.editorconfig`~~
- [x] Update namespaces of files to be consistent
- [x] Change folder structure to match a modern project
- [x] Fix NuGet package sources

...and various other changes that I will add here soon

PR #1885, opened 2025-01-27
Code Quality: Improved the way packages are managed in the codebase

Lamparter:
Just a small change, forwarding out #1815
NuGet doesn't need to be here (it is automatically registered as a package source by Visual Studio)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,6 @@ bin/
 Package/Release/
 *.binlog
 
-# NuGet Packages
-packages/
-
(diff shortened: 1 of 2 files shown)


slozier:
If we're ditching the `packages` folder then we might as well delete the file (I don't think the nunit myget is still pertinent). Also the `packages` filter should be removed from `.gitignore`.

slozier (approved):
Works for me. Thanks!

PR #1888, opened 2025-01-28
Emulate `fcntl.ioctl` for command `TIOCGWINSZ`

BCSharp:
This is a mock implementation of `fcntl.ioctl`, supporting only one command: `termios.TIOCGWINSZ` (i.e. Terminal I/O Control Get WINdow SiZe). It doesn't do any syscalls but provides the result using equivalent .NET functions. It ignores the file descriptor, and always returns the size of the console, so is not appropiate for pseudo-terminals (pty, tty). Nevertheless, it is still useful as some packages depend on this call to get the console window size (`prompt_toolkit.terminal.vt100_input`).

### DIFF
--- a/src/core/IronPython.Modules/fcntl.cs
+++ b/src/core/IronPython.Modules/fcntl.cs
@@ -27,6 +27,33 @@ public static class PythonFcntl {
         """;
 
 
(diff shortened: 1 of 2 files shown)


slozier (approved):
Works for me. Thanks!

PR #1889, opened 2025-01-28
Code Quality: Removed unnecessary files from GitHub statistics

Lamparter:
> @BCSharp while you're fixing this, might as well make this change.
> 
> ---
> 
> This hides files from showing up in GitHub here:
> image
> 
> IronPython doesn't use PSQL significantly enough for it to be detected, and HTML is used only for the documentation, which GitHub automatically enables filters for (that seem to not be working).

> So, this is how GitHub works… I had no idea.
> 
> You have raised a valid point, but although it also affects `.gitattributes`, it has nothing to do with code restructuring, so I'd rather it be done separately. Also, it deserves a proper discussion, e.g.:
> 
> * If we were to hide PLSQL, why not Roff?
> * If we hide PSQL and Roff, then perhaps other 0.0%-used languages will show up? Where does it end?
> * AFAICS, the HTML is being used for tests (and so is PLSQL; IDK about Roff). The question is, should tests be excluded from the "Laungages" statistics? My intuition says no, since we do need to write and maintain tests as well.
> * The above point notwithstanding, should we include/exclude StdLib in the statistics? Technically, it is not IronPython-originated code, but it is tweaked for IronPython and distributed alongside with IronPython, so I presume it should be included.
> * Perhaps StdLib tests could be excluded, but they still require sometimes some maintenance from our project, so it feels like they contribute to the codebase as well.

---

#1884 was merged making https://github.com/BCSharp/ironpython3/pull/1 invalid, the changes have been forwarded here

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -44,3 +44,6 @@ tests/suite/file_without_BOM.txt -text
 
 # Do not normalize on commit.
 *.pdf -text
+
+*.pck linguist-detectable=false
+*.html linguist-documentation=true
--- a/.gitattributes
+++ b/.gitattributes
@@ -14,8 +14,8 @@
 *.ps1 text
 
 # Explicitly declare files that should not have line endings modified, ever
-Tests/file_with_BOM.txt -text
-Tests/file_without_BOM.txt -text
+tests/suite/file_with_BOM.txt -text
+tests/suite/file_without_BOM.txt -text
 
 # Declare files that will always have CRLF line endings on checkout.
 *.bat text eol=crlf
--- a/src/core/IronPython.StdLib/.gitattributes
+++ b/src/core/IronPython.StdLib/.gitattributes
@@ -18,15 +18,15 @@
 *.zip binary
 
 # Specific binary files
-Lib/test/sndhdrdata/sndhdr.* binary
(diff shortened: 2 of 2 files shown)


slozier (commented):
I guess it's fine as long as it doesn't become a game of whack-a-mole. I'm not sure how it decided we have `Roff` (whatever that is) code since if I click on it, it shows no files.

It does highlight an issue with ancient HTML test plans which are (apparently) a big percentage of our codebase. I have to admit I have never even looked at them, but now that I do a lot of it seems to be generated API surface (which doesn't seem very useful).

BCSharp:
I am fine with that change too, it just raises more questions. Like, why is Roff included but not VB? We do have a few of them for testing. Also, the generated API surface with no useful human content can be deleted, I suppose? In any case, it's not something that is bothering me.

slozier:
Let the whack-a-mole begin!
image

slozier:
Is there a way to say we only want to show Python and C#? Not sure any of the other stuff is relevant?

Lamparter:
> Is there a way to say we only want to show Python and C#? Not sure any of the other stuff is relevant?

Sure!

PR #1890, opened 2025-01-28
Run test tasks in foreground

BCSharp:
Run test tasks sequenially in foreground if `-jobs 0`. The default did not change: it still runs all (up to the processor count) task groups in parallel jobs. `-jobs 1` also runs all tasks sequentially, but because it is run as a PowerShell job, the test output is not real-time, but displayed in 10 sec intervals.

This change has two purposes:
1. To be able to monitor the test progress in real-time, without the filter. This is useful when one or a small subset of tasks is run in a query, and especially useful if some tests that operate on stdin and std out misbehave.
2. On some devcontainer systems, PowerShell's [`Start-Job` does not work](https://github.com/PowerShell/PowerShell/discussions/24382) if not run by `root`. `-jobs 0` avoids `Start-Job` altogether.

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -111,6 +111,25 @@ function GenerateRunSettings([String] $framework, [String] $platform, [String] $
     return $fileName
 }
 
+function RunTestTasksInForeground($tasks) {
+    $failedTests = @()
+    foreach ($task in $tasks) {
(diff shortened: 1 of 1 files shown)


slozier (approved):
Nice!

PR #1892, opened 2025-01-29
Code Quality: Hide unnecessary languages from GitHub Linguist

Lamparter:
High score: 9 languages!
5 from main rounds, 4 from bonus round!

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -49,4 +49,7 @@ tests/suite/file_without_BOM.txt -text
 *.pck linguist-detectable=false
 *.c linguist-detectable=false
 
+*.bat linguist-detectable=false
+*.cmd linguist-detectable=false
+
 *.html linguist-documentation=true
--- a/.gitattributes
+++ b/.gitattributes
@@ -46,15 +46,7 @@ tests/suite/file_without_BOM.txt -text
 *.pdf -text
 
 # Hide certain languages from GitHub Linguist
-*.pck linguist-detectable=false
-*.c linguist-detectable=false
-
-*.bat linguist-detectable=false
-*.cmd linguist-detectable=false
-*.ps1 linguist-detectable=false
-*.1 linguist-detectable=false
-*.2 linguist-detectable=false
-*.3 linguist-detectable=false
-*.4 linguist-detectable=false
-
-*.html linguist-documentation=true
+* linguist-detectable=false
+*.cs linguist-detectable
+*.py linguist-detectable
+src/core/IronPython.StdLib/lib/** linguist-vendored
--- a/.gitattributes
+++ b/.gitattributes
@@ -45,5 +45,8 @@ tests/suite/file_without_BOM.txt -text
 # Do not normalize on commit.
 *.pdf -text
 
+# Hide certain languages from GitHub Linguist
 *.pck linguist-detectable=false
+*.c linguist-detectable=false
+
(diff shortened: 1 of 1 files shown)


Lamparter:
@slozier high score! Looks like the game gave me a bonus round as well.

slozier:
I guess this is the sort of list I was trying to avoid. 😀

How many of these are caused by the stdlib lib folder? Wondering if marking that folder as linguist-vendored would be a better approach.

Lamparter:
> How many of these are caused by the stdlib lib folder? Wondering if marking that folder as linguist-vendored would be a better approach.

Only batchfile is used in the stdlib, though it is also used in the tests..

.gitattributes
```diff
@@ -45,5 +45,16 @@ tests/suite/file_without_BOM.txt -text
...
 *.pck linguist-detectable=false
+*.c linguist-detectable=false
+
+*.bat linguist-detectable=false
+*.cmd linguist-detectable=false
+*.ps1 linguist-detectable=false
+*.1 linguist-detectable=false
+*.2 linguist-detectable=false
+*.3 linguist-detectable=false
+*.4 linguist-detectable=false
```
Lamparter, line 58: This is just the stupid design of Linguist - it detects all files with a numbered extension as being Roff.
When IronPython is bumped to later versions of Python [it will need to be changed](https://github.com/IronLanguages/ironpython3/blob/main/eng/package/deb/ipy3.4)

Lamparter:
@slozier another approach to this would to just do:
```gitattributes
* linguist-detectable=false
*.cs linguist-detectable
*.py linguist-detectable
```

.gitattributes
slozier: Right, but `1`, `2` and `3` are coming from the standard library so excluding it would eliminate those (as well as the `pck` that we filtered before). I can live with having `ps1`, `cmd`, `bat` and `c` being picked up by Linguist as we actually some of those in the codebase. That only leaves `4` which would need to be filtered out. So how about something simple like:
```gitattributes
src/core/IronPython.StdLib/lib/** linguist-vendored
eng/package/deb/* -linguist-detectable
*.html linguist-documentation
```
This would probably also end up giving better stats about the code we "own" (yes we do tweak the standard lib, but the changes are very limited).

Lamparter: /eng/package/deb also causes issues with this.
I think the recommendation I made in the comment below is the best solution.

slozier: I excluded eng/package/deb in my code snippet...

Lamparter: Oops, sorry I didn't see that

Lamparter: Why not just this? I think it's the best solution
```gitattributes
* linguist-detectable=false
*.cs linguist-detectable
*.py linguist-detectable
```

slozier: The advantage of `src/core/IronPython.StdLib/lib/** linguist-vendored` is that it gets excluded from out stats (which are disproportionately skewed towards Python right now).

Lamparter: Then just: 
```gitattributes
* linguist-detectable=false
*.cs linguist-detectable
*.py linguist-detectable
src/core/IronPython.StdLib/lib/** linguist-vendored
```

slozier (approved):
Thanks @Lamparter! Updated your PR with your most recent proposal!

PR #1893, opened 2025-01-29
Implement `fcntl` syscalls

BCSharp:
This PR implements the main `fcntl`, plus `flock` and `lockf` and enables `test_fcntl` from the StdLib. `ioctl` remains still a mock.

The signature of syscal `fcntl` is variadic, but luckily this call is supported by _Mono.Unix_. It means that there is the extra translation step from the given _fcntl_ command code to Mono's enum, and then back to the actual OS code. This step will also filter out (as unsupported) all commands that were not defined yet by the OS at the time _Mono.Unix_ was built. For one thing, I miss `F_GETPATH`, which was added in Python 3.9. It is only on macOS, but on Linux one can get the path name from a descriptor by `readlink /dev/fd/«fileno»`. Well, maybe I will write a workaround for this one at some point.

The calls `flock` and `lockf` are not supported by _Mono.Unix_, but their signature is simple enough that the P/Invoke prototypes are not a big deal. With them in place it is finally be possible to fix SQLite on macOS (and improve on Linux).

### DIFF
--- a/src/core/IronPython.Modules/fcntl.cs
+++ b/src/core/IronPython.Modules/fcntl.cs
@@ -118,6 +118,37 @@ public static class PythonFcntl {
     #endregion
 
 
+    #region ioctl
+
+    // supporting fcntl.ioctl(fileno, termios.TIOCGWINSZ, buf)
+    // where buf = array.array('h', [0, 0, 0, 0])
+    public static object ioctl(CodeContext context, int fd, int cmd, [NotNone] IBufferProtocol arg, int mutate_flag = 1) {
+        if (cmd == PythonTermios.TIOCGWINSZ) {
+            using IPythonBuffer buf = arg.GetBuffer();
+
+            Span<short> winsize = stackalloc short[4];
+            winsize[0] = (short)Console.WindowHeight;
+            winsize[1] = (short)Console.WindowWidth;
+            winsize[2] = (short)Console.BufferHeight;  // buffer height and width are not accurate on macOS
+            winsize[3] = (short)Console.BufferWidth;
+            Span<byte> payload = MemoryMarshal.Cast<short, byte>(winsize);
+
+            if (buf.IsReadOnly || mutate_flag == 0) {
+                byte[] res = buf.ToArray();
+                payload.Slice(0, Math.Min(payload.Length, res.Length)).CopyTo(res);
+                return Bytes.Make(res);
+            } else {
+                var res = buf.AsSpan();
+                payload.Slice(0, Math.Min(payload.Length, res.Length)).CopyTo(res);
+                return 0;
+            }
+        }
+        throw new NotImplementedException($"ioctl: unsupported command {cmd}");
+    }
+
+    #endregion
+
+
     #region  flock
 
     [DllImport("libc", SetLastError = true, EntryPoint = "flock")]
@@ -187,7 +218,8 @@ public static class PythonFcntl {
 
     #endregion
 
-    #region Private Methods
+
+    #region Helper Methods
 
     private static int GetFileDescriptor(CodeContext context, object? obj) {
         if (!PythonOps.TryGetBoundAttr(context, obj, "fileno", out object? filenoMeth)) {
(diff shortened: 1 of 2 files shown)


src/core/IronPython.Modules/fcntl.cs
```diff
@@ -27,6 +35,91 @@ a file object.
         """;
 
 
+    #region  fcntl
+
+    [LightThrowing]
```
slozier, line 40: Is the light throwing path expected to occur in normal use? If not then I would consider making a regular exception.

Similar question for the other light throwing methods below.

src/core/IronPython.Modules/fcntl.cs
```diff
@@ -53,10 +146,109 @@ public static object ioctl(CodeContext context, int fd, int cmd, [NotNone] IBuff
         throw new NotImplementedException($"ioctl: unsupported command {cmd}");
     }
 
+    #endregion
+
+
+    #region  flock
+
+    [DllImport("libc", SetLastError = true, EntryPoint = "flock")]
```
slozier, line 154: Wonder what happens on Alpine. 😉

slozier:
For whatever reason github mobile app wouldn't let me add a comment along with my review. My points are just technicalities and not blockers on the PR. I did not hit the approve button because tests are failing on Windows for whatever reason.

BCSharp:
The tests are failing not because of the `fcntl` implementation _per se_. Perhaps the failure is related to #245. In any case, this PR cannot be merged as is as long as the tests are failing. Since I can't fix it on a dime, I will likely just submit a workaround disabling the test on Windows.

src/core/IronPython.Modules/fcntl.cs
BCSharp: Will check 🤷

src/core/IronPython.Modules/fcntl.cs
```diff
@@ -27,6 +35,91 @@ a file object.
...
+        if (arg.Count > maxArgSize) {
+            throw PythonOps.ValueError("fcntl bytes arg too long");
+        }
+
+        if (!NativeConvert.TryToFcntlCommand(cmd, out FcntlCommand fcntlCommand)) {
+            throw PythonOps.OSError(PythonErrno.EINVAL, "unsupported fcntl command");
+        }
+
+        int argSize = arg.Count;
+        IntPtr ptr = Marshal.AllocHGlobal(argSize);
```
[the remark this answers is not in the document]
BCSharp: It's not really necessary. CPython uses 1024 bytes on the stack. I could do the same with `stackalloc`, but it looks to me like a lot of space to put on the stack. Or simply pin a byte array.

src/core/IronPython.Modules/fcntl.cs
BCSharp: It depends on the command but yes, sometimes. Those errors are very inexpensive to return so the usage pattern for some commands is simply try/error. For example handling errors `EAGAIN`, `EWOULDBLOCK`., `EACCES`. Having `LightThrowing` makes it relatively inexpensive here too. _Nota bene_ the code still throws regular exceptions for programming errors like invalid value. That's intentional so that the errors do not go unnoticed when these calls are done from C#.

BCSharp:
This seems not to be an IronPython-specific problem, I am getting the same behaviour with CPython 3.4.4. Skipping the test on Windows looks like the right choice.

src/core/IronPython.Modules/fcntl.cs
BCSharp: It works! However, `Mono.Unix` doesn't load, missing dependency `ld-linux-aarch64.so.1`, which makes ipy pretty unusable. Even when I installed extra package `gcompat` (GNU C Library compatibility layer for `musl`, providing `ld`), `Mono.Unix` was still failing with ` __sprintf_chk: symbol not found`.

BCSharp:
It is an IronPython-specific problem after all. The test should be skipped if it raises `unittest.SkipTest` exception. This is happening in several tests so perhaps it's best to fix it in `CaseExecuter`. I will open another PR for that.

src/core/IronPython.Modules/fcntl.cs
slozier: I guess my comment was more of a joke, but still unfortunate that `Mono.Unix` doesn't work on it. :cry:

slozier (approved):
Thanks!

PR #1895, opened 2025-01-31
Skip test if exception `SkipTest` is raised

BCSharp:
I think this is a better way of handling `fcntl_test` failure on Windows (#1893). It is skipping by exception name `SkipTest` rather than `unittest.SkipTest` because I coudn't figure out how to retrieve the defining module name.


slozier (approved):
Works for me! There are probably a few test cases that are excluded just because of this failure that we could "enable".

Issue #1887, opened 2025-01-28
Improvement of CLI array interface for ND cases


BCSharp:
### The Problem

The current interface of the CLI arrays (beyond of what is provided directly by .NET) is in my opinion somewhat unfortunate. It is not wrong: it simply follows the pattern of other builtin collections, like `list`, `tuple` etc. However, all those other types are 1-dimensional, so their API makes very much sense. At the same time, it does not scale well to more dimensions. This can be seen with `memoryview`, the only builtin Python type that has any notion of multidimensionality. It too follows the list pattern, and when operations are applied on an N-dimensional (ND) `memoryview`, it is either implicitly flattened or raises `NotImplementedError`.

To be more specific, I will use the two cases that the CLI array currently implements that are from the `list` pattern: the addition and multiplication operator.

With `list` etc. addition is concatenation:

```pycon
>>> [1, 2] + [3, 4]
[1, 2, 3, 4]
```

This is very useful for 1D structures, but does not scale up to higher dimensions: even if operator `+` were to mean concatenation, along which dimension should it operate? There is no way to pass an extra parameter indicating the dimension to use. It can be assumed that it is the last dimension, but it leaves a big gap in API for concatenating along arbitrary dimension.

With `list` etc. multiplication is repeated concatenation with itself:

```pycon
>>> (1, 2) * 3
(1, 2, 1, 2, 1, 2)
>>> "ab" * 3
'ababab'
```

This is occasionally useful, and again it does not scale up to higher dimensions.

The next one to consider is slicing. Slicing can be seen as as specific way of indexing, except that it retrieves a substructure, rather than an individual element. The question is how slicing would work on an ND array (currently it is not supported). The only guidance from Python itself would be the behaviour of `memoryview` as the only builtin ND structure. However, beyond a simple element retrieval, `memoryview` does not support much of anything.

```pycon
>>> m = memoryview(b"abcd")
>>> list(m)
[97, 98, 99, 100]
>>> m2 = m.cast('b', (2, 2))
>>> list(m2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NotImplementedError: multi-dimensional sub-views are not implemented
>>> m[0]
97
>>> m2[0, 0]
97
>>> m2[0]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NotImplementedError: multi-dimensional sub-views are not implemented
>>> list(m[:1])
[97]
>>> list(m2[:1])
[...]

PR #1898, opened 2025-02-01
Add checks for non-0-based arrays in addition and multiplication operators

BCSharp:
Followup on https://github.com/IronLanguages/ironpython3/pull/1828#issuecomment-2520379088. In light of proposal #1887 I don't think it is worthwhile to spend effort on those two legacy operator implementations beyond of what was already supported.


slozier (approved):
Thanks!

PR #1899, opened 2025-02-01
Implement simple termios functions

BCSharp:
Implements `tcsendbreak`, `tcdrain`, `tcflush`, `tcflow`, which makes `termios` function-complete up to and including Python 3.10. The exported constants are still not complete. `tcgetattr` and `tcsetattr` remain mocks. To implement `tcgetwinsize`, and `tcsetwinsize` added in 3.11, `fcntl.ioctl` needs to be better implemented (currently it's a stub).


slozier (approved):
Thanks!

PR #1904, opened 2025-02-05
Improve `isatty` on POSIX

BCSharp:
It can be seen as an optimization, or a "fast track", but more importantly, it removes the assumption that all open file descriptors on POSIX are registered with `PythonFileManager`. It is currently true (but only as long as only Python code is considered), but it was not always true in the past and may not be true in the future.

Plus some cleanup in `utime` code.


slozier (approved):
LGTM!

PR #1905, opened 2025-02-05
Support `ioctl` on Mono

BCSharp:
I think I've found a way to detect ARM64 on Mono that is reliable enough.

### DIFF
--- a/src/core/IronPython.Modules/fcntl.cs
+++ b/src/core/IronPython.Modules/fcntl.cs
@@ -362,22 +362,15 @@ public static class PythonFcntl {
     }
 
     private static bool TryGetInt64(object? obj, out long value) {
-        int success = 1;
-        value = obj switch {
-            Missing => default,
-            int i => i,
-            uint ui => ui,
-            long l => l,
-            ulong ul => (long)ul,
-            BigInteger bi => (long)bi,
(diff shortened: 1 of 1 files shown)


slozier (approved):
Lgtm! I guess I should try it on my Pi...

slozier:
> I guess I should try it on my Pi...

No luck on my Pi (not related to this PR though).
```pycon
>>> import fcntl, termios
>>> fcntl.ioctl(0, termios.TIOCGPGRP, "  ")
OSError: [Errno 14] Bad address
```

BCSharp:
What flavour of Linux are you running? Is the error the same across frameworks? `Bad address` usually means that OS got a null pointer iso the buffer. What is the output of `uname -m`?

slozier:
> What flavour of Linux are you running? Is the error the same across frameworks? `Bad address` usually means that OS got a null pointer iso the buffer. What is the output of `uname -m`?

Raspberry Pi OS (64-bit) which is based on Debian 12. I only tried .NET 8.

> Linux raspberrypi 6.6.62+rpt-rpi-v8 #&NoBreak;1 SMP PREEMPT Debian 1:6.6.62-1+rpt1 (2024-11-25) aarch64 GNU/Linux

PR #1907, opened 2025-02-06
Downgrade _macos-latest-large_ to _macos-latest_

BCSharp:
The IronPython3 project does not have access to _macos-latest-large_, so the CI runner on macOS always fails on start generating spurious failures (but then restarts and continues as standard and succeeds anyway. Sometimes.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-22.04, macos-latest-large]
+        os: [windows-latest, ubuntu-22.04, macos-latest]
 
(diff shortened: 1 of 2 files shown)


slozier (approved):
Works for me. I don't even remember why we switched to large...

BCSharp:
It traced it down to a commit enabling .NET 9.0 as the build SDK. Presumably it wasn't working on standard instances yet, and somehow the large instances were available.

BCSharp:
Test `IronPython.test_utf8_mode_stdlib` is extremely slow on Mono, about 6 times slower than on .NET 8.0. It takes 154 s on CI. Even on my M3 Max it takes 83 s. I've increased the timeout from 120 s to 240 s.

PR #1909, opened 2025-02-07
Implement `tcgetwinsize`, `tcsetwinsize` in `termios`

BCSharp:
These functions are added in Python 3.11, but I implemented them now while I have the internals still fresh in my memory.

### DIFF
--- a/src/core/IronPython.Modules/termios.cs
+++ b/src/core/IronPython.Modules/termios.cs
@@ -396,6 +396,64 @@ public static class PythonTermios {
         => tcflow(context, PythonFcntl.GetFileDescriptor(context, fd), action);
 
 
+    // Python 3.11: tcgetwinsize, tcsetwinsize
+
+    [LightThrowing]
+    public static object tcgetwinsize(CodeContext context, int fd) {
+        var ws = new ushort[4];
+        var buf = new MemoryBufferProtocolWrapper<ushort>(ws.AsMemory());
+
+        object result = PythonFcntl.ioctl(fd, TIOCGWINSZ, buf);
+
+        if (ToTermiosError(context, result) is not null and var ex) {
+            return ex;
+        }
+        return PythonTuple.MakeTuple((int)ws[0], (int)ws[1]);
+    }
+
+    [LightThrowing]
+    public static object? tcgetwinsize(CodeContext context, object? fd)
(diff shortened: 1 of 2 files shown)


src/core/IronPython.Modules/termios.cs
```diff
@@ -396,6 +396,64 @@ public static void PerformModuleReload(PythonContext context, PythonDictionary d
...
+        }
+
+        ws[0] = unchecked((ushort)winsize_0);
+        ws[1] = unchecked((ushort)winsize_1);
+        if (ws[0] != winsize_0 || ws[1] != winsize_1) {
+            throw PythonOps.OverflowError("winsize value(s) out of range");
+        }
+
+        result = PythonFcntl.ioctl(fd, TIOCSWINSZ, buf);
+        if (ToTermiosError(context, result) is not null and var ex2) {
```
slozier, line 445: Prefer `is not null and var ex2` over `is object ex2`? Anyway, in this case it could just be `return ToTermiosError(context, result)`. But I'm fine with keeping the pattern for consistency...

slozier (approved):

src/core/IronPython.Modules/termios.cs
BCSharp: > Prefer `is not null and var ex2` over `is object ex2`?

It wasn't a matter of a preference. The form `is object` simply didn't occur to me. I've checked IL and both cases unsurprisingly generate exactly the same code. So now it is a matter of preference, and I don't have any strong feelings around the choice. The former is more clear what the intention is ("test that it is not null and capture into a variable") but the latter is more concise, which is always a good thing. As for readability of the latter, I can learn a new idiom. Since I have used the former pattern on at least two other occasions, I will merge this PR as is, and change them all together once we establish a decisive preference.

> Anyway, in this case it could just be `return ToTermiosError(context, result)`. But I'm fine with keeping the pattern for consistency...

Yes, here I prefer to keep it as is. Not only for consistency with the pattern (easing the cognitive load) but also it better communicates the intent: the function discards the `result` of the `ioctl` call and always returns `null`. It just happens that `result` is always `null` at this point, but this is not relevant. Similarly, I added the explicit argument `mutate_flag: true` to the getter calls to emphasize that the call is expected to return data in-place. It happens that `true` is the default argument, but I still prefer to be explicit, for the sake of clarity of intentions. Note that the setter calls don't set that argument because there are no expectations around its value.

PR #1910, opened 2025-02-11
Implement `termios.tcgetattr`/`tcsetattr` using P/Invoke

BCSharp:
I kept a limited workaround for .NET because using `tcsetattr` to put stdin into the raw mode does not have the desired effect on .NET (works fine on Mono) — one can still edit the input line using all standard readline control keys. I have a feeling that for console support .NET is using lower level calls than the simple `read` on a descriptor 0, however, I didn't have time to investigate this in depth. I am submitting this PR with the workaround though before my window of availability for this project narrows.


Lamparter (commented):
Can you use LibraryImportAttribute instead?

Lamparter:
What about using `#if !NETFRAMEWORK && NET_7_0_OR_GREATER`?

BCSharp:
> What about using `#if !NETFRAMEWORK && NET_7_0_OR_GREATER`? That would be a bit messy though, maybe just leave it for now

`#if NET_7_0_OR_GREATER` should suffice. However, this would further multiply  the number of prototypes. Because of platform differences, there are already two per call, which would grow to four per call, which we would have to debug, test, and maintain. This module is a niche module, rarely used, and even if so, the calls are infrequent, so I don't think that the performance gain is worth sacrificing code maintainability. Perhaps `fcntl.lockf` and `flock` would be more worthwhile. Although I wonder how much performance gain is really there. The runtime stub generation should be happening only once per run.

I thought that the biggest advantage of `LimportLibrary` is that it allows for AOT compilation, but this is something that is already tricky (if not impossible for any meaningful use) for IronPython. I can be totally wrong here as I have no experience with the IronPython compiler (`ipyc`, `ngen`, etc.).

I am putting this PR in _draft_ because #1912 has to be sorted out first.

Lamparter:
> I thought that the biggest advantage of `LimportLibrary` is that it allows for AOT compilation, but this is something that is already tricky (if not impossible for any meaningful use) for IronPython. I can be totally wrong here as I have no experience with the IronPython compiler (`ipyc`, `ngen`, etc.).

Yeah, that's ultimately the goal
Though don't forget that IronPython is powered by the ***dynamic*** language runtime 😄

slozier (approved):

src/core/IronPython.Modules/termios.cs
```diff
@@ -19,6 +19,7 @@
 using IronPython.Runtime.Types;
 
 using static IronPython.Modules.PythonIOModule;
+using IronPython.Hosting;
```
slozier, line 22: Not sure if this is VS being helpful again or if it's actually needed?

BCSharp: No idea where it came from (or why). Funny how all those artificially intelligent tools are "clever" enough to automatically add extra `using` clauses, but not clever enough to remove them when not needed after all or even to keep them sorted.

PR #1911, opened 2025-02-12
Apply `is object` test

BCSharp:
OK, let's settle on `is object` test instead of `is not null and var`.


slozier (approved):
Works for me!

I'm sure there are a number of other patterns which are not consistently used throughout the codebase. Either because they didn't exist at the time, flavour of the day, author preference. Anyway, I'm not super bothered by it, but would be interesting to see if there are configs or analyzers that can be used help normalize stuff.

PR #1912, opened 2025-02-12
Fix conversions in `_SimpleCData`

BCSharp:
Besides handling `__index__`, it also better handles `BigInteger` values, which are now Python `int`. This probably should have been done as part of #1329, but at that time I wasn't aware of this corner of the codebase.

Also, there is a bunch of int/index conversions in `Converter` that partially overlap the conversions in `PythonOps`. I wonder it they could be consolidated somehow.


BCSharp:
It started as simply preferring `__index__` over `__int__` in ModuleOps, but it grew to encompass all kind of conversion fixes and improvements.

src/core/IronPython.Modules/ModuleOps.cs
```diff
@@ -357,13 +354,8 @@ public static IntPtr GetObject(object value) {
...
         public static long GetSignedLongLong(object value, object type) {
-            int? res = Converter.ImplicitConvertToInt32(value);
-            if (res != null) {
-                return res.Value;
-            }
-
-            if (value is BigInteger) {
-                return (long)(BigInteger)value;
+            if (PythonOps.TryToInt(value, out BigInteger bi)) {
+                return unchecked((long)(ulong)(bi & ulong.MaxValue));
```
slozier, line 358: I guess this will no longer throw if the value is too big for a `long`. Not too familiar with these methods, but just confirming that this is the intended behaviour?

src/core/IronPython.Modules/_ctypes/SimpleType.cs
```diff
@@ -709,10 +710,13 @@ void INativeType.EmitReverseMarshalling(ILGenerator method, LocalOrArg value, Li
                 }
             }
 
-            private static void EmitInt64ToObject(ILGenerator method, LocalOrArg value) {
+            private static void EmitXInt64ToObject(ILGenerator method, LocalOrArg value) {
```
slozier, line 713: What does `X` stand for? :smile:

src/core/IronPython.Modules/ModuleOps.cs
slozier: I guess `test_conversions_overflow` answers my question...

slozier (approved):
Not super familiar with this part of the codebase, but looks good to me.

src/core/IronPython.Modules/_ctypes/SimpleType.cs
BCSharp: Wildcard `*` 😄

PR #1913, opened 2025-02-14
Run ctypes tests on macOS

BCSharp:
It also adds a `.pyd` test module for architecture _aarch64_ on Linux.


src/core/IronPython.Modules/_ctypes/SimpleType.cs
```diff
@@ -251,11 +251,11 @@ object INativeType.SetValue(MemoryHolder/*!*/ owner, int offset, object value) {
                 switch (_type) {
                     case SimpleTypeKind.Boolean: owner.WriteByte(offset, ModuleOps.GetBoolean(value, this)); break;
                     case SimpleTypeKind.Char: owner.WriteByte(offset, ModuleOps.GetChar(value, this)); break;
-                    case SimpleTypeKind.SignedByte: owner.WriteByte(offset, ModuleOps.GetSignedByte(value, this)); break;
+                    case SimpleTypeKind.SignedByte: owner.WriteByte(offset, (byte)ModuleOps.GetSignedByte(value, this)); break;
```
slozier, line 254: Maybe add an `unchecked`. I know it's unchecked by default but I like to be explicit about it...

src/core/IronPython.Modules/_ctypes_test.cs
```diff
@@ -8,13 +8,19 @@
 
 using System;
 using System.IO;
+using System.Runtime.InteropServices;
+
+using Mono.Unix.Native;
 
 using IronPython.Runtime;
+using System.Runtime.Versioning;
```
slozier, line 16: Maybe sort the usings if you feel like it. :smile:

tests/IronPython.Tests/Cases/CPythonCasesManifest.ini
```diff
@@ -20,6 +17,14 @@ Reason=Current implementation of get_last_error needs to be debugged
 [CPython.ctypes.test_functions]
 Ignore=true
 
+[CPython.ctypes.test_frombuffer]
+RunCondition=NOT $(IS_MONO)
```
slozier, line 21: I'm fine disabling on Mono, but I guess it was running before on Linux?

slozier (approved):
Looks good to me! I guess this was prompted by the tests failing on https://github.com/IronLanguages/ironpython3/pull/1912?

src/core/IronPython.Modules/_ctypes/SimpleType.cs
BCSharp: Ha!, I thought you would comment on my converting `IntPtr` to `void*`. 😃 

About narrowing casts I think we are on the same page — I, too, prefer to be explicit about the intentions. As a rule, I always put a narrowing cast in a `checked` or `unchecked` guard, regardless how the project is being compiled by default — except if from the nature of the data or the program logic it is clear that no overflow is possible (an int containing a char, for example).

Here I have those casts guarded as unchecked already in the follow up commits in #1912; If you don't mind, I prefer to keep them there: If I did it here, the temptation would be to add `unchecked` to other preexisting casts in this switch statement (or farther), and not doing it would raise more questions ("why `unchecked` here but not there?"). In #1912 more unchecked casts are coming, so let's do them together. Also, here I just wanted to limit any changes to specifically what was required to fix marshalling bugs on ARM, so the less changes the better for my future self or whoever looking back trying to understand the whats and whys.

src/core/IronPython.Modules/_ctypes_test.cs
BCSharp: Sure!

tests/IronPython.Tests/Cases/CPythonCasesManifest.ini
BCSharp: Well, sort of, not really… It depends on when the GC runs the finalizers and on Mono it is unpredictable. So the test sometimes works and sometimes fails. This happens more likely on my ARM64 than CI, I've noticed, not sure why. Perhaps CI, as more memory-constrained, prompts Mono to run the GC more aggressively, so it just happens to work.

Since we have already way to many or too frequently false positives to the point of it being frustrating, whenever I see this pattern that Mono is not running finalizers on `GC.Collect()` I disable the test on Mono, with the appropriate comment. Once Mono releases a version with a fixed implementation of `GC.Collect()` (if it ever releases any new version), we can re-enable those tests. 

In this particular example it was possible to create `test_frombuffer_stdlib` and, just for Mono, run safe tests selectively, but, frankly, I have very little patience for Mono. The disabled test should not be framework-specific so .NET tests should do decent work anyway.

BCSharp:
> I guess this was prompted by the tests failing on #1912?

Indeed. Here I am, happily hacking away thinking I have the safety net of the tests, and, to my surprise, discovering that none of the `ctypes` tests are enabled on macOS, which is my primary development system nowadays. Then I am even more surprised to discover that a few years ago (#1441) I already discovered that `ctypes` on macOS were not supported but have completely forgotten that. I guess I am doing to many projects in parallel…

src/core/IronPython.Modules/_ctypes/SimpleType.cs
BCSharp: > As a rule, I always put a narrowing cast in a `checked` or `unchecked` guard,

I see that I was not following this rule in #1441. OK, let's call it a guideline… 😃

slozier: > Here I have those casts guarded as unchecked already in the follow up commits in #1912;

Fine by me. I guess the other PR will keep the return type of `GetUnsignedShort` as a `short` so there's no casting issue at this call site?

I've been meaning to try an turn on `CheckForOverflowUnderflow` at some point to see if it would find any errors. Unfortunately I can never seem to find the time to actually try it...

BCSharp: > I guess the other PR will keep the return type of `GetUnsignedShort` as a `short` so there's no casting issue at this call site?

I don't think I understand what you mean. Did you mean :

> the other PR will keep the return type of `GetUnsignedShort` as a `ushort`?

I changed the return type exactly because casting from a `short` returned from `GetUnsignedShort` call in the runtime-gerenated marshalling code was causing the sign bit extended incorrectly.  I guess this may be one of those Arm/Intel differences. In #1912 I intended to change more return signatures and then add `unchecked()` guards where appropriate. Anyway, we can continue this discussion in #1912, when I resubmit it.

PR #1921, opened 2025-02-25
Enable nullable annotations in StructType and UnionType from ctypes

BCSharp:


src/core/IronPython.Modules/_ctypes/StructType.cs
```diff
@@ -282,16 +282,18 @@ private void SetFields(object fields) {
...
 
                     if (bitCount != null) {
-                        size += lastType.Size;
+                        // incomplete last bitfield
+                        // bitCount not null implies at least one bitfield, so at least one iteration of the loop above
+                        size += lastType!.Size;
                     }
 
-                    _fields = allFields.ToArray();
+                    _fields = [..allFields];
```
slozier, line 290: Interesting, never seen this syntax before. I haven't been keeping up with the more recent C# versions.

Not that I'm concerned about performance here, but is this smart enough to use `ToArray` (which is presumably more performant than a `foreach` statement)?

slozier (approved):

src/core/IronPython.Modules/_ctypes/StructType.cs
BCSharp: > but is this smart enough to use `ToArray`

Yes, in this context it does, and in other contexts the compiler is allowed to do any other optimizations it is aware of. The `..` operation is called now [_spread_](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/collection-expressions#spread-element), which arguably is nicer than (but practically equivalent to) _splat_, which we [discussed before](https://github.com/IronLanguages/dlr/issues/249#issuecomment-728292246).

PR #1922, opened 2025-02-25
Implement _thread.RLock

slozier:
Implement `_thread.RLock` and enable some `test_threading` tests.

I got bit by `functools.lru_cache` which uses a dummy lock if `_thread.RLock` is not available. 😢

### DIFF
--- a/src/core/IronPython.Modules/_thread.cs
+++ b/src/core/IronPython.Modules/_thread.cs
@@ -35,7 +35,7 @@ namespace IronPython.Modules {
 
         #region Public API Surface
 
-        public static double TIMEOUT_MAX = 0; // TODO: fill this with a proper value
+        public static double TIMEOUT_MAX = Math.Floor(TimeSpan.MaxValue.TotalSeconds);
 
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Security", "CA2104:DoNotDeclareReadOnlyMutableReferenceTypes")]
         public static readonly PythonType LockType = DynamicHelpers.GetPythonTypeFromType(typeof(@lock));
@@ -138,13 +138,15 @@ namespace IronPython.Modules {
 
         #endregion
 
+#nullable enable
+
         [PythonType, PythonHidden]
         public sealed class @lock {
-            private AutoResetEvent blockEvent;
-            private Thread curHolder;
+            private AutoResetEvent? blockEvent;
+            private Thread? curHolder;
 
             public object __enter__() {
-                acquire(true, -1);
+                acquire();
                 return this;
             }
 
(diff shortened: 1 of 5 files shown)


slozier:
Probably not related to the PR, but got this failure during CI on Ubuntu:
```
2025-02-26T00:44:34.1619783Z [net462 cpython]  ERROR: test_fcntl_64_bit (__main__.TestFcntl)
2025-02-26T00:44:34.1620469Z [net462 cpython]  ----------------------------------------------------------------------
2025-02-26T00:44:34.1625679Z [net462 cpython]  Traceback (most recent call last):
2025-02-26T00:44:34.1626165Z [net462 cpython]    File "/home/vsts/work/1/s/src/core/IronPython.StdLib/lib/test/test_fcntl.py", line 121, in test_fcntl_64_bit
2025-02-26T00:44:34.1629707Z [net462 cpython]      fd = os.open(os.path.dirname(os.path.abspath(TESTFN)), os.O_RDONLY)
2025-02-26T00:44:34.1630169Z [net462 cpython]  PermissionError: [Errno 13] Permission denied: '/home/vsts/work/1/s/src/core/IronPython.StdLib/lib/test'
```

BCSharp (approved):
Looks good to me.

BCSharp:
> Probably not related to the PR, but got this failure during CI on Ubuntu

Were there any other tests running in parallel? Looks like not-parallel-safe issue.

BCSharp:
> I also saw it on github CI (see snippet below, not sure if you have access, but here are the [raw logs](https://productionresultssa1.blob.core.windows.net/actions-results/5c8371bf-8e93-4e02-9d1d-afe731cc4a01/workflow-job-run-b6923506-d790-581a-6b32-ef448705200b/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-02-26T14%3A12%3A34Z&sig=WbqfNRqpUKY1C1XiY%2BGfgwdhOJXRiDcz4fSdHnFFGn8%3D&ske=2025-02-26T23%3A25%3A49Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-02-26T11%3A25%3A49Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-01-05&sp=r&spr=https&sr=b&st=2025-02-26T14%3A02%3A29Z&sv=2025-01-05)) where each framework runs one after the other so there should be no parallel issues (although cpython and ironpython stuff could be running at the same time I guess).

I don't have access to those logs but I was able to see the failures in one of your failed runs in this PR and have reproduced it locally. It only happens on Mono. It is caused by the implementation of `os.open`, which on Mono, does not support opening directories. Recently, when working on the support of genuine file descriptors, I have migrated the .NET implementation of `open` to use a direct syscall, which can open directories. I have left Mono to do it the old way, not to introduce regression. That is, on Mono, `FileStream` is still being used for opening, mainly due to limitations in Mono's `MemoryMappedFile`.

The test used to pass without problems because it is being skipped if `fcntl.F_NOTIFY` or `fcntl.DN_MULTISHOT` are not defined, and these were added only very recently (the test is also skipped on macOS and on Linux/ARM). I don't want to paddle upstream with Mono, so I am inclined to leave it where it is, and simply disable the test on Mono. But instead of disabling it in the `.ini` file completely (the other tests in `test_fcntl` are passing and doing some useful tests) or create yet another  fine-grained `..._stdlib` test ovveride (extra overhead for IronPython tests), how about we simply hide the constant `F_NOTIFY` and all  the`DN_` constants in `fcntl` on Mono builds? This will cause the test to be skipped, and, since you can't get an open directory descriptor on Mono (by using the regular Python API at least), `F_NOTIFY` on Mono is pretty much useless anyway.

slozier:
I'm fine with hiding the constants on Mono. I was going to ask how we'd accomplish that but I guess a simple `#if NET` (or does it need to be `NETCOREAPP`) would do the trick.

PR #1925, opened 2025-03-05
Refactor CI workflow

Lamparter:
This PR splits each CI run into matrixes to make it easier to debug Ironpython builds and make out what's failing.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,21 +4,24 @@ on: [push, pull_request]
 
 jobs:
   build:
-
     runs-on: ${{ matrix.os }}
 
     strategy:
       fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-22.04, macos-latest]
+        framework: ['net462', 'net6.0', 'net8.0']
 
     steps:
+      # Prerequisites
     - name: Install tools
       if: matrix.os == 'ubuntu-22.04'
       run: sudo apt-get -yq install mono-vbnc dos2unix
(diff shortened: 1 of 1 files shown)


slozier (changes requested):
I don't have an objection to the concept, but unfortunately I don't think it's working as intended. Build is just producing 3x the same package and no tests are running.

I don't think framework arguments have an impact on build or package. And in any case, we want the packages to include binaries for all frameworks so it should not be restricted. I would replace the `build` step with `package` and just do a matrix on `os`. 

For the test issues you're going to have to do add back the build step before running the tests.

I guess we end up repeating the build multiple times, but probably going to be faster overall with everything running in parallel?

.github/workflows/main.yml
```diff
@@ -31,26 +33,59 @@ jobs:
...
+        framework: ['net462', 'net6.0', 'net8.0']
+
+    steps:
+      # Prerequisites
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+
+      # Setup .NET
+    - name: Setup .NET ${{ matrix.framework }}
```
slozier, line 73: This one should say `Setup .NET 6.0`.

Lamparter: Good catch, thanks!

slozier (approved):
Looks good to me!

BCSharp:
This should get rid of all those false positives due to tests not being parallel-safe, it think. 👍

PR #1938, opened 2025-04-03
Implement file locking in SQLite on Posix

BCSharp:
The file locking strategy in SQLite was basically Windows-oriented, with some provisions for limited-trust environment like Silverlight, WinRT, and Windows Mobile. These are obsolete but somebody repurposed the limited-trust strategy to work on Unix. This worked on Linux (sort of, shared locks were seemingly emulated in a way that only SQLite respected), but wasn't working at all on macOS.

This PR implements proper locking on Linux and macOS through a fcntl syscall via _Mono.Unix_. Locking on Windows is unchanged.

I resisted the temptation to start cleaning up the code (it is even badly formatted), because there would be no end to it, and it is better to replace it by a modern SQLite implementation in a longer run anyway.


slozier (approved):
LGTM!

> it is better to replace it by a modern SQLite implementation in a longer run anyway.

I agree trying to maintain this is not a good long term solution. Although I do like having a purely managed implementation...

PR #1939, opened 2025-04-04
ARM64 test workarounds

BCSharp:
The recent release of Mono — 6.20.0 — sports native support for ARM64 on macOS. Unfortunately, it appears that conversions to floating point numbers can be one LSB off from what is happening on x86_64. This causes some of the tests to fail. This PR contains a couple of workarounds so that the overall test tun on ARM64/macOS passes.

### DIFF
--- a/src/core/IronPython/Lib/iptest/test_env.py
+++ b/src/core/IronPython/Lib/iptest/test_env.py
@@ -4,6 +4,7 @@
 
 import os
 import sys
+import platform
 
 #------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


slozier (approved):
Hmm, there are new Mono releases? Google doesn't know about Mono 6.20. :smile: I did find a 6.14 which I assume is what you meant.

BCSharp:
> I did find a 6.14 which I assume is what you meant.

Yep, that's the one… I think I was dreaming about 6.20 😄 No binaries though, it's a DIY kit, so I haven't tested in on Linux/ARM64 yet.

PR #1940, opened 2025-04-05
Replace some documents in `./docs` with links to the wiki

BCSharp:
Recently I have been updating the existing IronPython documentation on the wiki (there are two new pages [Interop with .NET](https://github.com/IronLanguages/ironpython3/wiki/Interop-with-.NET) and [Running packages from PyPI](https://github.com/IronLanguages/ironpython3/wiki/Running-Python-packages-from-PyPI), comments and corrections welcome). So the wiki documentation and the originals in the `./docs` folder diverge more and more. To avoid confusion about which is up to date, I replaced the content of some documents in `/docs` that do not need to be versioned together with the code with links to the corresponding wiki pages. In theory those documents in `./docs` could be removed altogether, but they are already being referenced from some existing Issues and Pull Requests, and from Readmes of the packages published on NuGet and elsewhere.


slozier (approved):
Thanks!

PR #1943, opened 2025-04-08
Use actual container type name for repr

BCSharp:
This PR changes the `repr()` on the builtin container types for their subclasses to display the subclass name rather than always the root base type. Similarly what is already done for `set`: `repr(set()` is `set()` but if `class MySet(set): pass` is defined, `repr(MySet())` is `MySet()`. The changed container types are:
- `bytearray`
- `array.array`
- `collections.defaultdict`
- `collections.deque`

This change has been implemented in Python 3.7. but in 3.4 - 3.6 there is one test in StdLib that catches the base class name; this has been upgraded here to the test from 3.7.


slozier (approved):
Looks good to me!

PR #1945, opened 2025-04-11
Implement signal constants on POSIX platforms

BCSharp:
What has changed:
* The values of existing constants in module `signal` are changed to reflect the actual values on a given platform (previously the values were always Windows-specific).
* Missing signal constants are added for Unix signals.
* `test_signal` from the IronPython test suite is modified to pass with CPython on all OS platforms supported by IronPython.
* The implementation of module `signal` is modified to pass the modified `test_signal`. This in practice means primarily changes to error handling.

What has not changed:
* The set and values of signal constants on Windows stays the same. For what I can see it is pretty complete.
* Signal handling is the same (this means, all signals on Windows, and only `SIGINT` on Unix, no new signals are being handled yet).
* No new (missing) functions are added to `signal`.


slozier (approved):
LGTM. Thanks!

PR #1947, opened 2025-04-16
Rename files with subclasses of `PythonSignalState`

BCSharp:
These two files contain other parts of `partial class PythonSignal` (implementing module `signal`) so they belong together with file `signal.cs`. With this rename they are sorted together with `signal.cs` when displayed in the file explorer.

This commit is factored out from a larger upcoming PR for signal, but for the sake of a clean git history of renames I prefer to keep it as a separate commit, rather than squashed together with other changes.


slozier (approved):
Works for me. We could also consider preemptively adding the `_` prefix `_signal.NtSignalState.cs` since in 3.5 the C module was renamed to `_signal`.

BCSharp:
> We could also consider preemptively adding the `_` prefix `_signal.NtSignalState.cs` since in 3.5 the C module was renamed to `_signal`.

Not to defeat the purpose of this PR, this would require of renaming the existing `signal.cs` to `_signal.cs` as well. Unless you foresee problems with merges into 3.6, I'd keep the naming as is. I am counting on an assumption that with explicit renames like this one, merges should be easier as they should be able to track the file history across renames with 100% confidence.

BCSharp:
To clarify: after the next merge to 3.6, rename `signal.*.cs` to `_signal.*.cs` on the 3.6 branch only in a separate commit.

slozier (approved):

PR #1950, opened 2025-04-25
Obsolete PythonOps.CallWithArgsTuple

slozier:
Follow-up to my comment in https://github.com/IronLanguages/ironpython3/pull/1949.

### DIFF
--- a/src/core/IronPython.Modules/_ctypes/_ctypes.cs
+++ b/src/core/IronPython.Modules/_ctypes/_ctypes.cs
@@ -361,40 +361,44 @@ namespace IronPython.Modules {
             return new NativeArgument(instance, "P");
         }
 
-        public static object call_cdeclfunction(CodeContext context, int address, PythonTuple args) {
+#nullable enable
+
+        public static object? call_cdeclfunction(CodeContext context, int address, [NotNone] PythonTuple args) {
             return call_cdeclfunction(context, new IntPtr(address), args);
         }
 
-        public static object call_cdeclfunction(CodeContext context, BigInteger address, PythonTuple args) {
+        public static object? call_cdeclfunction(CodeContext context, BigInteger address, [NotNone] PythonTuple args) {
             return call_cdeclfunction(context, new IntPtr((long)address), args);
         }
 
-        public static object call_cdeclfunction(CodeContext context, IntPtr address, PythonTuple args) {
+        public static object? call_cdeclfunction(CodeContext context, IntPtr address, [NotNone] PythonTuple args) {
             CFuncPtrType funcType = GetFunctionType(context, FUNCFLAG_CDECL);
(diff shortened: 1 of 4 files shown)


BCSharp:
I suppose that once this get merged, the [wiki page](https://github.com/IronLanguages/ironpython3/wiki/Differences-between-versions#signature-changes-after-342) will need to be updated as well. Or do you intend to revert to the original signature (since the whole method is now obsolete and not used in the codebase)?

slozier:
> I suppose that once this get merged, the [wiki page](https://github.com/IronLanguages/ironpython3/wiki/Differences-between-versions#signature-changes-after-342) will need to be updated as well. Or do you intend to revert to the original signature (since the whole method is now obsolete and not used in the codebase)?

Hmm, I guess I could revert to the original and update the wiki to say the method is obsolete and will be removed in a future release.

My main issue with the public API surface is that it's very large and it's not clear whether the APIs are public:
1. so people can use them (e.g. something Ironclad or IronPython.SQLite would need)
2. as an implementation detail (e. g. used by Expressions)
3. because no one bothered to think about it

Might be interesting to create some reference assemblies with the API surface we want to support. If only I had more time... :smile:

BCSharp:
Frankly, I don't understand why there are any `Call*` methods in `PythonOps`. Why not use `PythonCalls.Call`? Are`PythonOps.Call*` used anywhere through reflection? I know that there are some public static methods defined specifically for the purpose of accessing from Expressions, we could mark the whole class hosting them as "obsolete" with a message that this is internal IronPython API, not to be used by user code (or "use at your own risk"). AFAIK, ObsoleteAttribute does not affect callability through reflection, only the compiler.

Reviewing the whole IronPython public API looks like a daunting task, that's why I didn't want to take a stance on `CallWithArgsTuple` lest I be tempted to start doing reviewing of stuff like that more and more. Currently I'm just trying to wrap a few loose ends before other projects saturate my availability.

slozier:
> Frankly, I don't understand why there are any `Call*` methods in `PythonOps`. Why not use `PythonCalls.Call`? Are`PythonOps.Call*` used anywhere through reflection? I know that there are some public static methods defined specifically for the purpose of accessing from Expressions, we could mark the whole class hosting them as "obsolete" with a message that this is internal IronPython API, not to be used by user code (or "use at your own risk"). AFAIK, ObsoleteAttribute does not affect callability through reflection, only the compiler.

Not sure why they're not going to `PythonCalls.Call` directly or if they're used by reflection. Maybe I'll do another PR to update the call sites to use `PythonCalls` directly (and then we can see if anything is still calling it). I've been marking the "internal" (reflection) stuff with `EditorBrowsable(EditorBrowsableState.Never)` to try and discourage use.

PR #1953, opened 2025-04-27
Unregister signal handlers during engine shutdown

BCSharp:
This PR is relevant for hosting scenarios. Importing module `signal` in one IronPython engine registers a Ctrl+C signal handler, but it was not properly unregistered on engine shutdown. This means the handler was being referenced from the static `System.Console` (on Windows I suppose from the callback marshalling code) and caused a memory leak and interference with another later instance of the engine (at best).

The solution in this PR is to make `PythonSignalState` disposable and modify `PythonContext` that it recognizes disposable module states. For good measure I implemented a finalizer too although I think it will never (i.e before process termination) be called because without a call to `Dispose()` reference will stay alive and not being GC-ed.

### DIFF
--- a/src/core/IronPython.Modules/signal.NtSignalState.cs
+++ b/src/core/IronPython.Modules/signal.NtSignalState.cs
@@ -27,6 +27,17 @@ namespace IronPython.Modules {
             }
 
 
+            protected override void Dispose(bool disposing) {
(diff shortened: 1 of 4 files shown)


slozier (approved):
Looks good to me.

PR #1981, opened 2026-01-31
Use string.Equals for string comparison

BCSharp:
Removes a warning turned error when built with .NET 10 SDK.

Maybe for the future `global.json` is needed.


slozier:
Thanks, looks like I had this change locally, guess I should have pushed it.

PR #1985, opened 2026-02-03
Downgrade macOS runner to Sonoma

BCSharp:
This is a temporary fix. macOS 14 Sonoma is the last [macOS runner image that comes with Mono pre-instaled](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md),
but will be [deprecated on 2026-07-06 and unsupported on 2026-11-02](https://github.com/actions/runner-images/issues/13518).

The Ubuntu runner has a similar problem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


slozier (approved):
Thanks! Works for now I guess.

carlaarchuleta707-lgtm (commented):
``

PR #1988, opened 2026-02-07
Upgrade to target net10.0

### LINKED ISSUE #1979: Support target .NET 10
.NET 10.0 is a Long Term Support release, available since last November. The SDK is now at 10.0.102, and Visual Studio at 18.2.1, seemingly resolving the early bugs. I have migrated two other projects to .NET 10 (though from 9.0) without any major issues, so I think it's time to add .NET 10 as target in IronPyton (and DLR). After that, .NET 6.0 support can be removed.

@slozier I am prepared to do the upgrade but I wanted first to check with you if you have already planned to do the same, so we don't duplicate the work.

BCSharp:
Resolves #1979

### DIFF
--- a/tests/suite/test_socket_stdlib.py
+++ b/tests/suite/test_socket_stdlib.py
@@ -6,7 +6,7 @@
 ## Run selected tests from test_socket from StdLib
 ##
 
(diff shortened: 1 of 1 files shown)


slozier (approved):
Thanks!

PR #1991, opened 2026-02-13
Register imported CLI sub-namespaces in sys.modules

### LINKED ISSUE #1989: Enum types are not picklable if defined in nested namespaces
This used to work on .NET:

```python
import clr
import pickle

clr.AddReference("System.Text.Json")
clr.AddReference("Microsoft.Scripting")

import System.Text.Json
import Microsoft.Scripting

data = [
    System.DayOfWeek.Wednesday,  # still works
    System.Text.Json.JsonValueKind.Object,  # nested namespace
    Microsoft.Scripting.Severity.Warning,  # different namespace than System
]

for value in data:
    print(pickle.dumps(value))
```

It does not work on .NET 8.0 and higher.

It was relying on `clr.Deserialize`, which was included if IronPython was compiled with `FEATURE_SERIALIZATION` (on by default). The implementation was using `BinaryFormatter`, which is now [unsupported](https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/8.0/binaryformatter-disabled). With `clr.Deserialize` and `clr.Serialize` gone, pickling relies on a generic object `__reduce_ex__` implementation. It fails for types that are placed in nested namespaces, because `__module__` attribute on those object is not the actual module as in imported `sys.modules` but the complete dotted-path namespace.

Two ideas come to mind on how to resolve this regression:

1. Tweak `pickle.py` in _StdLib_ to be more mindful of IronPython namespaces; for instance a translation of (_module_, _name_) from e.g. `(System.Text.Json, JsonValueKind)` to `(System, Text.Json.JsonValueKind)` is simple and would do the trick, but it will require using pickle protocol >= 4, which allows for dotted (nested) type names.
 
2. Capture the resolution of `__reduce_ex__` in `OneOffRTesolver` like it used to be when `FEATURE_SERIALIZATION` was defined, and redirect to `EnumOps.SerializeReduce`. This will handle all enums, but not other types that may have worked in case of option 1, OTOH, _StdLib_ is untouched.

(or the combination of the two).

Both options make objects pickled by IronPython/.NET 6.0 not unpicklable (not loadable) on .NET 8.0 and higher, but that is the consequence of `BinaryFormatter` being gone. AFAIK, pickle file format is not meant to be binary-portable across Python versions, so one can argue that this is better than the alternative of not being able to pickle at all.  Universal pickling of an arbitrary .NET object is probably not feasible, but it is feasible and desirable for simple built-in types. For user types for which pickle support is needed, the developer can implement their own `__reduce_ex__`.

See also #29

BCSharp:
Resolves #1989 as discussed (short term solution).

### DIFF
--- a/tests/suite/test_cliclass.py
+++ b/tests/suite/test_cliclass.py
@@ -1199,11 +1199,14 @@ End Class""")
                 System.StringSplitOptions.RemoveEmptyEntries,
                 ]
 
-        if False:
-            # TODO: Enum types are not picklable if defined in nested namespaces
-            # https://github.com/IronLanguages/ironpython3/issues/1989
-            clr.AddReference("System.Text.Json")
-            data.append(System.Text.Json.JsonValueKind.Object) # byte-based enum
+        if is_cli:
+            # enum types
+            clr.AddReference("Microsoft.Scripting")
+            import Microsoft.Scripting
+            data.append(Microsoft.Scripting.Severity.Warning)  # different namespace than System
+            if is_netcoreapp:
+                clr.AddReference("System.Text.Json")
+                data.append(System.Text.Json.JsonValueKind.Object) # byte-based enum
 
         data.append(list(data))     # list of all the data..
         data.append(tuple(data))    # tuple of all the data...


slozier (approved):
Looks good to me!

I did play around a bit with a meta path loader idea and it seems like we'd need to make the Python type of `NamespaceTracker` a bit more module-like to work with `importlib.__import__`, but it looks promising.

PR #1992, opened 2026-02-16
Remove target net6.0

BCSharp:
Since the project targets now net10.0, removing net6.0 (which is obsolete) will still keep two .NET frameworks targeted: net8.0 and net10.0


slozier (approved):
Thanks!

Issue #839, opened 2020-05-19
upper/lower/capitalize do not use the default case algorithm


slozier:
See https://docs.python.org/3/library/stdtypes.html#str.upper

>The uppercasing algorithm used is described in section 3.13 of the Unicode Standard.

This leads to the following failure in `string_tests.py`:
```Python
assert '\u1ff3\u1ff3\u1ffc\u1ffc'.capitalize() == '\u03a9\u0399\u1ff3\u1ff3\u1ff3'
```
(Note that capitalize changed with Python 3.8 so this assertion is only valid for previous versions).

PR #1993, opened 2026-02-16
Skip barred-lambda capitalize check for IronPython

BCSharp:
See #839


BCSharp:
It seems to me that both CPython and .NET are trying to be Unicode compliant, but it is a moving target. Since IronPython depends on .NET for this implementation, there isn't much to test here, except for rudimentary "it works" tests. So if the old CPython tests make assumptions that are no longer valid, I think simply bypassing them is good enough.

In this particular case what puzzles me is that the test used to work in the past, still works on CI, but fails locally on my machine. At first I thought that it fails with .NET 10, but no, it also fails with .NET 8. OTOH, I am using macOS 15 and macOS 26, and CI uses macOS 14.

Issue #1933, opened 2025-03-26
Migrate to SLNX


Lamparter:
SLNX (new XML-based Visual Studio solution format) is extensible and easier to edit than IronPython's growing solution file.
It would be a good idea to migrate to it.

slozier:
In theory I don't have an objection (it's indeed nicer). However, I would wait until the feature is no longer in preview. With the latest VS release (17.13.5) the feature is hidden behind a preview flag which is a bit of a barrier to entry. Currently, the main purpose of the solution file is to be opened in Visual Studio, so it has to work out-of-the-box.

Lamparter:
Yep, good idea

PR #1999, opened 2026-02-21
Migrate to SLNX 

slozier:
SLNX has been stable since 16.14 so we should be able to switch.

Resolves https://github.com/IronLanguages/ironpython3/issues/1933


BCSharp:
There is also `_ctypes_test.sln` in the repo but I guess we want to keep it as `.sln`? (I never used it, always use the `Makefile`).

slozier:
> There is also `_ctypes_test.sln` in the repo but I guess we want to keep it as `.sln`? (I never used it, always use the `Makefile`).

Good catch. I don't recall the last time I looked in that folder! :smile:

PR #2002, opened 2026-02-22
Prevent test filename clashes when running tests in parallel

BCSharp:
This is mostly useful when testing locally, because CI tests run in parallel on separate VMs.


slozier (approved):
I cry every time I look at code from the ipy suite... Thanks! 😁

PR #2004, opened 2026-02-24
Implement PEP 492: async/await support

mikasoukhov:
- Tokenizer/Parser: async def, async for, async with, await keywords
- AST nodes: AwaitExpression, AsyncForStatement, AsyncWithStatement
- Runtime: PythonCoroutine, CoroutineWrapper types
- Code generation: coroutines reuse generator state machine via yield from desugaring (await → yield from expr.__await__())
- Fix GeneratorRewriter VisitExtension to reduce one level at a time, preventing "must be reducible node" with DebugInfoRemovalExpression

Verified against CPython 3.14: 20/20 comparison tests identical.

### DIFF
--- a/src/core/IronPython/Runtime/Coroutine.cs
+++ b/src/core/IronPython/Runtime/Coroutine.cs
@@ -4,6 +4,9 @@
 
 #nullable enable
 
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
 using Microsoft.Scripting.Runtime;
 
 using IronPython.Runtime.Exceptions;
@@ -61,6 +64,38 @@ namespace IronPython.Runtime {
         public string __qualname__ {
             get => _generator.__name__;
         }
+
+        /// <summary>
+        /// Converts this coroutine into a .NET <see cref="Task{Object}"/>,
+        /// allowing C# code to <c>await</c> an IronPython async method.
+        /// The coroutine is driven on a single thread to avoid issues with
+        /// thread-local state in the Python generator runtime.
+        /// </summary>
+        public Task<object?> AsTask() {
+            return Task.Run(() => {
+                while (true) {
+                    object result = send(null);
+
+                    if (LightExceptions.IsLightException(result)) {
+                        var clrExc = LightExceptions.GetLightException(result);
+                        if (clrExc is StopIterationException) {
+                            var pyExc = ((IPythonAwareException)clrExc).PythonException;
+                            return pyExc is PythonExceptions._StopIteration si ? si.value : null;
+                        }
+                        throw clrExc;
+                    }
+
+                    if (result is Task task) {
+                        task.Wait();
+                    }
+                }
+            });
+        }
+
+        /// <summary>
+        /// Enables <c>await coroutine</c> from C# code.
+        /// </summary>
+        public TaskAwaiter<object?> GetAwaiter() => AsTask().GetAwaiter();
 
         internal PythonGenerator Generator => _generator;
 
--- a/tests/IronPython.Tests/CoroutineAsTaskTest.cs
+++ b/tests/IronPython.Tests/CoroutineAsTaskTest.cs
@@ -1,0 +1,169 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+#if NET
+
+using System;
+using System.Threading.Tasks;
+
+using IronPython.Hosting;
+using IronPython.Runtime;
+
+using Microsoft.Scripting.Hosting;
+
+using NUnit.Framework;
+
+namespace IronPythonTest {
+    public class CoroutineAsTaskTest {
+        private readonly ScriptEngine _engine;
+        private readonly ScriptScope _scope;
+
+        public CoroutineAsTaskTest() {
+            _engine = Python.CreateEngine();
+            _scope = _engine.CreateScope();
+        }
+
+        [Test]
+        public void AsTask_SimpleReturn() {
+            _engine.Execute(@"
+async def foo():
+    return 42
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = coro.AsTask().GetAwaiter().GetResult();
+            Assert.That(result, Is.EqualTo(42));
+        }
+
+        [Test]
+        public void AsTask_StringReturn() {
+            _engine.Execute(@"
+async def foo():
+    return 'hello'
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = coro.AsTask().GetAwaiter().GetResult();
+            Assert.That(result, Is.EqualTo("hello"));
+        }
+
+        [Test]
+        public void AsTask_AwaitCompletedTask() {
+            _engine.Execute(@"
+from System.Threading.Tasks import Task
+async def foo():
+    val = await Task.FromResult(10)
+    return val + 5
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = coro.AsTask().GetAwaiter().GetResult();
+            Assert.That(result, Is.EqualTo(15));
+        }
+
+        [Test]
+        public void AsTask_AwaitRealAsync() {
+            _engine.Execute(@"
+from System.Threading.Tasks import Task
+async def foo():
+    await Task.Delay(50)
+    return 'delayed'
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = coro.AsTask().GetAwaiter().GetResult();
+            Assert.That(result, Is.EqualTo("delayed"));
+        }
+
+        [Test]
+        public async Task AsTask_CanBeAwaited() {
+            _engine.Execute(@"
+from System.Threading.Tasks import Task
+async def foo():
+    await Task.Delay(50)
+    return 99
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = await coro.AsTask();
+            Assert.That(result, Is.EqualTo(99));
+        }
+
+        [Test]
+        public void AsTask_PropagatesException() {
+            _engine.Execute(@"
+async def foo():
+    raise ValueError('boom')
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            Assert.Throws<AggregateException>(() => coro.AsTask().Wait());
+        }
+
+        [Test]
+        public void AsTask_MultipleAwaits() {
+            _engine.Execute(@"
+from System.Threading.Tasks import Task
+async def foo():
+    a = await Task.FromResult(10)
+    b = await Task.FromResult(20)
+    return a + b
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = coro.AsTask().GetAwaiter().GetResult();
+            Assert.That(result, Is.EqualTo(30));
+        }
+
+        [Test]
+        public void AsTask_NoneReturn() {
+            _engine.Execute(@"
+async def foo():
+    pass
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = coro.AsTask().GetAwaiter().GetResult();
+            Assert.That(result, Is.Null);
+        }
+
+        [Test]
+        public async Task DirectAwait_Simple() {
+            _engine.Execute(@"
+async def foo():
+    return 42
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = await coro;
+            Assert.That(result, Is.EqualTo(42));
+        }
+
+        [Test]
+        public async Task DirectAwait_WithRealAsync() {
+            _engine.Execute(@"
+from System.Threading.Tasks import Task
+async def foo():
+    await Task.Delay(50)
+    return 'done'
+coro = foo()
+", _scope);
+
+            var coro = (PythonCoroutine)_scope.GetVariable("coro");
+            var result = await coro;
+            Assert.That(result, Is.EqualTo("done"));
+        }
+    }
+}
+
+#endif


tests/IronPython.Tests/Cases/CommonCases.cs
```diff
@@ -36,6 +37,42 @@ protected int TestImpl(TestInfo testcase) {
                 return -1;
             } finally {
                 m?.ReleaseMutex();
+                CleanupTempFiles(testcase);
```
slozier, line 40: Some test are running concurrently and this temp file cleanup breaks them.

slozier (commented):
Thanks this is great! It will probably take me a while to go over this properly but it definitely seems like a good start. So far I've taken a few notes but nothing blocking that can't be fixed later - will write them up when I have a bit more time.

It seems like CLA service is not happy. Maybe because of the brackets (`[` `]`) around company in the reply?

src/core/IronPython/Compiler/Ast/AstMethods.cs
```diff
@@ -79,6 +79,7 @@ internal static class AstMethods {
         public static readonly MethodInfo PushFrame = GetMethod((Func<CodeContext, FunctionCode, List<FunctionStack>>)PythonOps.PushFrame);
         public static readonly MethodInfo FormatString = GetMethod((Func<CodeContext, string, object, string>)PythonOps.FormatString);
         public static readonly MethodInfo GeneratorCheckThrowableAndReturnSendValue = GetMethod((Func<object, object>)PythonOps.GeneratorCheckThrowableAndReturnSendValue);
+        public static readonly MethodInfo MakeCoroutine = GetMethod((Func<PythonFunction, MutableTuple, object, PythonCoroutine>)PythonOps.MakeCoroutine);
```
slozier, line 82: Not spotting any calls to this one.

eng/scripts/generate_exceptions.py
```diff
@@ -85,6 +85,7 @@ def MakeNewException(self):
         ExceptionInfo('Exception', 'IronPython.Runtime.Exceptions.PythonException', None, (), (
                 ExceptionInfo('StopIteration', 'IronPython.Runtime.Exceptions.StopIterationException', None, ('value',), ()),
                 ExceptionInfo('StopAsyncIteration', 'IronPython.Runtime.Exceptions.StopAsyncIterationException', None, ('value',), ()),
+                ExceptionInfo('CancelledError', 'System.OperationCanceledException', None, (), ()),
```
slozier, line 88: `CancelledError` is not part of the bullt-in exception hierarchy. I guess this is here for .NET interop? I assume this is meant to be the same as Python's `asyncio.CancelledError`. Not sure how we'd do the mapping from `System.OperationCanceledException` to `asyncio.CancelledError` since it's defined in Python code. Maybe we don't need to and people should just catch `OperationCanceledException` when dealing with .NET async?

src/core/IronPython/Compiler/Ast/AsyncForStatement.cs
```diff
@@ -0,0 +1,146 @@
...
+                return n;
+            }
+
+            T WithSpan<T>(T node) where T : Node {
+                node.IndexSpan = span;
+                return node;
+            }
+
+            // _iter = ITER.__aiter__()
+            var aiterAttr = WithSpan(new MemberExpression(List, "__aiter__") { Parent = parent });
```
slozier, line 86: In theory `__aiter__` should use a special method lookup to get it from the class definition and not the instance. Probably not a big deal in practice.

For reference: https://docs.python.org/3/reference/compound_stmts.html#the-async-for-statement

src/core/IronPython/Compiler/Ast/AsyncForStatement.cs
```diff
@@ -0,0 +1,146 @@
...
+            var aiterCall = WithSpan(new CallExpression(aiterAttr, null, null) { Parent = parent });
+            var assignIter = WithSpan(new AssignmentStatement(new Expression[] { MakeName(iterName) }, aiterCall) { Parent = parent });
+
+            // running = True
+            var trueConst = new ConstantExpression(true) { Parent = parent }; trueConst.IndexSpan = span;
+            var assignRunning = WithSpan(new AssignmentStatement(new Expression[] { MakeName(runningName) }, trueConst) { Parent = parent });
+
+            // TARGET = await __aiter.__anext__()
+            var anextAttr = WithSpan(new MemberExpression(MakeName(iterName), "__anext__") { Parent = parent });
```
slozier, line 95: `__anext__` should use a special method lookup.

src/core/IronPython/Compiler/Ast/AsyncWithStatement.cs
```diff
@@ -0,0 +1,123 @@
...
+                n.IndexSpan = span;
+                return n;
+            }
+
+            // mgr = EXPR
+            var assignMgr = new AssignmentStatement(new Expression[] { MakeName("__asyncwith_mgr") }, ContextManager) { Parent = parent };
+            assignMgr.IndexSpan = span;
+
+            // await mgr.__aenter__()
+            var aenterAttr = new MemberExpression(MakeName("__asyncwith_mgr"), "__aenter__") { Parent = parent };
```
slozier, line 67: `__aenter__` should use a special method lookup.

src/core/IronPython/Compiler/Ast/AsyncWithStatement.cs
```diff
@@ -0,0 +1,123 @@
...
+                assignVar.IndexSpan = span;
+                bodyStmt = new SuiteStatement(new Statement[] { assignVar, Body }) { Parent = parent };
+            } else {
+                var exprStmt = new ExpressionStatement(awaitEnter) { Parent = parent };
+                exprStmt.IndexSpan = span;
+                bodyStmt = new SuiteStatement(new Statement[] { exprStmt, Body }) { Parent = parent };
+            }
+
+            // await mgr.__aexit__(None, None, None)
+            var aexitAttr = new MemberExpression(MakeName("__asyncwith_mgr"), "__aexit__") { Parent = parent };
```
slozier, line 86: `__aexit__` should use a special method lookup.

src/core/IronPython/Compiler/Ast/AsyncWithStatement.cs
```diff
@@ -0,0 +1,123 @@
...
+            var aexitCallNormal = new CallExpression(aexitAttr,
+                new Expression[] { none1, none2, none3 }, null) { Parent = parent };
+            aexitCallNormal.IndexSpan = span;
+            var awaitExitNormal = new AwaitExpression(aexitCallNormal);
+
+            // try/finally: await __aexit__ on normal exit
+            var finallyExprStmt = new ExpressionStatement(awaitExitNormal) { Parent = parent };
+            finallyExprStmt.IndexSpan = span;
+            var tryFinally = new TryStatement(bodyStmt, null, null, finallyExprStmt) { Parent = parent };
+            tryFinally.IndexSpan = span;
```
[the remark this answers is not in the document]
slozier: Indeed, not quite the same semantics as CPython... but probably good enough to bootstrap things. Here's the expected logic: https://docs.python.org/3/reference/compound_stmts.html#the-async-with-statement

tests/suite/test_async.py
```diff
@@ -0,0 +1,638 @@
...
+            cm = CM()
+            async with cm:
+                async for x in AsyncIter([1, 2]):
+                    cm.log.append(x)
+            return cm.log
+
+        self.assertEqual(run_coro(test()), ['enter', 1, 2, 'exit'])
+
+
+class DotNetAsyncInteropTest(unittest.TestCase):
```
slozier, line 322: Needs a `@skipUnlessIronPython()` guard

tests/suite/test_async.py
```diff
@@ -0,0 +1,638 @@
...
+    def test_cancellation_token_cancel(self):
+        """CancellationToken can be used with .NET async APIs."""
+        from System.Threading import CancellationTokenSource
+        cts = CancellationTokenSource()
+        token = cts.Token
+        self.assertFalse(token.IsCancellationRequested)
+        cts.Cancel()
+        self.assertTrue(token.IsCancellationRequested)
+
+    def test_await_valuetask(self):
```
slozier, line 426: Probably need a guard for .NET Framework which doesn't have `ValueTask`.

src/core/IronPython/Compiler/Parser.cs
```diff
@@ -1457,15 +1457,63 @@ private WithItem ParseWithItem() {
...
+            }
+
+            Eat(TokenKind.KeywordWith);
+            var withItem = ParseWithItem();
+            var header = GetEnd();
+            Statement body = ParseSuite();
+            AsyncWithStatement ret = new AsyncWithStatement(withItem.ContextManager, withItem.Variable, body);
+            ret.HeaderIndex = header;
+            ret.SetLoc(_globalParent, asyncStart, GetEnd());
+            return ret;
```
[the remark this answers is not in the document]
slozier: Did not verify the AI suggestion, but indeed should handle multiple with items.

src/core/IronPython/Compiler/Parser.cs
```diff
@@ -1457,15 +1457,63 @@ private WithItem ParseWithItem() {
...
         private Statement ParseAsyncStmt() {
+            var start = GetStart();
             Eat(TokenKind.KeywordAsync);
-            ReportSyntaxError("invalid syntax");
 
-            if (PeekToken().Kind == TokenKind.KeywordDef) {
-                FunctionDefinition def = ParseFuncDef(true);
-                return def;
+            switch (PeekToken().Kind) {
+                case TokenKind.KeywordDef:
```
slozier, line 1464: Not specifically related to this line, but also need to handle `async def` when parsing interactive input.

src/core/IronPython/Runtime/FunctionAttributes.cs
```diff
@@ -23,6 +23,10 @@ public enum FunctionAttributes {
         /// </summary>
         Generator = 0x20,
         /// <summary>
+        /// Set if the function is a coroutine (async def).
+        /// </summary>
+        Coroutine = 0x100,
```
slozier, line 28: This value is CO_ITERABLE_COROUTINE (0x100), should it be CO_COROUTINE (0x80) instead?

BCSharp: Hm, neither is a perfect match. Since IronPython's async function is basically still a generator (I can use `yield from` in the body), `CO_ITERABLE_COROUTINE` seems fitting.

But since the generator is wrapped in `PythonCoroutine` giving an external appearance of a native async function, I agree with @slozier that `CO_COROUTINE` is more appropiate, since it can be used by external code to decide whether to look up `cr_*` or `gi_*` attributes.

tests/IronPython.Tests/Cases/IronPythonCasesManifest.ini
```diff
@@ -4,6 +4,9 @@ WorkingDirectory=$(TEST_FILE_DIR)
 Redirect=false
 Timeout=120000 # 2 minute timeout
 
+[IronPython.test_async]
+IsolationLevel=PROCESS # loads IronPythonTest assembly, causes a failure in IronPython.test_attrinjector
```
BCSharp, line 8: I do not see any inherent reason why this test cannot be run the `SCOPE` isolation level (default). If it works in `PROCESS` but fails in `SCOPE`, this may indicate a bug in the expression rewriting code, since no rewriting is happening in the `PROCESS` isolation.

src/core/IronPython/Compiler/GeneratorRewriter.cs
```diff
@@ -589,11 +612,15 @@ protected override Expression VisitExtension(Expression node) {
                 return VisitYield(yield);
             }
 
-            if (node is FinallyFlowControlExpression ffc) {
-                return Visit(node.ReduceExtensions());
+            // Reduce one level and re-visit so that extension nodes produced
```
slozier, line 615: I've been trying to figure out what this is trying to solve. Maybe it's just my lack of understanding of the expression visitors, but I don't understand how `node.Reduce` is different from `node.ReduceExtensions` in this case. The only difference I can see is that `node.ReduceExtension` will call `node.ReduceAndCheck` instead of `node.Reduce`. `ReduceAndCheck` just calls `Reduce` with some extra checks to make sure the expression got reduced (same as the added if check) and a verifies that the type of the reduced node can be assigned to the type of the original node (why wouldn't we want this?)

slozier:
@BCSharp Unless you have any objections I think I'll merge this one as-is and start hacking away at issues.

BCSharp:
@slozier Go ahead. I am kind of waiting on this PR to start hacking on the DLR/net11 level.

tests/IronPython.Tests/Cases/IronPythonCasesManifest.ini
slozier: Looks like it's because two `IronPythonTest` assemblies were getting loaded. Once in `test_async` via `clr.AddReference("IronPythonTest")` and then a different one in `test_attrinjector` via `iptest.ipunittest.load_ironpython_test`. This causes `test_attrinjector` to raise a SystemError:
> Found multiple SpecialName methods for GetBoundMember on type System.Xml.XmlElement

 I wonder if the test order could be randomized... guessing we'd see more of these errors.

src/core/IronPython/Compiler/Parser.cs
```diff
@@ -1922,6 +1973,28 @@ private Expression ParsePower() {
...
+
+            if (current != null) {
+                current.IsGenerator = true;
+                current.GeneratorStop = GeneratorStop;
+            }
+
+            var start = GetStart();
+
+            // Parse the awaitable expression at the unary level
+            Expression expr = ParsePower();
```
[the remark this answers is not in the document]
slozier: `await -x` is not valid syntax, however the precendence relative to `**` was indeed incorrect. Fixed via https://github.com/IronLanguages/ironpython3/pull/2024

tests/IronPython.Tests/Cases/CommonCases.cs
slozier: Ended up removing this cleanup method. Resolved via https://github.com/IronLanguages/ironpython3/pull/2015

tests/suite/test_async.py
slozier: Resolved via https://github.com/IronLanguages/ironpython3/pull/2015

tests/suite/test_async.py
slozier: Resolved via https://github.com/IronLanguages/ironpython3/pull/2015

src/core/IronPython/Compiler/Parser.cs
slozier: Resolved via https://github.com/IronLanguages/ironpython3/pull/2021

src/core/IronPython/Runtime/FunctionAttributes.cs
slozier: Ended up changing it to `0x80` since there was a test in 3.6 checking for this value. Resolved in https://github.com/IronLanguages/ironpython3/pull/2017

tests/IronPython.Tests/Cases/IronPythonCasesManifest.ini
slozier: Resolved via https://github.com/IronLanguages/ironpython3/pull/2015

src/core/IronPython/Compiler/Parser.cs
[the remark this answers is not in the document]
slozier: Resolved this in https://github.com/IronLanguages/ironpython3/pull/2018

src/core/IronPython/Compiler/Ast/AsyncForStatement.cs
slozier: Resolved by https://github.com/IronLanguages/ironpython3/pull/2025

src/core/IronPython/Compiler/Ast/AsyncForStatement.cs
slozier: Resolved by https://github.com/IronLanguages/ironpython3/pull/2025

eng/scripts/generate_exceptions.py
Lamparter: Isn't asyncio implemented within Ironpython?

slozier: > Isn't asyncio implemented within Ironpython?

`asyncio` is part of the standard library (and probably doesn't work very well under IronPython) so `CancelledError` is not a built-in error, it's just defined in Python code.

src/core/IronPython/Compiler/GeneratorRewriter.cs
BCSharp: My understanding is that `node.ReduceExtensions` reduces the node recursively until the node is not reducible, while `node.Reduce` just performs one reduce step. Does it make sense?

slozier: Ah I see, `node.ReduceExtensions` is doing the reduce in a while loop. 🤦 

Still, I tried to figure out what code would require this change (without any luck) so I could add a unit test.

PR #2033, opened 2026-04-03
Replace Dictionary with HashSet

BCSharp:
This is the equivalent of IronLanguages/dlr#322 on the IronPython side. Maybe `HashSet` did not exist yet when this code was written.

It is a very opportunistic and straightforward change. Perhaps more possibilities for optimizations exist. One such place is `FunctionCode.cs`, which contains class `PythonDebuggingPayload.TracingWalker`. It defines several dictionaries where the value is `bool`. They look like sets, if the value is always, `true`, `false`, or ignored. Maybe this can be simplified somehow or maybe it is unfinished work, but will require more code analysis than what I am doing here for such a mechanical upgrade. Besides, this seems to be used only for debugging/tracing, when performance is not that important.


slozier (approved):
Thanks!

PR #2034, opened 2026-04-09
Fix whitespace

BCSharp:
Whitespace only. Brings the format from IronLanguages/dlr#324.


slozier:
Should we rename the file to fix the casing while we're at it?

BCSharp:
I haven't noticed it, but yes. Another difference is that IronPython uses `Project` with `ToolsVersion="Current"`, while the DLR uses `ToolsVersion="4.0"`. I think with modern SDK-stype projects it is ignored anyway. Should we get rid of them? Or standardize on one value?

I was also thinking of simply replacing the content of the IronPython `Tasks.targets` with an `<Import>` to `Tasks.targets` in the DLR, and avoid duplicating the file. Shall we do it?

slozier:
> I was also thinking of simply replacing the content of the IronPython `Tasks.targets` with an `<Import>` to `Tasks.targets` in the DLR, and avoid duplicating the file. Shall we do it?

Yeah that's fine since we already have some dependencies on the DLR repo. Although it might be interesting to be able to lose those at some point and have the projects be more independent.

BCSharp:
Well, if your intention is to further decouple the two repositories, let's not create more dependencies.

slozier (approved):

PR #2039, opened 2026-05-11
First class span types

BCSharp:
Now that we have target `net10.0` and all relevant projects use `LangVersion` _latest_, the code is compiled with C# 14, which introduces feature [First-class Span Types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-14.0/first-class-span-types). Consequently, the existing code can be simplified a bit — compiler takes over the span conversions.


slozier (approved):
Thanks!

PR #2040, opened 2026-05-12
Change CA2249 to warning

slozier:
This remove `CA2249`  from `.editorconfig` (thus changing it to warning) and ensures the polyfill is available.


BCSharp:
You may also want to update the Meizantou.Polyfill version at the same time, like the DLR. Polyfill `Contains` calls `IndexOf`, which in v.1.0.125+ has been updated with the `IndexOf` from the ex-DLR polyfill.

PR #2046, opened 2026-05-28
Implement async functions and generators (.NET)

BCSharp:
This is an alternative implementation of PEP 492 from #2004 by @mikasoukhov (Python 3.5). It builds on top of async support in the DLR (IronLanguages/dlr#362), which is around native .NET async primitives: `Task`/`Task<T>`, `IAsyncEnumerable<T>`, `IAsyncDisposable`. As such, it fits better within a .NET ecosystem, e.g. IronPython engine embedded within a larger C# application. As a .NET async mechanism, it obeys the .NET scheduling and context following mechanisms. For Python behavioural compatibility, wrappers and bridges are provided.

It also implements PEP 525 (async generators), which are accepted in Python 3.6.

The previous async functionality in IronPython is still there, parallel to the new one, and which one is being used depends whether `FEATURE_NET_ASYNC` is being defined during compilation. If defined, the new functionality is being used, otherwise the previous functionality is being used. It is practical for testing and development, but perhaps some discrepancies will occur. I think in the end it is best to settle on just one mechanism, but I'd postpone this decision until the whole `asyncio` support is complete.

As it is now, `FEATURE_NET_ASYNC` is defined on .NET platforms, but undefined on .NET Framework 4.6.2.

### DIFF
--- a/src/core/IronPython/Runtime/AsyncGenerator.cs
+++ b/src/core/IronPython/Runtime/AsyncGenerator.cs
@@ -1,0 +1,153 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+
+#if FEATURE_NET_ASYNC
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.Scripting.Runtime;
+
+using IronPython.Runtime.Exceptions;
+using IronPython.Runtime.Operations;
+using IronPython.Runtime.Types;
+
+namespace IronPython.Runtime {
+    /// <summary>
+    /// PEP 525 async generator object — what an <c>async def</c> with <c>yield</c> returns.
+    /// Façade over the DLR's <c>IAsyncEnumerable&lt;object&gt;</c> (produced by AsyncEnumerableExpression /
+    /// AsyncHelpers.DriveAsyncEnumerable) plus the channels that let the consumer drive it:
+    /// <list type="bullet">
+    ///   <item><c>_sendSlot</c> — the value delivered into the body at the next <c>yield</c> resume
+    ///   (<c>x = yield z</c>). Set by <see cref="asend"/> / <see cref="__anext__"/> before advancing.</item>
+    ///   <item><c>_throwSlot</c> — an exception rethrown at the next <c>yield</c> resume (athrow/aclose).</item>
+    ///   <item><c>_cts</c> — backs the underlying enumerator's cancellation.</item>
+    /// </list>
+    /// </summary>
+    [PythonType("async_generator")]
+    public sealed class PythonAsyncGenerator {
+        private readonly IAsyncEnumerator<object?> _enumerator;
+        private readonly StrongBox<object?> _sendSlot;
+        private readonly StrongBox<Exception?> _throwSlot;
+        private readonly CancellationTokenSource _cts;
+        private readonly string _name;
+        private bool _started;
+        private bool _closed;
+
+        internal PythonAsyncGenerator(IAsyncEnumerable<object?> source,
+                                       StrongBox<object?> sendSlot,
+                                       StrongBox<Exception?> throwSlot,
+                                       CancellationTokenSource cts,
+                                       string name) {
+            _enumerator = source.GetAsyncEnumerator(cts.Token);
+            _sendSlot = sendSlot;
+            _throwSlot = throwSlot;
+            _cts = cts;
+            _name = name;
+        }
+
+        public PythonAsyncGenerator __aiter__() => this;
+
+        /// <summary>
+        /// Advance the generator with no value sent in (equivalent to <c>asend(None)</c>). Returns an
+        /// awaitable yielding the next produced value or raising StopAsyncIteration.
+        /// </summary>
+        public object __anext__() {
+            _sendSlot.Value = null;
+            return new AsyncEnumeratorAwaitable<object?>(_enumerator);
+        }
+
+        /// <summary>
+        /// Send a value into the generator; it becomes the value of the <c>yield</c> the body is suspended
+        /// at (<c>x = yield z</c>). The first call after creation must send None (the body hasn't reached a
+        /// yield yet), matching CPython.
+        /// </summary>
+        public object asend(object? value) {
+            if (!_started) {
+                _started = true;
+                if (value is not null) {
+                    throw PythonOps.TypeError("can't send non-None value to a just-started async generator");
+                }
+            }
+            _sendSlot.Value = value;
+            return new AsyncEnumeratorAwaitable<object?>(_enumerator);
+        }
+
+        [LightThrowing]
+        public object athrow(object? type) => athrow(type, null, null);
+
+        [LightThrowing]
+        public object athrow(object? type, object? value) => athrow(type, value, null);
+
+        /// <summary>
+        /// Throw an exception into the generator at the suspended <c>yield</c>. Returns an awaitable; when
+        /// awaited the exception is rethrown at the resume point — if the body catches it and yields again
+        /// that value is produced, otherwise the exception propagates (or StopAsyncIteration if the body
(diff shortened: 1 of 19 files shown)


src/core/IronPython/Runtime/AsyncGenerator.cs
```diff
@@ -0,0 +1,160 @@
...
+                    throw PythonOps.TypeError("can't send non-None value to a just-started async generator");
+                }
+            }
+            _sendSlot.Value = value;
+            return new AsyncEnumeratorAwaitable<object?>(_enumerator);
+        }
+
+
+        [LightThrowing]
+        public object athrow(object? type) => athrow(type, null, null);
```
slozier: Just looking at future versions, it looks like the `athrow` argument is documented as `value` since 3.7 which seems reasonable to adopt? Also the other two overloads are deprecated in 3.12 (parhaps add a TODO comment?).

src/core/IronPython/Runtime/Coroutine.cs
```diff
@@ -14,6 +16,152 @@
...
+                // TODO: coroutines should be steppable via send(), one await at a time.
+                ((IAsyncResult)task).AsyncWaitHandle.WaitOne();
+            }
+
+            return SettleCompletedTask(task);
+        }
+
+
+        [LightThrowing]
+        public object @throw(object? type) => @throw(type, null, null);
```
slozier: Same as `AsyncGenerator`, the argument is `value` in 3.7 and the other overloads are deprecated in 3.12.

src/core/IronPython/Runtime/Coroutine.cs
```diff
@@ -14,6 +16,152 @@
 using IronPython.Runtime.Types;
 
 namespace IronPython.Runtime {
+
+#if FEATURE_NET_ASYNC
+
+    [PythonType("coroutine")]
```
slozier, line 20: Since there's some shared code, perhaps move the class signature outside the feature branch instead or repeating it? Although don't feel strongly either way...

tests/suite/test_asyncgen.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+    def test_async_generator_type(self):
+        async def agen():
+            yield 1
+        g = agen()
+        self.assertEqual(type(g).__name__, 'async_generator')
+
+
+    def test_asend(self):
```
slozier, line 98: I guess a bunch of these tests are not .NET specific. Maybe we could put them in a class that's not skipped? I guess our `run_coro` is .NET only, but that's easily substituted with `asyncio.run` (new in 3.7)... Ran into this while testing the tests against CPython.

slozier (approved):
Thanks! Apologies for the delay in reviewing this. Looks good to me!

I guess not having async generators in the other implementation is a bit of a blocker for a 3.6 release so this is good.

src/core/IronPython/Runtime/AsyncGenerator.cs
BCSharp: I do not see any changes in 3.7 on that. Do you have an URL to the documentation? When I run `help(mygen.athrow)` on 3.11.14, simply I get:

```
Help on built-in function athrow:

athrow(...) method of builtins.async_generator instance
    athrow(typ[,val[,tb]]) -> raise exception in generator.
```

Do you mean replacing the three overloads with one method and two default parameters? (This would relax the allowed calls to the method, e.g. `type` and `traceback` given, but null `value` — not in CPython)

src/core/IronPython/Runtime/Coroutine.cs
BCSharp: OK.

tests/suite/test_asyncgen.py
BCSharp: This is a good point but I am hesitant to make this test runnable under CPython as yet. The test uses `run_coro`, which is .NET specific, and its purpose it to be a smoke test for the new asyncgen code. There are some changes in `asyncio` past 3.4, and we don't have 3.6 tests in the main branch.

Ideally, we would migrate the main development to 3.6 and follow that `asyncio` tests. In 3.4, I wanted to enable some existing `asyncio` tests (not this PR), as long as they are not CPython-specific and do not conflict with 3.6+. They should cover whatever `test_asyncgen.py` covers for .NET.

BCSharp:
> Thanks! Apologies for the delay in reviewing this. Looks good to me!
> 
> I guess not having async generators in the other implementation is a bit of a blocker for a 3.6 release so this is good.

No problem, I, too, am less available for IronPython at the moment. My motivation with async was to get done whatever is blocking a switch to 3.6 as the main development branch. Can you add whatever you think is required to [milestone 3.6-alpha](https://github.com/IronLanguages/ironpython3/milestone/2)?

src/core/IronPython/Runtime/AsyncGenerator.cs
slozier: Here's from the 3.7 docs: https://docs.python.org/3.7/reference/expressions.html#agen.athrow

I initially noticed it in the help from 3.12.10 which is my "main" version these days:
```
Help on built-in function athrow:

athrow(...) method of builtins.async_generator instance
    athrow(value)
    athrow(type[,value[,tb]])

    raise exception in generator.
    the (type, val, tb) signature is deprecated,
    and may be removed in a future version of Python.
```

My idea was to leave it as three overloads since that matches the CPython positional-only style, but changing the argument of the first overload to `value`... although I guess practically it doesn't really change anything beyond documentation.

slozier:
> My motivation with async was to get done whatever is blocking a switch to 3.6 as the main development branch. Can you add whatever you think is required to [milestone 3.6-alpha](https://github.com/IronLanguages/ironpython3/milestone/2)?

I was actually thinking of how I could go about completely dropping the 3.6 branch and having all versions run off the main branch. I've already (mostly) converged the two implementations on the C# side of things opting to backport most features. So that leaves the Python code which varies quite a bit. For our test suite, I think things are mostly the same so we might be able to get away with some version checks? As for the standard library, I was thinking of splitting it off into its own repo which would have multiple branches. The idea would be to "just" set the version number and check out the standard library branch you're interested in working on. Anyway, just something I've been thinking about, not sure if you have thoughts or opinions on the matter.

Back to your question, I too would like to move the project along to version 3.6. Looking at the existing milestone list, I think some of them are in a state that's good enough for an alpha. So my short list would actually be:
- Insertion-order preserving dictionaries are a big one. I know it's technically a 3.7, but it was true and 3.6 and I expect a lot of code targetting 3.6 was dependant on this. There's also the related PEP 468.
- "Complete" async syntax support. I think there's still some work to be done regarding comprehension (see https://github.com/IronLanguages/ironpython3/pull/2023). Maybe not a big deal if some of it doesn't actually work, but syntax errors are problematic.
- UTF-8 console

This is conditional on seeing if I run into real-world issues with some of the common pure-Python packages (and by common I mean the ones I actually use 🤣).

Will try and get the list updated...

BCSharp:
> Will try and get the list updated...

Yes, please do. I will try to to the same on my end.

Form me, when I say "IronPython 3.6" I mainly mean IronPython with StdLib v3.6, because IronPython itself is a mix of features from various versions of Python anyway. I don't really mean "branch 3.6", I like developing on our main, so if you merge 3.6 to main, that's great. Having StdLib 3.4 and 3.6 side-by-side is an interesting idea, go for it if this is how you see it (and have the time :). I was simply thinking of one last release of 3.4 and then simply dropping StdLib 3.4 and replacing it by StdLib 3.6 (call it IronPython 3.6 alpha…).

For the order preserving dictionaries, you made a PR (#1908), which was a good approach and on which I commented, and then it stalled… Do you have comments on my comments, or how do you want to continue?

I can look into async comprehensions; I did some work on comprehensions in the past (though it was long time ago). My main concern with the whole async is having good test coverage without reinventing the wheel, i.e. letting the StdLib tests do the bulk of the work.

src/core/IronPython/Runtime/AsyncGenerator.cs
BCSharp: I looked into CPython's source code and it seems that `value`, when passed as the first parameter, is simply the same as `type` (dual use, type or value), so there is indeed no change to the code, just documentation.

PR #2056, opened 2026-06-16
Add tracing point for coroutine factory

### LINKED ISSUE #2055: IndexError when running empty async def when tracing is enabled
With the DLR based async implementation, running the cli with tracing enabled (`-X Tracing`) produces an `IndexError` (or an assertion error in debug mode). This occurs because the code below appears in `_collections_abc`. To reproduce, run the cli with the following arguments `-X Tracing -S` (`-S` is to avoid the `site` import which triggers the exception) and execute:
```py
async def _coro(): pass

_coro = _coro()
```

@BCSharp

BCSharp:
Resolves #2055. Apparently, every lambda has to have at least one tracing point.


slozier (approved):
Thanks!

Issue #2057, opened 2026-06-16
Incorrect debug rewriting of async generators


BCSharp:
```python
async def ag():          
    yield 1

ag()
```

Run with `-X Tracing -S`. Gives:

```
SystemError: yield and generator must have the same LabelTarget object
```

This is possibly requires a fix in the DLR. Discovered while working on #2055.

BCSharp:
Fixed, but the actual instrumentation requirement is under #2065.

PR #2059, opened 2026-06-17
Add tracing point for async generator

BCSharp:
This is the counterpart of #2056. It does not solve #2057, but it aligns async generators with coroutines, regardless how those two are being traced (to be done in a separate PR).


slozier (approved):
Thanks!

PR #2064, opened 2026-06-18
Use standard library from submodule

slozier:
This moves the standard library to a separate [ironpython3-stdlib](https://github.com/IronLanguages/ironpython3-stdlib/) repository. Any changes to the standard library have to be done in this new repository (though I expect these changes to be minimal). When building this repo, the version is dependant on the branch of the standard library that is checked out. This is done by redirecting the `CurrentVersion.props` to the submodule. The submodule will default to the `3.4` branch of `ironpython3-stdlib` until 3.6 becomes our main target. Assuming this is merged in, the `3.6` branch of this repo will become obsolete and will be removed.

This effectively unifies ironpython3 into a single codebase living on the `main` branch. For diverging implementations between 3.4 and 3.6 there are some conditional symbols (e.g. `PYTHON_34`, `PYTHON_36`, `PYTHON_36_OR_GREATER`) that can be used to branch the code. As for testing, I think we should be able to have side-by-side testing of both versions by also running CI with the `3.6` branch of `ironpython3-stdlib` checked out.

@BCSharp if you have any concerns please let me know. I don't want to impose this if you think it will be too complicated to work with. My main concern right now is that is complicates bisecting, but hopefully that will go away with time.


BCSharp:
I like the approach; it is definitely better than two IronPython branches that require merging back and forth. I agree with you that CI should test commits against both StdLib versions automatically. However, I can't find any place where the conditional symbols (e.g. PYTHON_34, PYTHON_36, PYTHON_36_OR_GREATER) are defined. I expected them being defined somewhere in the StdLib submodule props, and then being imported by IronPython. Is there a different mechanism? You may want to conditionally define them somewhere in IronPython props, based on property `MinorVersion`, for example.

For bisecting — I haven't tried it but setting:

```
git config --worktree submodule.IronPython.StdLib.update none
```
should allow you to keep a specific submodule untouched in the current worktree, checked out at a specific commit/branch. Assuming that the changes to StdLib are very limited and backward-compatible, this may be good enough.

slozier:
>However, I can't find any place where the conditional symbols (e.g. PYTHON_34, PYTHON_36, PYTHON_36_OR_GREATER) are defined. I expected them being defined somewhere in the StdLib submodule props, and then being imported by IronPython. Is there a different mechanism? You may want to conditionally define them somewhere in IronPython props, based on property `MinorVersion`, for example.

These are currently defined in `Directory.Build.props`. It'll break down once we hit 3.10, but that's a problem for the future. :smile:
https://github.com/IronLanguages/ironpython3/blob/864344dc3f73dbdfccd12f3705c86b186d3f571b/Directory.Build.props#L30-L31

BCSharp:
Thanks, works for me 👍

PR #2074, opened 2026-07-15
Add SingleOps.__round__ and optimizations in Builtin.round.

mickp:
Testing
===
```
>>> import System
>>> v = 987.654321
>>> assert round(System.Double(v)) == round(System.Single(v))
>>> print(round(System.Double(v), 3), round(System.Single(v), 3))
987.654 987.654
```


mickp:
I attempted to add assertions in tests/suite/test_builtinfunc.py, but I can't get those tests to run: they fail with
```
RUNNING IRONPYTHON TESTS USING THESE FLAGS:
  File "run.py", line 330, in <module>
  File "run.py", line 295, in runTest
  File "run.py", line 241, in runTestSlow
  File "run.py", line 226, in run_one_command
AttributeError: 'module' object has no attribute 'popen3'
```

slozier:
Thanks for the PR. I think it generally seems fine. I won't have time to properly review/test this for a couple weeks, but I'll get around to it as soon as I can.

Not sure why the test changes aren't working. I've been doing some structural changes recently so I may be to blame...

So I don't forget... I want to check are how it compares to "prior art" such as `np.float32`.

slozier (approved):
Thanks! Looks good to me. Might be useful to have a test to avoid future regressions.

Issue #2072, opened 2026-07-10
Stack frames missing in other threads when settrace used.


mickp:
Type checks in other threads were failing with `ValueError: call stack not deep enough` when I tried to trace my main thread. It looks like `settrace` causes stack to be lost. I do not see the same in CPython.

Repro
===

```
import argparse
import typing
import collections.abc
import sys
import threading

# This reproduction script demonstrates a "ValueError: call stack is not deep enough"
# in IronPython 3 when performing isinstance/issubclass checks against ABCs
# that have parameterized generics as subclasses.

class NotAMapping:
    pass
    
def tracer(frame, event, arg):
    return tracer

def do_check():
    print("Attempting isinstance check...")
    try:
        isinstance(NotAMapping(), collections.abc.Mapping)
        print("Success.")
    except Exception as e:
        print("\nCaught error: {}: {}".format(type(e).__name__, e))
        import traceback
        traceback.print_exc()

parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
parser.add_argument('-t', '--trace', action='store_true', help='Set tracer.')
args = parser.parse_args()
 
print(sys.version)

if args.trace:
    print("Enabling trace...")
    sys.settrace(tracer)

t = threading.Thread(target=do_check)
t.start()
t.join()
```

CPython 3.4
===
```
$ python repro_issue.py -t
3.4.0 (v3.4.0:04f714765c13, Mar 16 2014, 19:24:06) [MSC v.1600 32 bit (Intel)]
Enabling trace...
Attempting isinstance check...
Success.
```

IronPython 3.4
===
```
$ ipy.exe repro_issue.py -t
3.4.2 (3.4.2.1000)
[.NETFramework,Version=v4.6.2 on .NET Framework 4.8.9324.0 (64-bit)]
Enabling trace...
Attempting isinstance check...

Caught error: ValueError: call stack is not deep enough
Traceback (most recent call last):
  File "repro_issue.py", line 20, in do_check
    isinstance(NotAMapping(), collections.abc.Mapping)
  File "C:\Program Files\IronPython 3.4\lib\abc.py", line 191, in __instancecheck__
    return cls.__subclasscheck__(subclass)
  File "C:\Program Files\IronPython 3.4\lib\abc.py", line 226, in __subclasscheck__
    if issubclass(subclass, scls):
[...]

PR #2075, opened 2026-07-16
Fix traceback in threads when SetTrace used.

mickp:
Testing
===
The following script returns normamly, where previously it encountered: `ValueError: call stack is not deep enough`.

```
import sys
import threading
import typing

def tracer(frame, event, arg):
    return tracer

def do_check():
    isinstance(None, typing.Iterable)

sys.settrace(tracer)

t = threading.Thread(target=do_check)
t.start()
t.join()
```


mickp:
Resolves #2072

mickp:
I think the CI tests are failing here, for each configuration.

image

I see that fail intermittently when I test locally.

When I attach the debugger, the pass rate may be higher, and I see occasional failures at the location above, and here:

image

slozier:
Thanks for the PR! Will be a couple weeks before I  have time to properly look at this.

mickp:
I think this addresses the root cause of the call stack loss in threads, but I can not figure out how the change is getting hit to cause that one test failure in all runtime targets.

